PR #40, opened 2016-11-23
NewItem7 Minsky Model

halcrichtonstandish:
A minsky model of NewItem7.


highperformancecoder:
There is no Minsky model here, just an updated png file. Did you do a git add on the new file?

PR #41, opened 2016-11-30
Readme - getting started

kristianmandrup:
Converted Latex Getting Started doc to a Readme

### DIFF
--- a/Architecture.md
+++ b/Architecture.md
@@ -1,0 +1,58 @@
+# Architecture
+
+### engine
+
+The main engine written in C++
+
+`DAG` ie. Directed Acyclical Graph
+
+- `derivative` - contains implementation of symbolic differentiation
+- `equations` - `MathDAG` and more ...
+- `evalGodley` - evaluate a Godley Table
+- `evalOp` - evaluate an Operation. `EvalOp`, `ConstantEvalOp`, `EvalOpVector`
+- `expr` - syntactic sugar for representing expressions
+- `flowCoef` - a numerical coefficient times a variable (a "flow")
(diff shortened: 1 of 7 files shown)


Readme.md
```diff
@@ -0,0 +1,799 @@
...
+Go to our [SourceForge](http://minsky.sf.net) to download a version 
+
+Linux packages are available from the [OpenSUSE build service]
+(https://build.opensuse.org/package/show/home:hpcoder1/minsky)
+
+## Getting help
+Press the F1 key, or select ``help'' from the context menu. Help is
+context-sensitive.
+
+### Components of the Program
```
highperformancecoder: I'm inclined to think that anything after this point is a little too detailed for a README file. 

Since this is a top level web page, I would have thought something like a copy of what appears on https://minsky.sf.net might be more appropriate. Since we're going to be using both SourceForge and Github, the top level readmes need to be fairly short so they can be kept in sync. One difference between the sites is that the SourceForge site caters for end users (file release system, tickets, the manual - which currently is not available, it appears, mailing lists), whereas Github caters for developers (code repo, developer documentation). Much of the SF wiki could be ported over to Github, as it is mostly developer documentation.

Architecture.md
```diff
@@ -0,0 +1,58 @@
+# Architecture
```
highperformancecoder: These files are useful. You might like to take a look at the SourceForge wiki to see what to incorporate (https://sourceforge.net/p/minsky/wiki/Home/)

mandricore:
Yes, too much for a Readme, I agree... feel free to re-arrange as you see fit. Otherwise I will give it a go tmrw or this weekend.

PR #67, opened 2019-08-22
Added rudimentary non-linear least square fitting capability, still a WIP

wdednam:
This change is [Reviewable](https://reviewable.io/reviews/highperformancecoder/minsky/67)


model/minsky.cc
```diff
@@ -713,6 +717,103 @@ namespace minsky
...
+   };      
+	
+   int residual_func(const gsl_vector * x, void *params,
+          gsl_vector * f)
+   {
+	  if (params==NULL) return GSL_EBADFUNC;
+	   	  
+	  struct model_params *par = ((struct model_params *)params); 
+	  
+      int n;
```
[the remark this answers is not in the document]
wdednam: Thank you. In the updated version I committed today, the second option has been used.

model/minsky.cc
```diff
@@ -713,6 +717,103 @@ namespace minsky
...
+      for (size_t i = 0; i < n; i++)
+         xx[i] = gsl_vector_get(x,i); 
+       
+      // would like integration to continue from Minsky::step(). So, still a WORK IN PROGRESS
+      
+      //err=gsl_odeiv2_driver_apply(par->d->driver, par->t,numeric_limits<double>::max(),&xx[0]);
+	
+	  // define residual vector function to be optimized 
+      for (size_t i = 0; i < n; i++)
+          gsl_vector_set (f, i, xx[i] - par->x_init[i]);
```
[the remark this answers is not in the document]
wdednam: Still have to fix this. I could not find a vector operation in the online GSL documentation in order to accomplish this.

model/minsky.cc
```diff
@@ -713,6 +717,103 @@ namespace minsky
...
+      /* define the function to be minimized */
+      fdf.f = residual_func;
+      fdf.df = NULL; //func_df;   /* set to NULL for finite-difference Jacobian */
+      fdf.fvv = NULL;     /* not using geodesic acceleration */
+      fdf.n = n;
+      fdf.p = p;
+      fdf.params = params; 
+      
+      /* allocate workspace with default parameters */
+      w = gsl_multifit_nlinear_alloc (T, &fdf_params, n, p);
```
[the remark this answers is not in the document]
wdednam: This also remains to be fixed in my implementation. I will do so when I rework the residual function so that it calls Minsky::step(), as you have suggested.

model/minsky.cc
```diff
@@ -713,6 +717,103 @@ namespace minsky
...
+  patterned on first Large Nonlinear Least Squares Example.
+  See online documentation at https://www.gnu.org/software/gsl/doc/html/nls.html
+  */
+      
+namespace
+{	
+	
+   struct model_params {
+	   double *t;
+	   double *x_init;
```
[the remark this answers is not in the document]
wdednam: I have added a comment to this effect, thanks. I did not manage to rewrite the structure using references instead of pointers, so it will also have to done when I rework the implementation.

model/minsky.h
```diff
@@ -297,6 +297,7 @@ namespace minsky
     bool running=false; ///< controls whether simulation is running
     bool reverse=false; ///< reverse direction of simulation
     void reset(); ///<resets the variables back to their initial values
+    int LM_optimize(double x_out[],void *opt_pars); ///<minimize residual function by LM optimization
```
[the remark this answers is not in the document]
wdednam: Thanks, I have fixed this in the new commit I submitted today.

model/minsky.cc
```diff
@@ -713,6 +725,103 @@ namespace minsky
...
+namespace
+{	
+
+   int residual_func(const gsl_vector * x, void *params,
+          gsl_vector * f)
+   {
+	if (params==NULL) return GSL_EBADFUNC;
+	else { 	  
+		
+	  struct OPTpars *par = ((struct OPTpars *)params);
```
highperformancecoder: The modern way is to write this as 

`auto par=reinterpret_cast<OPTPars*>(params);`

highperformancecoder:
Wynand - note where it says "All checks have failed" above. If you press the "Details" button, it takes you to the Travis log where is fails to build. You might like to follow up on that. It does look like a merge issue - does your branch build locally?

model/minsky.cc
```diff
@@ -713,6 +725,103 @@ namespace minsky
...
+                throw error("unspecified error GSL_FAILURE returned");
+              case GSL_EBADFUNC: 
+                gsl_odeiv2_driver_reset(par->d->driver);
+                throw error("Invalid arithmetic operation detected");
+              default:
+                throw error("gsl error: %s",gsl_strerror(err));
+              }          
+          
+      }
+
```
highperformancecoder: It looks like you are missing a close brace here, for terminating the body of residual_func. I think that is why your compiles are failing.

Also note that you are doing an early return on line 741, so the else statement on line 742 is not necessary and only serves to increase the indent level unnecessarily (not good).

model/minsky.cc
```diff
@@ -713,6 +717,103 @@ namespace minsky
...
+  /*
+  For using GSL non-linear least square fitting routines
+  patterned on first Large Nonlinear Least Squares Example.
+  See online documentation at https://www.gnu.org/software/gsl/doc/html/nls.html
+  */
+      
+namespace
+{	
+	
+   struct model_params {
```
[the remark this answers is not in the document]
wdednam: I have now renamed the structs and variables to be consistent with PascalCase and camelCase, respectively.

model/minsky.cc
wdednam: I have used par = static_cast<OptParams*>(params), if that is okay?

model/minsky.cc
```diff
@@ -713,6 +725,103 @@ namespace minsky
...
+	else { 	  
+		
+	  struct OPTpars *par = ((struct OPTpars *)params); 
+	      
+          int err = GSL_SUCCESS;
+          auto n = ValueVector::stockVars.size();
+          vector<double> xx(n);
+          
+          for (size_t i = 0; i < n; i++)
+             xx[i] = gsl_vector_get(x,i);
```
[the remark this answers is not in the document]
wdednam: I have used memcpy(), and the time integration is now much smoother and more efficient, thank you.

model/minsky.cc
```diff
@@ -713,6 +725,103 @@ namespace minsky
...
+              {
+              case GSL_SUCCESS: case GSL_EMAXITER: return GSL_SUCCESS;
+              case GSL_FAILURE:
+                throw error("unspecified error GSL_FAILURE returned");
+              case GSL_EBADFUNC: 
+                gsl_odeiv2_driver_reset(par->d->driver);
+                throw error("Invalid arithmetic operation detected");
+              default:
+                throw error("gsl error: %s",gsl_strerror(err));
+              }
```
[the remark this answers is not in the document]
wdednam: The pattern inside residualFunc() is now "if ... try ... catch ... return" as in RKfunction() and jacobian().

model/minsky.cc
```diff
@@ -724,18 +833,28 @@ namespace minsky
...
+	          struct OPTpars pars { &tp, &stockVarsCopy[0],ode};	
+			  
+	          gsl_odeiv2_driver_set_nmax(pars.d->driver, nSteps);
+                  // we need to update Minsky's t synchronously to support the t operator
+                  // potentially means t and stockVars out of sync on GUI, but should still be thread safe
+                  err=gsl_odeiv2_driver_apply(pars.d->driver, pars.t, numeric_limits<double>::max(),pars.x_init);                      
+                             
+              	  // minimize difference between initial stock variable values and values at end point of prior integration step, still a WORK IN PROGRESS
+                  // the optimized stock variables are in theory updated in pars.x_init which is an alias for stockVarsCopy
+                  opt_err = LM_optimize(pars.x_init,pars);
```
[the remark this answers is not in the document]
wdednam: According to the GSL online documentation, there are three possibilities for opt_err returned by optimize: GSL_SUCCESS, GSL_EMAXITER or GSL_ENOPROG. I have changed the switch(opt_err) on line 911 to reflect these return values bu have to recommit these changes, sorry.

model/minsky.h
```diff
@@ -297,6 +298,7 @@ namespace minsky
     bool running=false; ///< controls whether simulation is running
     bool reverse=false; ///< reverse direction of simulation
     void reset(); ///<resets the variables back to their initial values
+    int LM_optimize(double x_out[],OPTpars& opt_params); ///<minimize residual function by Levenberg-Marquardt optimization
```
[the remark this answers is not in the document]
wdednam: I think I have done the same way @return has been used elsewhere, but please let me know when I recommit in a short while.

model/minsky.cc
wdednam: Thank you, I have fixed this.

Makefile
```diff
@@ -9,8 +9,8 @@ SF_WEB=hpcoder@web.sourceforge.net:/home/project-web/minsky/htdocs
 # location of TCL and TK libraries 
 TCL_PREFIX=$(shell grep TCL_PREFIX $(call search,lib*/tclConfig.sh) | cut -f2 -d\')
 TCL_VERSION=$(shell grep TCL_VERSION $(call search,lib*/tclConfig.sh) | cut -f2 -d\')
-TCL_LIB=$(dir $(shell find $(TCL_PREFIX) -name init.tcl -path "*/tcl$(TCL_VERSION)*" -print))
-TK_LIB=$(dir $(shell find $(TCL_PREFIX) -name tk.tcl -path "*/tk$(TCL_VERSION)*" -print))
+TCL_LIB=/usr/lib/tcl8.6#$(dir $(shell find $(TCL_PREFIX) -name init.tcl -path "*/tcl$(TCL_VERSION)*" -print))
+TK_LIB=/usr/lib/tk8.6#$(dir $(shell find $(TCL_PREFIX) -name tk.tcl -path "*/tk$(TCL_VERSION)*" -print))
```
highperformancecoder: You should avoid adding distro specific information like this. EcoLab's configure scripts should be able to find TCL/Tk - if not, then let me know so I can fix it.

Makefile
```diff
@@ -33,7 +33,7 @@ endif
 include $(ECOLAB_HOME)/include/Makefile
 
 # override the install prefix here
-PREFIX=/usr/local
+PREFIX=/home/wd/usr/minsky#/usr/local
```
highperformancecoder: You can override Makefile variables on the command line:``make PREFIX=~/usr/minsky``

Makefile
```diff
@@ -58,7 +58,8 @@ FLAGS+=-DENABLE_DARWIN_EVENTS -DMAC_OSX_TK
 endif
 
 # TODO - remove dependency on GUI directory here
-FLAGS+=-std=c++11 -Ischema -Iengine -Imodel $(OPT) -UECOLAB_LIB -DECOLAB_LIB=\"library\" -Wno-unused-local-typedefs
+#FLAGS+=-std=c++11 -Ischema -Iengine -Imodel -IRESTService -I/home/wd/usr/json_spirit/include $(OPT) -UECOLAB_LIB -DECOLAB_LIB=\"library\" -Wno-unused-local-typedefs
+FLAGS+=-std=c++11 -Ischema -Iengine -Imodel -IRESTService -I/home/wd/usr/json_spirit/include $(OPT) -UECOLAB_LIB -DECOLAB_LIB=\"library\" -Wno-unused-local-typedefs
```
highperformancecoder: If json_spirit were installed in a standard place (/usr, /usr/local or ~/usr), then this is unnecessary.

Makefile
```diff
@@ -102,7 +103,7 @@ BOOST_EXT=
 $(warning Boost extension=$(BOOST_EXT))
 endif
 
-LIBS+=	-ljson_spirit \
+LIBS+= -L/home/wd/usr/json_spirit/lib -ljson_spirit \
```
highperformancecoder: Ditto

model/minsky.cc
```diff
@@ -713,6 +726,96 @@ namespace minsky
...
+	if (params==NULL) return GSL_EBADFUNC;
+	try { 	  
+		
+	      auto pars=static_cast<OptParams*>(params);  
+	      
+          int err = GSL_SUCCESS;
+          auto n = ValueVector::stockVars.size();
+          vector<double> xx(n);
+          
+          memcpy(xx.data(),reinterpret_cast<const vector<double>*>(x),n);
```
highperformancecoder: This is wrong. A gsl_vector is most categorically not a vector of double. It might accidentally work if the address of the gsl_vector's data coincides with its address, but that doesn't strike me as very likely. How would it know how long it is, fo instance?

model/minsky.cc
```diff
@@ -713,6 +726,96 @@ namespace minsky
...
+		((Minsky*)params)->threadErrMsg=e.what();
+        return GSL_EBADFUNC; 
+	 }   
+	 return  GSL_SUCCESS;      
+                        
+   }
+   
+   static void errHandlerOpt(const char* reason, const char* file, int line, int gsl_errno) {
+         throw error("gsl: %s:%d: %s",file,line,reason);
+   }
```
highperformancecoder: I'm a bit dubious about this approach. This callback will be called from within GSL, and we shouldn't throw exceptions within GSL, as it's not exception safe.

model/minsky.cc
```diff
@@ -713,6 +726,96 @@ namespace minsky
...
+         throw error("gsl: %s:%d: %s",file,line,reason);
+   } 
+}      
+
+  int Minsky::optimize(double xf[], OptParams& optPars)
+  {
+    	
+      gsl_set_error_handler(errHandlerOpt);	
+      const gsl_multifit_nlinear_type *T = gsl_multifit_nlinear_trust;
+      gsl_multifit_nlinear_workspace *w;
```
highperformancecoder: I don't see anywhere where this variable is initialised

model/minsky.cc
```diff
@@ -713,6 +726,96 @@ namespace minsky
...
+      gsl_multifit_nlinear_fdf fdf;
```
highperformancecoder: Should move this declaration to just before it is initialised on line 793.

model/minsky.cc
highperformancecoder: Ah found it. Please move the declaration to line 891 where it is initialised.

model/minsky.cc
```diff
@@ -713,6 +726,96 @@ namespace minsky
...
+      gsl_multifit_nlinear_init (&x.vector, &fdf,w);
+      
+      /* compute initial cost function */
+      f = gsl_multifit_nlinear_residual(w);
+      
+      /* solve the system with a maximum of 200 iterations */
+      status = gsl_multifit_nlinear_driver(200, xtol, gtol, ftol,
+                                           NULL, NULL, &info, w);
+                                           
+      memcpy(xf,reinterpret_cast<const vector<double>*>(w->x),n);
```
highperformancecoder: The cast is wrong. Is w->x a simple array of doubles?

model/minsky.h
```diff
@@ -297,6 +298,8 @@ namespace minsky
     bool running=false; ///< controls whether simulation is running
     bool reverse=false; ///< reverse direction of simulation
     void reset(); ///<resets the variables back to their initial values
+    /// @return optimize() returns an integer of the status of non-linear least square optimization, either GSL_SUCCESS, GSL_EMAXITER or GSL_ENOPROG
+    int optimize(double xf[],OptParams& optPars); ///<minimize residual function by LM optimization
```
highperformancecoder: optimize may as well be a private method, as there is no way you can use it without knowing the structure of OptParams.

Makefile
wdednam: I have removed them, they are not needed.

Makefile
wdednam: I have removed the local prefix, thanks.

Makefile
wdednam: I have reinstalled json in ~/usr, so no need to refer to the local include directory anymore.

Makefile
wdednam: It is gone from the Makefile.

model/minsky.cc
wdednam: Okay, I don't know why I thought I had to recast the gsl_vector x to the same type as xx, which is of type vector<double>, because it "works" as just memcpy(xx.data(),x,n). But, then again, it doesn't "work" when I use memcpy(xx.data(),x,sizeof(x)) as I'm supposed to? So, this makes me doubt the implementation now...

model/minsky.cc
wdednam: Okay, I need to think about moving it, because I don't think it belongs in class Minsky either. As you have suggested before, the whole optimization data structure will have to go into a pimpl pattern much as RKdata, which will also take care of the public/private method question you pose below.

model/minsky.cc
wdednam: Done.

model/minsky.cc
wdednam: I have rearranged the code in Minsky::optimize() to make it more logical, thanks.

model/minsky.cc
wdednam: I have replaced it with memcpy(xf,w->x,n) as inside residualFunc, but am dubious about the size of w->x too.

model/minsky.h
wdednam: Agreed, it will have to go into a pimpl pattern like RKdata and I don't think it can be merged at all as it currently stands. At the very minimum, a tick box in the Runge Kutta drop-down menu in Minsky's GUI will be needed as an option to activate optimization. Most likely, it will require its own drop-down menu.

model/minsky.cc
```diff
@@ -713,6 +727,101 @@ namespace minsky
...
+	if (params==NULL) return GSL_EBADFUNC;
+	try { 	  
+		
+	      auto pars=static_cast<OptParams*>(params);  
+	      
+          int err = GSL_SUCCESS;
+          auto n = ValueVector::stockVars.size();
+          vector<double> xx(n);
+          
+          memcpy(xx.data(),x,n);
```
highperformancecoder: The correct way of doing this should be
``vector<double> xx(x->size); 
memcpy(x.data(), x->data, x->size * sizeof(double));``

Doing it the way your doing will most likely lead to memory corruption, I don't know why it is working for your.
See [gsl_vector](https://www.gnu.org/software/gsl/doc/html/vectors.html)

highperformancecoder: Note - the above code is only valid if x->stride == 1.

highperformancecoder: If stride>1, then you may be able to use gsl_vector_memcpy instead.

wdednam: Thank you, this is an important lesson. I should never assume that because gsl_vector and vector<double> are both vectors that their internal data structures will be similar.

model/godleyTableWindow.cc
```diff
@@ -94,11 +94,19 @@ namespace minsky
...
+        break;
+      case 3:
+        if (pos==middle)             // Fourth button on first and second row cannot move initial conditions row. For ticket 1064
+=======
         else if (pos !=first)                  // Third button cannot swap column headings and initial conditions row values. For ticket 1064     
           godleyIcon.table.moveRow(idx,-1);   
         break;
       case 3:
         if (pos==second || pos==middle)             // Fourth button on first and second row cannot move initial conditions row. For ticket 1064
+>>>>>>> e0543ca9043a4b33cdf459b9ce96c719a8bc8450
```
jnavila: Obviously a missing conflict resolution.

wdednam: Thank you, I have removed the conflicting lines, but travis has also picked up ambiguous parentheses in if statements which I've just fixed. I hope the code will build now...

highperformancecoder:
Hi Wynand, this is now in a completely unmergeable state. I think what you need to do is merge upstream into this pull request branch. You may as well merge it into your master branch at the same time.

wdednam:
I've created a new branch "fixes" following the suggestions in the third answer at [](https://stackoverflow.com/questions/16748115/how-to-modify-github-pull-request). I did not manage to find a way to merge the upstream branch 2.15.10 into the existing pull request branch (because I couldn't find its ref from the terminal). So perhaps we can close this pull request? I will put in a new one from the new base branch.

highperformancecoder:
As per your request, closing this PR

PR #68, opened 2019-10-04
Proposed solutions to bugs #1061 and #1064

wdednam:
This change is [Reviewable](https://reviewable.io/reviews/highperformancecoder/minsky/68)

### DIFF
--- a/gui-tk/wiring.tcl
+++ b/gui-tk/wiring.tcl
@@ -1198,7 +1198,6 @@ proc setIntegralIValue {} {
     global constInput
     set item minsky.canvas.item
     $item.description $constInput(Name)
-    $item.name $constInput(Name)
     # description may have change intVar, so use value instead to set init
     getValue [$item.valueId]
     value.init $constInput(Value)
--- a/gui-tk/wiring.tcl
+++ b/gui-tk/wiring.tcl
@@ -562,6 +562,13 @@ proc renameVariableInstances {} {
     wm title .renameDialog "Rename [minsky.canvas.item.name]"
 }
 
+proc renameIntegralInstances {} {
+    textEntryPopup .renameDialog [minsky.canvas.item.description] {
+        canvas.renameAllInstances [.renameDialog.entry get]
+        canvas.requestRedraw
+    }
+    wm title .renameDialog "Rename [minsky.canvas.item.description]"
+}
 
 proc findDefinition {} {
     if [canvas.findVariableDefinition] {
@@ -629,13 +636,13 @@ proc contextMenu {x y X Y} {
             .wiring.context add command -label "Copy" -command "canvas.copyItem"
             .wiring.context add command -label "Flip" -command "$item.flip; flip_default"
             if {[$item.type]=="integrate"} {
-                .wiring.context add command -label "Toggle var binding" -command "minsky.canvas.item.toggleCoupled; canvas.requestRedraw"
+               .wiring.context add command -label "Toggle var binding" -command "minsky.canvas.item.toggleCoupled; canvas.requestRedraw"
             .wiring.context add command -label "Select all instances" -command {
                 canvas.selectAllVariables
             }
             .wiring.context add command -label "Rename all instances" -command {
-                renameVariableInstances
-            }
+				renameIntegralInstances
+			}	
             }
         }
         "PlotWidget" {
@@ -1191,6 +1198,7 @@ proc setIntegralIValue {} {
     global constInput
     set item minsky.canvas.item
     $item.description $constInput(Name)
+    $item.name $constInput(Name)
     # description may have change intVar, so use value instead to set init
     getValue [$item.valueId]
     value.init $constInput(Value)
--- a/model/godleyTableWindow.cc
+++ b/model/godleyTableWindow.cc
@@ -86,19 +86,17 @@ namespace minsky
       case 0:
         godleyIcon.table.insertRow(idx+1);
         break;
-      case 1:      
-        if (pos!=first && pos!=firstAndLast) godleyIcon.deleteRow(idx+1); // Initial conditions row cannot be deleted, even when it is the only row in the table. For ticket 1064
+      case 1:
+        godleyIcon.deleteRow(idx+1);
         break;
-      case 2:  
-        if (pos==first && pos==firstAndLast) // Third button cannot move initial conditions row, even when it is the only row in the table. For ticket 1064
-          godleyIcon.table.moveRow(idx,0); 
-        else if (pos==second)                  // Third button of second row cannot swap initial conditions and second row. For ticket 1064
-           godleyIcon.table.moveRow(idx,1); 
-        else if (pos!=first && pos!=firstAndLast)                  // Third button cannot swap column headings and initial conditions row values. For ticket 1064     
-          godleyIcon.table.moveRow(idx,-1);   
+      case 2:
+        if (pos==first)
+          godleyIcon.table.moveRow(idx,1);
+        else
+          godleyIcon.table.moveRow(idx,-1);
         break;
       case 3:
-        if (pos==middle)             // Fourth button on first and second row cannot move initial conditions row. For ticket 1064
+        if (pos==middle)
           godleyIcon.table.moveRow(idx,1);
         break;
       }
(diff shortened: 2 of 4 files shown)


gui-tk/wiring.tcl
```diff
@@ -1191,6 +1198,7 @@ proc setIntegralIValue {} {
     global constInput
     set item minsky.canvas.item
     $item.description $constInput(Name)
+    $item.name $constInput(Name)
```
highperformancecoder: My understanding is that at this point, item is an integral, so this change would be wrong.

model/godleyTableWindow.cc
```diff
@@ -86,17 +86,19 @@ namespace minsky
       case 0:
         godleyIcon.table.insertRow(idx+1);
         break;
-      case 1:
-        godleyIcon.deleteRow(idx+1);
+      case 1:
```
highperformancecoder: It is better practice to have separate pull request for separate issues, as it make reviewing and so on easier. But OK, lets not worry about that now.

model/godleyTableWindow.cc
```diff
@@ -86,17 +86,19 @@ namespace minsky
...
+        if (pos!=first && pos!=firstAndLast) godleyIcon.deleteRow(idx+1); // Initial conditions row cannot be deleted, even when it is the only row in the table. For ticket 1064
         break;
-      case 2:
-        if (pos==first)
-          godleyIcon.table.moveRow(idx,1);
-        else
-          godleyIcon.table.moveRow(idx,-1);
+      case 2:  
+        if (pos==first && pos==firstAndLast) // Third button cannot move initial conditions row, even when it is the only row in the table. For ticket 1064
+          godleyIcon.table.moveRow(idx,0);
```
highperformancecoder: Is this a no operation?

highperformancecoder: pos cannot be both first and firstAndLast concurrently, so this stanza is never executed,

model/godleyTableWindow.cc
```diff
@@ -347,41 +349,44 @@ namespace minsky
               highlightColumn(surface->cairo(),selectedCol);
               highlightColumn(surface->cairo(),motionCol);
             }
-          else if (motionRow>=0 && selectedCol==0 && selectedRow>0) // whole col being moved
+          else if (motionRow>=0 && selectedCol==0 && selectedRow>0) // whole Row being moved
```
highperformancecoder: Thanks!

model/godleyTableWindow.cc
```diff
@@ -347,41 +349,44 @@ namespace minsky
...
               highlightRow(surface->cairo(),selectedRow);
               highlightRow(surface->cairo(),motionRow);
             }
           else if (selectedCol==0 || /* selecting individual cell */
-                   (selectedCol>=int(scrollColStart) && selectedCol<int(godleyIcon->table.cols())))
+                   (selectedCol>=int(scrollColStart) && selectedCol<int(godleyIcon->table.cols())))   
             {
-              if (selectedRow!=0 || selectedCol!=0) // can't select flows/stockVars label
+              if (selectedRow!=0 || selectedCol!=0) // can't select flows/stockVars
                 {
```
highperformancecoder: I think this would be more simply achieved by changing this condition to
if (selectedRow>1 || selectedCol!=0)
rather than nesting the conditional.

Need to check if selectedRow could be negative.

model/godleyTableWindow.cc
```diff
@@ -903,12 +914,12 @@ namespace minsky
 
   void GodleyTableWindow::checkCell00()
   {
-    if (selectedCol==0 && selectedRow==0)
-      // (0,0) cell not editable
+    if ((selectedCol==0 && selectedRow==0) || (selectedCol==0 && selectedRow==1))
```
highperformancecoder: Simplify to selectedCol==0 && (selectedRow==0 || selectedRow==1), or even
selectedCol==0 && selectedRow>1

model/godleyTableWindow.h
```diff
@@ -42,7 +42,7 @@ namespace minsky
   {
     CLASSDESC_ACCESS(ButtonWidget);
   protected:
-    GodleyIcon& godleyIcon;
+    GodleyIcon& godleyIcon;  //reference
```
highperformancecoder: Comment is redundant. Comments should not tell you what the code does, but rather the intent or design. It answer why questions, not how questions.

model/godleyTableWindow.cc
```diff
@@ -987,13 +998,15 @@ namespace minsky
...
     int idx=0;
-    drawButton(cairo,"+",0,1,0,idx++);
-    drawButton(cairo,"—",1,0,0,idx++);
-    if (pos!=first && pos!=firstAndLast)
-      drawButton(cairo,rowCol==row? "↑": "←",0,0,0,idx++);
-    if (pos!=last && pos!=firstAndLast)
-      drawButton(cairo,rowCol==row? "↓": "→",0,0,0,idx++);
-  }
+      drawButton(cairo,"+",0,1,0,idx++);
+      if ((pos!=first && pos!=firstAndLast) || rowCol == col) 	// no delete button for first row containing initial conditions. For ticket 1064
```
highperformancecoder: These comments are good!

model/godleyTableWindow.h
```diff
@@ -80,10 +80,10 @@ namespace minsky
     /// minimum column width (for eg empty columns)
     static constexpr double minColumnWidth=4*ButtonWidget<col>::buttonSpacing;
 
-    std::shared_ptr<GodleyIcon> godleyIcon;
+    std::shared_ptr<GodleyIcon> godleyIcon;   //pointer
```
highperformancecoder: ditto.

highperformancecoder:
The proposed code changes look good. Just a couple of minor improvements, and you need to fix the failing unit test.

gui-tk/wiring.tcl
wdednam: Oops, I forgot that there. Thanks, I've deleted it.

model/godleyTableWindow.cc
wdednam: Should I rather only submit one PR at a time? I ask because when I tried to create a second pull request (across forks) for ticket 1061, Github would only let me edit the one for ticket 1064 and not create an additional one.

model/godleyTableWindow.cc
wdednam: Thanks, I've removed that operation.

model/godleyTableWindow.cc
wdednam: Sure :)

model/godleyTableWindow.cc
wdednam: Great, thanks, it works perfectly. I've made it 
 if ((selectedRow>1 || selectedRow <0) || selectedCol!=0)
to check whether selectedRow could also be negative.

model/godleyTableWindow.cc
wdednam: I'm using your second suggestion, thanks :)

model/godleyTableWindow.h
wdednam: Oops, I forgot that there. I've removed it, thanks for catching it.

model/godleyTableWindow.cc
wdednam: Thanks :)

model/godleyTableWindow.h
wdednam: I've removed it.

wdednam:
Thank you, I'll commit and push them again. From the travis log it looks like test/00/t0001a.sh is failing because of the changes I've made to the rowWidget buttons? So, if I understand correctly, I only need to modify testModel.cc between lines 1067 and 1386 to reflect my changes?

model/godleyTableWindow.cc
highperformancecoder: I fell into this trap the first time I did a pull request. Each pull request has its own branch. Committing to that branch updates the pull request. So instead of pull requesting of the one branch (fixes), you would have separate branches bugfix-1061, bugfix-1064 (these are just suggested names, you can name them how you like) based from highperformancecoder:master, and cherry pick the changesets you want into those branches. Instead of cherry picking, you can just work in the bugfix/feature branches, then have a "develop" branch which is a merge of the pull request branches for your own testing/usage purposes. Either way works.

But yes - you can have multiple pull requests open concurrently, but each needs its own branch. Please keep the pull requests to a single issue in the future. I don't mind making the exception for this one, but in this case #1061 is fixed completely, but #1064 still has a bit of work, so we could have merged one PR already :).

model/godleyTableWindow.cc
highperformancecoder: But not committed yet?

highperformancecoder:
You've got it!  Those are the 3 failing tests, and these need to be modified to account for your changes. The idea is to keep the intent of the testing, or improve the testing, not make them worse. As one of my Rapiscan colleague quips: "[We are not Volkswagen](https://github.com/auchenberg/volkswagen)"

model/godleyTableWindow.cc
wdednam: Thank you for clearing that up, it can be confusing. I will have separate branches for bugfixes and feature additions from now on.

model/godleyTableWindow.cc
wdednam: I have decided to fix testCode.cc first. But will commit everything now.

wdednam:
That makes perfect sense, thanks. I completely missed the test modules. But I have now made changes to testModel.cc account for changes in godleyTableWindow.cc. I have also added a function checkCell10() to godleyTableWindow.cc and godleyTableWindow.h to check for the selection of cell (1,0), instead of having it all in checkCell00(). The Volkswagon Github page is funny: "Volkswagen detects when your tests are being run in a CI server, and _makes them pass_." Is this in any way related to the problems they've had in Europe with their emission standards?

test/testModel.cc
```diff
@@ -1257,14 +1274,17 @@ SUITE(GodleyTableWindow)
...
+      // Row 1 can no longer move down, so the following is no longer needed
+      
+      //x=2*ButtonWidget<row>::buttonSpacing+1, y=5+topTableOffset+rowHeight;
+      //CHECK_EQUAL(rowWidget, clickType(x,y));
+      //mouseDown(x,y);
+      //// should have invoked moving row 1 down
+      //CHECK_EQUAL("r2c2",godleyIcon->table.cell(1,2));
+      //CHECK_EQUAL("r1c2",godleyIcon->table.cell(2,2));
+      //CHECK_EQUAL("r2c1",godleyIcon->table.cell(1,3));
+      //CHECK_EQUAL("r1c1",godleyIcon->table.cell(2,3));
```
highperformancecoder: Actually, that quip about Volkswagen does apply here :). You should check that those mouse operations do not move the first row down, and then add an extra test to check you can move the second row down.

test/testModel.cc
```diff
@@ -1357,16 +1377,18 @@ SUITE(GodleyTableWindow)
...
       keyPress(XK_Tab,"");
-      CHECK_EQUAL(2,selectedCol);
-      keyPress(XK_Right,"");
       CHECK_EQUAL(3,selectedCol);
-      keyPress(XK_Right,"");
+      keyPress(XK_Right,"");     // check wrap around
       CHECK_EQUAL(0,selectedCol);
+      // (1,0) cell no longer selectable
+      //keyPress(XK_Right,"");
+      //CHECK_EQUAL(0,selectedCol);
```
highperformancecoder: I don't understand what this code block is doing. Probably just dead code that should be removed.

test/testModel.cc
```diff
@@ -1357,16 +1377,18 @@ SUITE(GodleyTableWindow)
       CHECK_EQUAL(1,insertIdx);
       keyPress(XK_Left,"");
       CHECK_EQUAL(0,insertIdx);
-      keyPress(XK_Left,"");
-      CHECK_EQUAL(0,selectedCol);
```
highperformancecoder: We should check that the (1,0) cell isn't selected. What should happen - do we wrap (go to the bottom right cell) or do we do nothing? Either option are viable choices, I don't quite remember what was done.

highperformancecoder:
Still need some changes in the unit test code to test the functionality you've just implemented, namely that you can't move or delete the initial condition row.

After that, this will be good to merge.

test/testModel.cc
wdednam: Yup, I _made_ them pass, didn't I.... Won't do that again :)

Okay, I've modified the original test to check that row 1 does not move and also added a test to check that row 2 does. The first two lines of comparison in each test  work (those involving r1c2, r2c2 for row 1, and r2c2 and r3c2 for row 2). But those involving c1 do not (r2c1, r1c1 for row 1, and r2c1 and r3c1). I don't understand why. They are still there, but commented out.

test/testModel.cc
wdednam: I've removed it. I left it there to show how the test has been modified, but you're right that it is now redundant.

test/testModel.cc
wdednam: I now check that cell (1,0) is not selected by checking that selectedRow == SelectedCol == -1 (i.e., deselection). So, essentially nothing is done. If a button is now pressed, selectedRow and selectedCol will be set to the values on lines 632-643 of godleyTableWindow.cc. The latter has also been modified slightly on line 639: selectedRow = 2 when the down arrow is pressed from a deselected state, so that the first modifiable cell in column 0 becomes highlighted (no longer cell(1,0)).

test/testModel.cc
```diff
@@ -1275,16 +1278,26 @@ SUITE(GodleyTableWindow)
...
       CHECK_EQUAL("r2c1",godleyIcon->table.cell(2,3));
       
-      // Row 1 can no longer move down, so the following is no longer needed
+      x=2*ButtonWidget<row>::buttonSpacing+1, y=5+topTableOffset+rowHeight;
+      CHECK_EQUAL(rowWidget, clickType(x,y));
+      mouseDown(x,y);
+      // Row 1 cannot move down
+      CHECK_EQUAL("r2c2",godleyIcon->table.cell(2,2));
+      CHECK_EQUAL("r1c2",godleyIcon->table.cell(1,2));
+      // Not sure why column 1's contents cannot be tested????
```
highperformancecoder: What happens when you do?

PR #69, opened 2019-10-08
Mouse drag and cell edit fixes for ticket 1064

wdednam:
The immutability of row 1 guaranteed now.

However, a few tests on lines 1288,1289, 1298, 1299 of testModel.cc fail with the following output:

testModel.cc:1288:1: error: Failure in mouseButtons: Expected r2c1 but was 
testModel.cc:1289:1: error: Failure in mouseButtons: Expected r1c1 but was 
testModel.cc:1298:1: error: Failure in mouseButtons: Expected r3c1 but was 
testModel.cc:1299:1: error: Failure in mouseButtons: Expected r2c1 but was 

It appears cells in column 1 cannot be clicked on by the mouse, even in rows beneath the initial conditions (row 1)? I cannot pinpoint the reasons from mouseDown(), starting on line 430 in godleyTableWindow.cc, and I have not modified any code between lines 432 and 472.

---
This change is [Reviewable](https://reviewable.io/reviews/highperformancecoder/minsky/69)

### DIFF
--- a/gui-tk/godley.tcl
+++ b/gui-tk/godley.tcl
@@ -230,12 +230,14 @@ proc godleyContext {id x y X Y} {
             $id.insertIdx 0
             $id.selectIdx 0
         }
-        if {[string length [$id.godleyIcon.table.getCell $r $c]] && [$id.godleyIcon.table.getCell $r $c]!=[$id.godleyIcon.table.getCell 1 0]} {   # Cannot Cut or Copy cell(1,0). For ticket 1064
-            .$id.context add command -label "Cut" -command "$id.cut"
+        if {[string length [$id.godleyIcon.table.getCell $r $c]] && ($r!=1 || $c!=0)} {    # Cannot Cut cell(1,0). For ticket 1064
+            .$id.context add command -label "Cut" -command "$id.cut"    
+        }    
+        if {[string length [$id.godleyIcon.table.getCell $r $c]]} {     
             .$id.context add command -label "Copy" -command "$id.copy"
         }
     }
-    if {![catch {clipboard get -type UTF8_STRING}] && [$id.godleyIcon.table.getCell $r $c]!=[$id.godleyIcon.table.getCell 1 0]} {   # Cannot Paste into cell(1,0). For ticket 1064
+    if {![catch {clipboard get -type UTF8_STRING}]  && ($r!=1 || $c!=0)} {   # Cannot Paste into cell(1,0). For ticket 1064
         .$id.context add command -label "Paste" -command "$id.paste"
     }
     tk_popup .$id.context $X $Y
--- a/gui-tk/godley.tcl
+++ b/gui-tk/godley.tcl
@@ -230,12 +230,12 @@ proc godleyContext {id x y X Y} {
             $id.insertIdx 0
             $id.selectIdx 0
         }
-        if [string length [$id.godleyIcon.table.getCell $r $c]] {
+        if {[string length [$id.godleyIcon.table.getCell $r $c]] && [$id.godleyIcon.table.getCell $r $c]!=[$id.godleyIcon.table.getCell 1 0]} {   # Cannot Cut or Copy cell(1,0). For ticket 1064
             .$id.context add command -label "Cut" -command "$id.cut"
             .$id.context add command -label "Copy" -command "$id.copy"
         }
     }
-    if {![catch {clipboard get -type UTF8_STRING}]} {
+    if {![catch {clipboard get -type UTF8_STRING}] && [$id.godleyIcon.table.getCell $r $c]!=[$id.godleyIcon.table.getCell 1 0]} {   # Cannot Paste into cell(1,0). For ticket 1064
         .$id.context add command -label "Paste" -command "$id.paste"
     }
     tk_popup .$id.context $X $Y
--- a/model/godleyTableWindow.cc
+++ b/model/godleyTableWindow.cc
@@ -198,7 +198,7 @@ namespace minsky
               }
             
             CairoSave cs(surface->cairo());
-            if (row!=0 || col!=0)                  
+            if (row!=0 || col!=0)
               {
                 string text=godleyIcon->table.cell(row,col);
                 if (!text.empty())
@@ -220,9 +220,9 @@ namespace minsky
                       }
                     // the active cell renders as bare LaTeX code for
                     // editing, all other cells rendered as LaTeX
-                    if (int(row)!=selectedRow || int(col)!=selectedCol && !godleyIcon->table.initialConditionRow(row))
+                    if (int(row)!=selectedRow || int(col)!=selectedCol)
                       {
-                        if (row>0 && col>0)
+                        if (row>0 && col>0 && !godleyIcon->table.initialConditionRow(row))
                           { // handle DR/CR mode and colouring of text
                             if (fc.coef<0)
                               cairo_set_source_rgb(surface->cairo(),1,0,0);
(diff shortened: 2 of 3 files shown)


gui-tk/godley.tcl
```diff
@@ -230,12 +230,12 @@ proc godleyContext {id x y X Y} {
             $id.insertIdx 0
             $id.selectIdx 0
         }
-        if [string length [$id.godleyIcon.table.getCell $r $c]] {
+        if {[string length [$id.godleyIcon.table.getCell $r $c]] && [$id.godleyIcon.table.getCell $r $c]!=[$id.godleyIcon.table.getCell 1 0]} {   # Cannot Cut or Copy cell(1,0). For ticket 1064
```
highperformancecoder: Why not make the condition && ($r!=1 || $c!=0) ? Currently, you are just testing that the source cell is not "Iniitial  condtitions", which is not the same thing.

gui-tk/godley.tcl
```diff
@@ -230,12 +230,12 @@ proc godleyContext {id x y X Y} {
...
             .$id.context add command -label "Cut" -command "$id.cut"
             .$id.context add command -label "Copy" -command "$id.copy"
         }
     }
-    if {![catch {clipboard get -type UTF8_STRING}]} {
+    if {![catch {clipboard get -type UTF8_STRING}] && [$id.godleyIcon.table.getCell $r $c]!=[$id.godleyIcon.table.getCell 1 0]} {   # Cannot Paste into cell(1,0). For ticket 1064
```
highperformancecoder: Ditto

gui-tk/godley.tcl
highperformancecoder: Also, there is no problem in allowing the user to copy cell (1,0).

model/godleyTableWindow.cc
```diff
@@ -414,12 +414,15 @@ namespace minsky
         selectedCol>=0 && size_t(selectedCol)<godleyIcon->table.cols())
       {
         auto& str=godleyIcon->table.cell(selectedRow,selectedCol);
-        pango.setMarkup(defang(str));
-        int j=0;
-        if (selectedCol>=int(scrollColStart)) j=selectedCol-scrollColStart+1;
-        x-=colLeftMargin[j]+2;
-        x*=zoomFactor;
-        return x>0 && str.length()>0?pango.posToIdx(x)+1: 0;
+        if (str!=godleyIcon->table.cell(1,0))                           // No text index needed for a cell that is immutable. For ticket 1064
```
highperformancecoder: Don't you just want tp avoid (selectedRow,selectedCol)!=(1,0)? We can't assume that (1,0) is the only cell with "Initial Conditions" in it.

model/godleyTableWindow.cc
```diff
@@ -463,10 +466,12 @@ namespace minsky
...
         if (selectedRow>=0 && selectedRow<int(godleyIcon->table.rows()) &&
             selectedCol>=0 && selectedCol<int(godleyIcon->table.cols()))
-          {
-            selectIdx=insertIdx = textIdx(x);
-            savedText=godleyIcon->table.cell(selectedRow, selectedCol);
-          }
+           {
+             selectIdx=insertIdx = textIdx(x);
+             auto& str=godleyIcon->table.cell(selectedRow,selectedCol);     
+             if (str!=godleyIcon->table.cell(1,0))                               // Cannot save text in cell(1,0). For ticket 1064
```
highperformancecoder: Same - you should just test directly on selectedCol, selectedRow.

model/godleyTableWindow.cc
```diff
@@ -487,10 +492,10 @@ namespace minsky
       }
     else if (r>0 && selectedCol==0)
       {
-        if (r!=selectedRow)
+        if (r!=selectedRow && !godleyIcon->table.initialConditionRow(selectedRow) && !godleyIcon->table.initialConditionRow(r))  // Cannot move Intitial Conditions row. For ticket 1064
           godleyIcon->table.moveRow(selectedRow,r-selectedRow);
       }
-    else if ((c!=selectedCol || r!=selectedRow) && c>0 && r>0)
+    else if ((c!=selectedCol || r!=selectedRow) && c>0 && r>0 && !godleyIcon->table.initialConditionRow(selectedRow) && !godleyIcon->table.initialConditionRow(r)) // Cannot swap individual cells in IC row. For ticket 1064
```
highperformancecoder: This is probably a little too restrictive. Just check that (c,r)!=(0,1) and (selectedCol,selectedRow)!=(0,1)

model/godleyTableWindow.cc
```diff
@@ -553,77 +558,78 @@ namespace minsky
...
-              if (insertIdx>=str.length()) insertIdx=str.length();
-              str.insert(insertIdx,utf8);
-              selectIdx=insertIdx+=utf8.length();
-            }
-          else
-            {
-              switch (utf8[0]) // process control characters
+          {			  	  
+            auto& str=table.cell(selectedRow,selectedCol);
+            if (str!=table.cell(1,0))                                 // Cell (1,0) is off-limits. For ticket 1064
```
highperformancecoder: Again - just test selectedCol,selectedRow

gui-tk/godley.tcl
wdednam: Thanks :) Yes, testing for string "Initial Conditions" produces unexpected behaviour in other cells.

gui-tk/godley.tcl
wdednam: Sorted, thanks.

model/godleyTableWindow.cc
wdednam: Very true, using (selectedCol!=0 || selectedRow!=1) instead, thank you.

model/godleyTableWindow.cc
wdednam: Done

model/godleyTableWindow.cc
wdednam: Okay, cells other than (1,0) in the first row can now also be swapped out for a (selectedCol!=0 || selectedRow!=1) condition.

model/godleyTableWindow.cc
wdednam: Also done

test/testModel.cc
```diff
@@ -1285,8 +1285,8 @@ SUITE(GodleyTableWindow)
       CHECK_EQUAL("r2c2",godleyIcon->table.cell(2,2));
       CHECK_EQUAL("r1c2",godleyIcon->table.cell(1,2));
       // Not sure why column 1's contents cannot be tested????
-      CHECK_EQUAL("r2c1",godleyIcon->table.cell(2,1));
-      CHECK_EQUAL("r1c1",godleyIcon->table.cell(1,1));
+      //CHECK_EQUAL("r2c1",godleyIcon->table.cell(2,1));
+      //CHECK_EQUAL("r1c1",godleyIcon->table.cell(1,1));
```
highperformancecoder: Column 1 has been moved to column 3 at line 1274. But I think it is enough of a test to test on column 2, like you do, so this code can be removed.

test/testModel.cc
```diff
@@ -1295,8 +1295,8 @@ SUITE(GodleyTableWindow)
       CHECK_EQUAL("r3c2",godleyIcon->table.cell(2,2));
       CHECK_EQUAL("r2c2",godleyIcon->table.cell(3,2));
       // Not sure why column 1's contents cannot be tested????
-      CHECK_EQUAL("r3c1",godleyIcon->table.cell(2,1));
-      CHECK_EQUAL("r2c1",godleyIcon->table.cell(3,1));
+      //CHECK_EQUAL("r3c1",godleyIcon->table.cell(2,1));
+      //CHECK_EQUAL("r2c1",godleyIcon->table.cell(3,1));
```
highperformancecoder: Ditto.

test/testModel.cc
wdednam: So no wonder they appear empty in the comparison, I never defined initial "r*c3" values for column 3. I might not have been surprised to see a mismatch, but was confused by the fact that the comparison was with empty values. Thanks for clearing that up. :)

test/testModel.cc
wdednam: I've removed lines 1288, 1289, 1298 and 1299 and tidied up the test a bit.

PR #70, opened 2019-10-10
Proposed solution for ticket 1066

wdednam:
Proposed solution for ticket 1066. Makes "NoAssetClass" designation of columns redundant in the Godley table, which I'm not sure is entirely desirable.

---
This change is [Reviewable](https://reviewable.io/reviews/highperformancecoder/minsky/70)

### DIFF
--- a/model/godleyTableWindow.cc
+++ b/model/godleyTableWindow.cc
@@ -481,19 +481,20 @@ namespace minsky
     y/=zoomFactor;
     int c=colX(x), r=rowY(y);
     motionRow=motionCol=-1;
-    if (selectedRow==0)
+    // Cannot swap cell(1,0) with another. For ticket 1064. Also cannot move cells outside an existing Godley table to create new rows or columns. For ticket 1066. 
+    if ((selectedCol==0 && selectedRow==1) || (c==0 && r==1) || size_t(selectedRow)>=(godleyIcon->table.rows()) || size_t(r)>=(godleyIcon->table.rows()) || size_t(c)>=(godleyIcon->table.cols()) || size_t(selectedCol)>=(godleyIcon->table.cols()))
+      return;  
+    else if (selectedRow==0)
       {
-        if (c>0 && selectedCol!=0 && c!=selectedCol)      // Cannot move Intitial Conditions cell from col 0 to one with a "NoAssetClass" designation. For ticket 1064/1066
+        if (c>0 && selectedCol!=0 && c!=selectedCol)      // Disallow moving flow labels column. For ticket 1064/1066
           godleyIcon->table.moveCol(selectedCol,c-selectedCol);
       }
     else if (r>0 && selectedCol==0)
       {
-		// Cannot move Intitial Conditions row. For ticket 1064. Also cannot move last row down, i.e., button widgets have to be used to create new rows, otherwise an error occurs. For ticket 1066.  
-        if (r!=selectedRow && r<int(godleyIcon->table.rows()) && !godleyIcon->table.initialConditionRow(selectedRow) && !godleyIcon->table.initialConditionRow(r))  
+        if (r!=selectedRow && !godleyIcon->table.initialConditionRow(selectedRow) && !godleyIcon->table.initialConditionRow(r))  // Cannot move Intitial Conditions row. For ticket 1064.
           godleyIcon->table.moveRow(selectedRow,r-selectedRow);
-      }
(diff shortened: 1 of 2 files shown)


model/godleyTableWindow.cc
```diff
@@ -483,15 +483,17 @@ namespace minsky
     motionRow=motionCol=-1;
     if (selectedRow==0)
       {
-        if (c>0 && c!=selectedCol)
+        if (c>0 && selectedCol!=0 && c!=selectedCol)      // Cannot move Intitial Conditions cell from col 0 to one with a "NoAssetClass" designation. For ticket 1064/1066
```
highperformancecoder: Comment doesn't apply = (0,0) is not the initial conditions cell, but the one labelled "flows V/stocks ->"
Perhaps change the comment to read "disallow moving flow labels column"

highperformancecoder: But nice catch!

model/godleyTableWindow.cc
```diff
@@ -483,15 +483,17 @@ namespace minsky
...
           godleyIcon->table.moveCol(selectedCol,c-selectedCol);
       }
     else if (r>0 && selectedCol==0)
       {
-        if (r!=selectedRow && !godleyIcon->table.initialConditionRow(selectedRow) && !godleyIcon->table.initialConditionRow(r))  // Cannot move Intitial Conditions row. For ticket 1064
+		// Cannot move Intitial Conditions row. For ticket 1064. Also cannot move last row down, i.e., button widgets have to be used to create new rows, otherwise an error occurs. For ticket 1066.  
+        if (r!=selectedRow && r<int(godleyIcon->table.rows()) && !godleyIcon->table.initialConditionRow(selectedRow) && !godleyIcon->table.initialConditionRow(r))
```
highperformancecoder: I'm puzzled about why you don't bounds check selectedRow as well. I must admit, I can't trigger out of bounds on row in the existing code, so its probably checked elsewhere, but there is no harm in being defensive here.

model/godleyTableWindow.cc
```diff
@@ -483,15 +483,17 @@ namespace minsky
...
+		// Cannot move Intitial Conditions row. For ticket 1064. Also cannot move last row down, i.e., button widgets have to be used to create new rows, otherwise an error occurs. For ticket 1066.  
+        if (r!=selectedRow && r<int(godleyIcon->table.rows()) && !godleyIcon->table.initialConditionRow(selectedRow) && !godleyIcon->table.initialConditionRow(r))  
           godleyIcon->table.moveRow(selectedRow,r-selectedRow);
       }
-    else if ((c!=selectedCol || r!=selectedRow) && c>0 && r>0 && (c!=0 || r!=1) && (selectedCol!=0 || selectedRow!=1)) // Cannot swap cell(1,0) with another. For ticket 1064
+     // Cannot swap cell(1,0) with another. For ticket 1064. Also cannot move cells outside an existing Godley table to create new rows or columns. For ticket 1066.  
+    else if ((c!=selectedCol || r!=selectedRow) && (selectedCol!=0 || selectedRow!=1) && c>0 && r>0 && selectedRow<int(godleyIcon->table.rows()) && r<int(godleyIcon->table.rows()) && c <int(godleyIcon->table.cols()) && selectedCol<int(godleyIcon->table.cols()))
```
highperformancecoder: What happened to (c!=0 || r!=1)?

highperformancecoder: Watch out for the else clause on line 502. That might be triggered now with the initial conditions cell, or out of bounds cell reference. I think it might be best, and cleaner to separate out the no op cases into an early return if statement at the top of the function block:

if (c==0 && r==1 || selectCol==0 && selectedRow==1 || c>=cols() || selectedCol>=cols() || r>=rows() || selectedRow>=rows()) return;

highperformancecoder: We don't need to redraw of there is no change of state.

model/godleyTableWindow.cc
wdednam: Thanks. :)  I've replaced the comment with your suggestion.

model/godleyTableWindow.cc
wdednam: Oops, I should have added that in there. It will be fixed post haste!

model/godleyTableWindow.cc
wdednam: Yikes, should not have removed (c!=0 || r!=1). It fell by the wayside during my attempts to shorten the if-clause.

Moving the no op cases to the top is a great idea, thanks. Will implement it right away.

PR #74, opened 2019-10-25
Proposed solution for ticket 1062

wdednam:
Reduces the amount by which the invisible bounding box extends beyond a canvas item, to make it easier to select the centre handle of a wire from the output port of the item, especially when the wire passes near one of the corners of the invisible bounding box. It is not an ideal solution for operations, for which a circle or a triangle bounding box may be possible solutions. But I believe this will require modifying cairo_base.cc and cairo_base.h under ecolab. In particular, an alternative to [rectangular recording surfaces](https://www.cairographics.org/manual/cairo-Recording-Surfaces.html) could be useful, if it exists. I have looked but have not found one yet.

---
This change is [Reviewable](https://reviewable.io/reviews/highperformancecoder/minsky/74)

### DIFF
--- a/gui-tk/wiring.tcl
+++ b/gui-tk/wiring.tcl
@@ -520,20 +520,34 @@ proc doubleButton {x y} {
         editItem
     }
 }
-
+# for ticket 1062, new hierarchy of context menu access on mouse right click: wires, items and background canvas.
 bind .wiring.canvas <<contextMenu>> {
-    if [getItemAt %x %y] {
+    if [getWireAt %x %y] {
+        wireContextMenu %X %Y  	
+    } elseif [getItemAt %x %y] {
         switch [minsky.canvas.item.classType] {
             GodleyIcon {rightMouseGodley %x %y %X %Y}
             Group {rightMouseGroup %x %y %X %Y}
             default {contextMenu %x %y %X %Y}
-        }
-    } elseif [getWireAt %x %y] {
-        wireContextMenu %X %Y
+		}
     } else {
         canvasContext %x %y %X %Y
     }
 }
+
+#bind .wiring.canvas <<contextMenu>> {
+#    if [getItemAt %x %y] {
+#        switch [minsky.canvas.item.classType] {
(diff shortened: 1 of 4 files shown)


highperformancecoder:
Not much to comment on here, really.

wdednam:
The new proposed solution for this ticket involves accessing the context menu of items and wires on the canvas in reverse order. Lines 524-536 of wiring.tcl now accesses the context menu of wires first before accessing those of items on the canvas. One has to right-click inside the item for its context menu to appear.

gui-tk/wiring.tcl
```diff
@@ -520,21 +520,35 @@ proc doubleButton {x y} {
...
+#            Group {rightMouseGroup %x %y %X %Y}
+#            default {contextMenu %x %y %X %Y}
+#        }
+#    } elseif [getWireAt %x %y] {
+#        wireContextMenu %X %Y
+#    } else {
+#        canvasContext %x %y %X %Y
+#    }
+#}
+
```
highperformancecoder: Should remove the dead code

wdednam: Sorry, I keep forgetting that I don't need to include the previous code for comparison as Github already takes care of that. It's gone now.

wdednam:
I don't understand why the RESTService stuff is still there because I force-pulled the most up-to-date upstream branch of Minsky into my master branch, and when I compare across forks, Github says that there is nothing to compare. However, when I issue "git status" from a terminal, there are "untracked" files to commit, e.g., RESTService/RESTService, which I'm afraid to commit because I think it will add the RESTService stuff back in? Please see attached screencaps.

github_master_branches_even
git_RESTService_to_add_not_sure

highperformancecoder:
You can see exactly what you're contributing by clicking on the "File changed" tab above (https://github.com/highperformancecoder/minsky/pull/74/files).

Basically, you have ended up removing a comment on the RESTService stuff - I couldn't exactly work out what happened, but I suspect a bad merge is at play.

I'm going to have an interesting time when I come to reenable RESTService on the master branch :).

PR #75, opened 2019-11-04
proposed fix for bug 991

wdednam:
Proposed replacement for external function TkMakeBezierCurve(Tk_Canvas,double*,int,int,void*,double*) in wire.cc. Smooth Bezier curves now pass through all the handles on curved wires. Adapted from https://www.stkent.com/2015/07/03/building-smooth-paths-using-bezier-curves.html.

---
This change is [Reviewable](https://reviewable.io/reviews/highperformancecoder/minsky/75)

### DIFF
--- a/model/wire.cc
+++ b/model/wire.cc
@@ -179,7 +179,7 @@ namespace
   
   vector<pair<float,float>> constructTargetVector(int n, const vector<pair<float,float>> knots) {
 	
-	assert(knots.size() % 2 == 0);
+	assert(knots.size() > 2);
 	  
     vector<pair<float,float>> result(n); 
 
@@ -199,7 +199,7 @@ namespace
 
  vector<pair<float,float>> computeControlPoints(int n, const vector<pair<float,float>> knots) {
     
-    assert(knots.size() % 2 == 0);
+    assert(knots.size() > 2);
     
     vector<pair<float,float>> result(2*n); 
     
--- a/model/wire.cc
+++ b/model/wire.cc
@@ -195,26 +195,10 @@ namespace
  *  There are two steps in this calculation. Step 1 involves simultaneously decomposing $A = LU$ and solving $Lk\'=k$ for $k\'$ in a forward sweep, with
  *  $Uc=k\'$ as result. In step 2, $Uc=k\'$ is solved for $c$ in a backward sweep. The calculation goes as follows for a small system:
  *  
- *       \f[
- *       \begin{equation}
- *          \begin{pmatrix} 
- *             b_1 & a_1 & 0  \\
- *             d_1 & b_2 & a_2  \\
- *             0 & d_2 & b_3  \\
- *          \end{pmatrix} 
- *          \begin{pmatrix}
- *             c_1
- *             c_2
- *             c_3
- *          \end{pmatrix}
- *          =
- *          \begin{pmatrix}
- *              k_1
- *              k_2
- *              k_3
- *          \end{pmatrix}
- *       \end{equation} 
- *       \f]         
+ *      $\begin{equation}
+ *         \begin{pmatrix} b_1 & a_1 & 0  \\ d_1 & b_2 & a_2  \\  0 & d_2 & b_3 \end{pmatrix} \begin{pmatrix} c_1 & c_2 & c_3 \end{pmatrix} = \begin{pmatrix} k_1 & k_2 & k_3 \end{pmatrix}
+ *      \end{equation} $
+ *                      
  *        
  *  Forward sweep: row 1:
  *  
@@ -223,7 +207,7 @@ namespace
  *  \f$ c_1+\frac{c_1}{b_1}c_2 = \frac{k_1}{b_1} \f$ 
  *  after dividing through by $b_1$
  * 
- *  Rewrite row 1 as \f$ c_1 + \alpha_1c_2 = k\'_1 \f$, where $\alpha_1=\frac{c_1}{b_1}$ and $k\'_1 = \frac{k_1}{b_1}$
+ *  Rewrite row 1 as $ c_1 + \alpha_1c_2 = k\'_1$, where $\alpha_1=\frac{c_1}{b_1}$ and $k\'_1 = \frac{k_1}{b_1}$
  * 
  *  row 2: 
  * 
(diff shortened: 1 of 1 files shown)


model/wire.cc
```diff
@@ -27,7 +27,7 @@
 using namespace std;
 
 // undocumented internal function in the Tk library
-extern "C" int TkMakeBezierCurve(Tk_Canvas,double*,int,int,void*,double*);
+//extern "C" int TkMakeBezierCurve(Tk_Canvas,double*,int,int,void*,double*);
```
highperformancecoder: remove dead code

model/wire.cc
```diff
@@ -137,12 +137,125 @@ namespace minsky
...
+
+}
+
   void Wire::draw(cairo_t* cairo) const
   {
     auto coords=this->coords();
     if (coords.size()<4 || !visible()) return;
 
-    double angle, lastx, lasty;
+    //double angle, lastx, lasty;
```
highperformancecoder: dead code

model/wire.cc
```diff
@@ -153,25 +266,52 @@ namespace minsky
...
+        //vector<double> points(2*numSteps*(dcoords.size()+1));
         // TODO - find a way of doing this that doesn't involve Tk!
-        int numPoints=
-          TkMakeBezierCurve(0,dcoords.data(),dcoords.size()/2,numSteps,
-                            nullptr,points.data());
+        //int numPoints=
+          //TkMakeBezierCurve(0,dcoords.data(),dcoords.size()/2,numSteps,
+          //                  nullptr,points.data());
+        
+        //cairo_move_to(cairo, points[0], points[1]);
```
highperformancecoder: Dead code

model/wire.cc
```diff
@@ -153,25 +266,52 @@ namespace minsky
...
+        cairo_move_to(cairo, coords[0], coords[1]);
+        //for (int i=2; i<2*numPoints; i+=2)
+        //  cairo_line_to(cairo, points[i], points[i+1]);
+        //int numPoints = coords.size()/2;
```
highperformancecoder: Dead code

model/wire.cc
```diff
@@ -153,25 +266,52 @@ namespace minsky
...
+        }     
+               
         cairo_stroke(cairo);
-        angle=atan2(points[2*numPoints-1]-points[2*numPoints-3], 
-                    points[2*numPoints-2]-points[2*numPoints-4]);
-        lastx=points[2*numPoints-2]; lasty=points[2*numPoints-1];
-      }
+        //angle=atan2(points[2*numPoints-1]-points[2*numPoints-3], 
+        //            points[2*numPoints-2]-points[2*numPoints-4]);
+        //lastx=points[2*numPoints-2]; lasty=points[2*numPoints-1];
```
highperformancecoder: dead code

model/wire.cc
```diff
@@ -260,15 +400,15 @@ namespace minsky
...
-
+    //bool inTriangle(float x0, float y0, float x1, float y1,
+    //                float x2, float y2, float x, float y)
+    //{
+    //  float l1 = (x-x0)*(y2-y0) - (x2-x0)*(y-y0), 
+    //    l2 = (x-x1)*(y0-y1) - (x0-x1)*(y-y1), 
+    //    l3 = (x-x2)*(y1-y2) - (x1-x2)*(y-y2);
+    //  return (l1>0 && l2>0  && l3>0) || (l1<0 && l2<0 && l3<0);
+    //}
+
```
highperformancecoder: dead code

model/wire.cc
```diff
@@ -137,12 +137,125 @@ namespace minsky
...
+      result[i].second = 2.0*(2.0*knots[i].second+(knots[i+1].second));  
+    }
+    
+    result[result.size() - 1].first = 8.0*knots[n-1].first+knots[n].first; 
+    result[result.size() - 1].second = 8.0*knots[n-1].second+knots[n].second;
+
+    return result;
+  }  
+
+ vector<pair<float,float>> computeControlPoints(int n, const vector<pair<float,float>> knots) {
```
highperformancecoder: I feel this method needs more documentation than just a link to a web page, that may not exist at some point in the future. It is possible to embed LaTeX into Doxygen comments, so don't feel restricted to plain ASCII.

The code feels like it can be expressed via some matrix equation, where the matrix is tridiagonal.

highperformancecoder:
Need to fix the build

model/wire.cc
wdednam: Thank you, that is a very helpful tip. Based on it, I have found this very well-documented implementation of Thomas's algorithm for solving the matrix equation Ac = r where A is nearly tridiagonal: https://github.com/CodyJae/Thomas-Algorithm/blob/master/Thomas-Algorithm.cpp. In our case, the matrix is fully tridiagonal, so the implementation will be simpler than at the aforementioned Github page.

Am I allowed to use Doxygen in .cc files? I've only seen it in the header files. For the documentation using LaTeX in Doxygen, https://en.wikibooks.org/wiki/Algorithm_Implementation/Linear_Algebra/Tridiagonal_matrix_algorithm#C++ looks like a good template.

I will implement the above and remove all the dead code.

highperformancecoder: Yes - you can use Doxygen in .cc files. Whether doxygen processes .cc files or not is a configuration option. Embedding LaTeX into comments in a .cc file is a reasonable use case. Cross referencing to other parts of the code base is another reason.

But for comments about interface related matters, it is better to please these in the header files - curiously in a competing documentation processor called gtk, the tradition is to add the documentation to the .c files (its mostly used for C-based code, not C++), rather than .h. They have their reasons, but I find it annoying coming across a header file in such a project, to find no useful comments whatsoever, and of course the C files have all disappeared into an object code library.

highperformancecoder: Re what needs to go here, - we don't need a treatise. I would basically mention that two control points are added in between each handle, and the curves and first and second derivatives matched, second derivatives set to zero at the ends to absorb the extra degrees of freedom, giving rise to to that matrix equation (can copy the one from the blog you refer to). Then Thomas algorithm used to solve the matrix equation. Probably only about 2 paragraphs max, but enough bread crumbs that that algorithm can be reconstructed or analysed without access to the blog article.

highperformancecoder:
Re the Travis error, hopefully this one is resolved after merging from master again.

model/wire.cc
wdednam: Okay, I have replaced the three vector<pair<float>> functions to construct the upper diagonal, diagonal, and lower diagonal vectors with a single vector<vector<float>> matrix function. I tried to implement the single function at the Github page I linked to in my last reply, but could not manage to get it to work. 

The current implementation now calls two anonymous namespace functions in Wire::draw(), which respectively construct the matrix A and target knots k (in Ac = k), and sends c (the coordinates of all the handles), A and k to an anonymous namespace function computeControlPoints(A,c,k). The latter, in turn, applies Thomas' algorithm to Ac =k and returns the new control points. 

All that remains to be done is to add the Doxygen documentation, which I will do today. Thanks for the outline of what it should look like. :)

wdednam:
Great, thanks :)

model/wire.cc
wdednam: I've added the Doxygen documentation in wire.cc and fixed all LaTeX related errors that came up when running doxygen from the source directory. The pdf file generated under the latex directory of doxydoc is mostly empty though...

wdednam:
Test t0026.sh fails during travis bulld but passes on my machine. I don't understand why?
test_t0026_pass

model/wire.cc
```diff
@@ -136,50 +133,41 @@ namespace minsky
...
-      result[i] = 4.0;
-    }
-    
-    result[result.size() - 1] = 7.0; 
-
-    return result;
+// For ticket 991. Construct tridoagonal matrix A which relates control points c and knots k (curved wire handles): Ac = k.
+ vector<vector<float>> constructTriDiag(int length) {
+
+   vector<vector<float>> result(length,vector<float>(length));
```
highperformancecoder: This is not the most optimum way of doing this, but it will do for now.

highperformancecoder:
> Test t0026.sh fails during travis bulld but passes on my machine. I don't understand why?
> test_t0026_pass

For some reason, wire deletion is failing. I trying to check out your branch now to see locally what might be going on.

model/wire.cc
wdednam: Would making it a void function and modifying the matrix in place be a better option?

highperformancecoder:
When checked out on my laptop, it passes too! As you say weird, particularly now that the Travis container is running the same OS as my laptop. Maybe the docker image needs updating...

wdednam:
Thank you. I couldn't figure out why, I tried the test from my older minsky_dev/ directory and the one I pulled from upstream into my master branch, which is where I create new branches for bug fixes.

highperformancecoder:
Whoops! I ran make sure on your master branch, not bugfix-991. So I can reproduce the failure in the deleteKey test. The test may be failing due to having a smaller target now for wire selection. I'll try fixing it now on master branch.

wdednam:
Okay, now I also know the origin of the bug in ticket #1079, which I opened today. It is harder to select the wire when it passes through all the handles, especially near the ends, and new handles often are inserted at the wrong place as a result.

highperformancecoder:
I have just pushed a fix for the failing deleteKey test, as well as
got the LaTeX to work in wire.cc. I found I had to enable MATHJAX
option - as the embedded image generation code didn't work
properly. Actually I'm quite impressed by Math JAX, and might consider
it for LaTeX2HTML work in the future, such as for the Minsky manual.

PS - I think you may have gone overboard with too much algorithmic
detail here, but let's leave that for another day :).

Cheers
--

Principal, High Performance Coders
Visiting Senior Research Fellow        hpcoder@hpcoders.com.au
Economics, Kingston University         http://www.hpcoders.com.au

wdednam:
Thanks Russell, I have pulled the latest version into my master branch.

The html documentation looks very good, but a bit too long yes :). I also struggled to get the \f$ \f$ inline commands to work on all the equations, so I removed them and only left $$, but now see they are essential. Thanks for fixing them.

I have made a few more tweaks to the documentation in wire.cc, because I didn't start all the text at the appropriate distance from the Doxygen starting * at the start of each line, so they appear in code blocks and not as text. I will push a new commit and submit a new pull request for that.

Cheers.

PR #83, opened 2019-11-17
Outer product math bug

halcrichtonstandish:
Outer product section missing explanation for variables as well as displaying maths. Note that the online version of the manual still doesn't display the maths correctly.

---
This change is [Reviewable](https://reviewable.io/reviews/highperformancecoder/minsky/83)

### DIFF
--- a/doc/Reference.tex
+++ b/doc/Reference.tex
@@ -212,8 +212,13 @@ given axis, and $r_1$ and $r_2$ are the ranks of $x$ an $y$ respectively.
 
 
 \subsection{outer product $\otimes$}\label{Operation:outerProduct}
-Computes $z_{i_1,i_2,\ldots,i_{r_1},j_1,\ldots,j_{r_2}} =
-  x_{i_1,,i_2,\ldots,i_{r_1}}y_{j_1,\ldots,j_{r_2}}$.
+Computes 
+\begin{displaymath}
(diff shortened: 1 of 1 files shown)


doc/Reference.tex
```diff
@@ -212,8 +212,13 @@ \subsection{inner product $\cdot$}\label{Operation:innerProduct}
...
 
 \subsection{outer product $\otimes$}\label{Operation:outerProduct}
-Computes $z_{i_1,i_2,\ldots,i_{r_1},j_1,\ldots,j_{r_2}} =
-  x_{i_1,,i_2,\ldots,i_{r_1}}y_{j_1,\ldots,j_{r_2}}$.
+Computes 
+\begin{displaymath}
+$z_{i_1,i_2,\ldots,i_{r_1},j_1,\ldots,j_{r_2}} =
+x_{i_1,,i_2,\ldots,i_{r_1}}y_{j_1,\ldots,j_{r_2}}$.
+\end{displaymath}
+where $a$ is the
```
highperformancecoder: There is no a in the above formula.

highperformancecoder: I've noticed a problem with the inner product formula. Not sure when that happened, but it looks like I made the commit :(.

highperformancecoder:
outerp

This the existing file  looks like when rendered in seamonkey - I can't see a problem here. Its possible that the LaTeX2html installation on your laptop is not correct - entirely possible, as it was quite difficult IIRC.

PR #86, opened 2019-11-23
proposed fix for bugs 1079 and 1085

wdednam:
Uses same bits of code from wire.cc that is used to insert handles in curved wires to now also select them.

---
This change is [Reviewable](https://reviewable.io/reviews/highperformancecoder/minsky/86)

### DIFF
--- a/model/wire.cc
+++ b/model/wire.cc
@@ -464,10 +464,40 @@ namespace
     assert(c.size()>=4);
     if (c.size()==4)
       return segNear(c[0],c[1],c[2],c[3],x,y);
-    else
+    else {
+	  // fixes for tickets 1079 and 1085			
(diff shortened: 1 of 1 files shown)


highperformancecoder:
So basically eliminating near from the ends of the wire. Sounds reasonable.

wdednam:
It's a little bit better, but not quite there yet. I reverted again locally last night but forgot to push the second revert, sorry. I need to think about this some more, because it is not sufficient to access the coords vector incrementally from one direction to see whether the mouse cursor passes near an arbitrary segment. And as we've seen, stepping through contiguous segments until we get a "return true" leads to over selection, which is what causes this bug in the first place. Going to see if I can't write an any_of() with an appropriate lambda to deal with this.

PR #88, opened 2019-11-24
Bugfix 1085

wdednam:
Was necessary to regress test deleteKey.sh to a previous version too in order to pass Travis build.

---
This change is [Reviewable](https://reviewable.io/reviews/highperformancecoder/minsky/88)

### DIFF
--- a/test/00/deleteKey.sh
+++ b/test/00/deleteKey.sh
@@ -37,10 +37,11 @@ proc afterMinskyStarted {} {
   set numWires [model.numWires]
   focus .wiring.canvas
 
-  set w [minsky.model.wires.@elem 0]
-  set x {[\$w.coords.@elem 0]}
-  set y {[\$w.coords.@elem 1]}
-  deleteKey \$x \$y
+#  set w [minsky.model.wires.@elem 1]
+#  set x {[\$w.coords.@elem 0]}
+#  set y {[\$w.coords.@elem 1]}
+#  deleteKey \$x \$y
+  deleteKey 404 265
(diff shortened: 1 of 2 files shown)


test/00/deleteKey.sh
```diff
@@ -37,10 +37,11 @@ proc afterMinskyStarted {} {
...
 
-  set w [minsky.model.wires.@elem 0]
-  set x {[\$w.coords.@elem 0]}
-  set y {[\$w.coords.@elem 1]}
-  deleteKey \$x \$y
+#  set w [minsky.model.wires.@elem 1]
+#  set x {[\$w.coords.@elem 0]}
+#  set y {[\$w.coords.@elem 1]}
+#  deleteKey \$x \$y
+  deleteKey 404 265
```
highperformancecoder: Yes - but this makes the test more fragile again. What about reinstating the previous TCL code, but setting x and y to the midpoints:
set x [expr 0.5*[\$w.coords.@elem 0]+[\$w.coords.@elem 2]]
set Y [expr 0.5*[\$w.coords.@elem 1]+[\$w.coords.@elem 3]]

wdednam: Thanks for the suggestion, it hadn't occurred to me. I've tried it for element 0 and some others, but it has not yet worked for me. If I succeed I will put in a commit. I don't understand why curly braces are required around these expressions?

  set x {[expr 0.5*[[\$w.coords.@elem 0]+[\$w.coords.@elem 2]]]}
  set y {[expr 0.5*[[\$w.coords.@elem 1]+[\$w.coords.@elem 3]]]}

They aren't in other tcl files where coordinate variables are set.

PR #91, opened 2019-12-01
New parameter and variable tabs. Still a WIP

wdednam:
Added non-functional variable and parameter tabs. Both still very much WIPs.

---
This change is [Reviewable](https://reviewable.io/reviews/highperformancecoder/minsky/91)

### DIFF
--- a/model/parameterSheet.cc
+++ b/model/parameterSheet.cc
@@ -1,0 +1,65 @@
+/*
+  @copyright Steve Keen 2018
+  @author Russell Standish
+  This file is part of Minsky.
+
+  Minsky is free software: you can redistribute it and/or modify it
+  under the terms of the GNU General Public License as published by
+  the Free Software Foundation, either version 3 of the License, or
+  (at your option) any later version.
+
+  Minsky is distributed in the hope that it will be useful,
+  but WITHOUT ANY WARRANTY; without even the implied warranty of
+  MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE.  See the
+  GNU General Public License for more details.
+
+  You should have received a copy of the GNU General Public License
+  along with Minsky.  If not, see <http://www.gnu.org/licenses/>.
+*/
+#include "parameterSheet.h"
+#include "latexMarkup.h"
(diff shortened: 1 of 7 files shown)


model/parameterSheet.cc
```diff
@@ -0,0 +1,65 @@
+/*
+  @copyright Steve Keen 2018
+  @author Russell Standish
```
highperformancecoder: Should change the @author tag :)

model/variableSheet.cc
```diff
@@ -0,0 +1,67 @@
...
+using ecolab::cairo::Surface;
+using ecolab::Pango;
+using ecolab::cairo::CairoSave;
+
+namespace minsky
+{
+	
+namespace
+{    
+  struct CroppedPango: public Pango
```
highperformancecoder: This class is a candidate for being moved into a common header file

model/variableSheet.h
```diff
@@ -0,0 +1,43 @@
...
+{
+	 
+  class VariableSheet: public ecolab::CairoSurface
+  {
+  public:
+    double xoffs=80;  
+    double rowHeight=0;
+    double colWidth=50;  
+    void redraw(int, int, int width, int height) override;
+    void requestRedraw() {if (surface.get()) surface->requestRedraw();}
```
highperformancecoder: Maybe this bog standard requestRedraw (which exposes this functionality to TCL) should be provided by ecolab::CairoSurface. It can be overridden when needed (PlotWidget needs to do something different)

model/parameterSheet.cc
wdednam: Thanks, will change it :)

model/variableSheet.cc
wdednam: True, it is repeated in three different headers. Maybe I can just put it in minsky.h like EquationDisplay?

model/variableSheet.h
wdednam: We use  ecolab::CairoSurface in redraw() so it makes sense to use it in requestRedraw() too. Okay, understood.

PR #92, opened 2019-12-06
Initial commit for feature 47. Still a WIP

wdednam:
Seems to load data of most CSVs without running out of memory. Re-exporting to CSV after initial import picks up values from original CSV but format not correctly reproduced. Have not yet tested how the code behaves under operations.

---
This change is [Reviewable](https://reviewable.io/reviews/highperformancecoder/minsky/92)

### DIFF
--- a/engine/CSVParser.cc
+++ b/engine/CSVParser.cc
@@ -342,16 +342,15 @@ namespace minsky
     string buf;
     typedef vector<string> Key;
     map<Key,double> tmpData;
-    multimap<Key,double> tmpAll; 
     map<Key,int> tmpCnt;
     vector<map<string,size_t>> dimLabels(spec.dimensionCols.size());
     bool tabularFormat=false;
     vector<XVector> xVector;
     vector<string> horizontalLabels;
-	       
+
     for (size_t i=0; i<spec.nColAxes(); ++i)
       if (spec.dimensionCols.count(i))
-        xVector.push_back(i<spec.dimensionNames.size()? spec.dimensionNames[i]: "dim"+str(i));
+        xVector.emplace_back(i<spec.dimensionNames.size()? spec.dimensionNames[i]: "dim"+str(i));
 
     try
       {
@@ -403,20 +402,7 @@ namespace minsky
                       if (c==spec.decSeparator)
                         s+='.';
                       else if (!isspace(c) && c!='.' && c!=',')
-                        s+=c;                    
-                    
-                    // for feature 47, stores all values, duplicates included in a temporary map
-                    auto j=tmpAll.find(key);
-                    try
-                     {
-                      double vv=stod(s);    
-                                          
-                      if (j==tmpAll.end()) tmpAll.emplace(pair<Key,double>(key,vv));  
-				     }
-				     catch (...)
-                      {
-                        tmpAll.insert(pair<Key,double>(key,spec.missingValue));
-                      }
+                        s+=c;
 
                     auto i=tmpData.find(key);
                     try
@@ -424,7 +410,7 @@ namespace minsky
                         double v=stod(s);
                         if (i==tmpData.end())
                           tmpData.emplace(key,v);
-                        else	
+                        else
                           switch (spec.duplicateKeyAction)
                             {
                             case DataSpec::throwException:
@@ -461,85 +447,26 @@ namespace minsky
                   }
               }
           }
-                  
-        v.sparsityRatio =  static_cast<double>(1.0-static_cast<double>(tmpData.size())/v.numDenseElements()); 
-		
-		// for feature 47
-		if (v.sparsityRatio <= 0.5) 
-	    {
-          v.setXVector(xVector);          
-          if (!cminsky().checkMemAllocation(v.numDenseElements()*sizeof(double)))
-            throw runtime_error("memory threshold exceeded");
-		  v.index.clear();	            
-          // stash the data into vv tensorInit field
-          v.tensorInit.data.clear();
-          v.tensorInit.data.resize(v.numDenseElements(), spec.missingValue);  
-          auto dims=v.tensorInit.dims=v.dims();    
-          for (auto& i: tmpData)
-            {
-              size_t idx=0;
-              assert (dims.size()==i.first.size());
-              assert(dimLabels.size()==dims.size());
-              for (int j=dims.size()-1; j>=0; --j)
-                {
-                  assert(dimLabels[j].count(i.first[j]));
-                  idx = (idx*dims[j]) + dimLabels[j][i.first[j]];
-                }
-              v.tensorInit.data[idx]=i.second;  
-            }
-        }    
-        else 
-	    {	
-		  v.numSparseElements=tmpData.size();	
-          v.setXVector(xVector);          
-          if (!cminsky().checkMemAllocation(v.numSparseElements*sizeof(double)))
-            throw runtime_error("memory threshold exceeded");		  
-		  v.index.clear();			
-		  v.tensorInit.data.clear();	
-		  v.tensorInit.data.resize(v.numSparseElements, spec.missingValue); 	
-		  vector<double> tempData;
-          auto dims=v.tensorInit.dims=v.dims();    
-          for (auto& i: tmpData)
-            {
-              size_t idx=0;
-              assert (dims.size()==i.first.size());
-              assert(dimLabels.size()==dims.size());
-              for (int j=dims.size()-1; j>=0; --j)
-                {
-                  assert(dimLabels[j].count(i.first[j]));
-                  idx = (idx*dims[j]) + dimLabels[j][i.first[j]];
-                }
-               if (!isnan(i.second)) {
-                 v.index.push_back(idx);
-                 tempData.push_back(i.second);
-               }                   
-            }
-           
-            // for feature 47           
-            for (auto& i: tmpAll)
-            {
-			  size_t hIdx=0;	
-              assert (dims.size()==i.first.size());
-              assert(dimLabels.size()==dims.size());
-              for (int j=dims.size()-1; j>=0; --j)
-                {
-                  assert(dimLabels[j].count(i.first[j]));
-                  hIdx = (hIdx*dims[j]) + dimLabels[j][i.first[j]];
-                }                   
-                auto idx = find(v.index.begin(),v.index.end(),hIdx);   
-                // For feature 47. Stores only non-Nan elements from data in tensorInit.data                 
-                if (idx != v.index.end() && *idx<v.tensorInit.data.size()) v.tensorInit.data[*idx]=tempData[*idx];  
-             }               
-            
-            tempData.clear();   
-            
-            assert(tempData.size()*v.tensorInit.data.size()==0);
-            						  
-            if (v.tensorInit.data.size()!=v.numSparseElements) throw std::runtime_error("the number of non-NaN elements should be "+str(v.numSparseElements)+" but is "+str(v.tensorInit.data.size())+ " in a hypercube of "+str(v.numDenseElements())+" elements");
-            v.setXVector(xVector);
-            
-     	}          
-
+  
+        v.setXVector(xVector);
+        if (!cminsky().checkMemAllocation(v.numElements()*sizeof(double)))
+          throw runtime_error("memory threshold exeeded");
+        // stash the data into vv tensorInit field
+        v.tensorInit.data.clear();
+        v.tensorInit.data.resize(v.numElements(), spec.missingValue);
+        auto dims=v.tensorInit.dims=v.dims();    
+        for (auto& i: tmpData)
+          {
+            size_t idx=0;
+            assert (dims.size()==i.first.size());
+            assert(dimLabels.size()==dims.size());
+            for (int j=dims.size()-1; j>=0; --j)
+              {
+                assert(dimLabels[j].count(i.first[j]));
+                idx = (idx*dims[j]) + dimLabels[j][i.first[j]];
+              }
+            v.tensorInit.data[idx]=i.second;
+          }
       }
     catch (const std::bad_alloc&)
       { // replace with a more user friendly error message
--- a/engine/CSVParser.cc
+++ b/engine/CSVParser.cc
@@ -404,19 +404,6 @@ namespace minsky
                         s+='.';
                       else if (!isspace(c) && c!='.' && c!=',')
                         s+=c;                    
-                    
-                    // for feature 47, stores all values, duplicates included in a temporary map
-                    auto j=tmpAll.find(key);
-                    try
-                     {
-                      double vv=stod(s);    
-                                          
-                      if (j==tmpAll.end()) tmpAll.emplace(pair<Key,double>(key,vv));  
-				     }
-				     catch (...)
-                      {
-                        tmpAll.insert(pair<Key,double>(key,spec.missingValue));
-                      }
 
                     auto i=tmpData.find(key);
                     try
@@ -462,15 +449,18 @@ namespace minsky
               }
           }
                   
-        v.sparsityRatio =  static_cast<double>(1.0-static_cast<double>(tmpData.size())/v.numDenseElements()); 
+        size_t numHyperCubeElems=1;
+        for (auto& i : xVector) numHyperCubeElems*=i.size();
+                           
+        double sparsityRatio =  static_cast<double>(1.0-static_cast<double>(tmpData.size())/numHyperCubeElems); 
 		
 		// for feature 47
-		if (v.sparsityRatio <= 0.5) 
+		if (sparsityRatio <= 0.5) 
 	    {
+		  v.index.clear();	
           v.setXVector(xVector);          
           if (!cminsky().checkMemAllocation(v.numDenseElements()*sizeof(double)))
-            throw runtime_error("memory threshold exceeded");
-		  v.index.clear();	            
+            throw runtime_error("memory threshold exceeded");            
           // stash the data into vv tensorInit field
           v.tensorInit.data.clear();
           v.tensorInit.data.resize(v.numDenseElements(), spec.missingValue);  
@@ -490,15 +480,12 @@ namespace minsky
         }    
         else 
 	    {	
-		  v.numSparseElements=tmpData.size();	
+          v.index.clear();			
           v.setXVector(xVector);          
-          if (!cminsky().checkMemAllocation(v.numSparseElements*sizeof(double)))
-            throw runtime_error("memory threshold exceeded");		  
-		  v.index.clear();			
-		  v.tensorInit.data.clear();	
-		  v.tensorInit.data.resize(v.numSparseElements, spec.missingValue); 	
-		  vector<double> tempData;
-          auto dims=v.tensorInit.dims=v.dims();    
+          if (!cminsky().checkMemAllocation(tmpData.size()*sizeof(double)))
+            throw runtime_error("memory threshold exceeded");	  	  		
+          auto dims=v.tensorInit.dims=v.dims();	
+          v.tensorInit.data.clear();   
           for (auto& i: tmpData)
             {
               size_t idx=0;
@@ -509,36 +496,13 @@ namespace minsky
                   assert(dimLabels[j].count(i.first[j]));
                   idx = (idx*dims[j]) + dimLabels[j][i.first[j]];
                 }
-               if (!isnan(i.second)) {
-                 v.index.push_back(idx);
-                 tempData.push_back(i.second);
-               }                   
+               if (!isnan(i.second)) {  
+				  v.index.push_back(idx);
+                  v.tensorInit.data.push_back(i.second);
+			   }
             }
-           
-            // for feature 47           
-            for (auto& i: tmpAll)
-            {
-			  size_t hIdx=0;	
-              assert (dims.size()==i.first.size());
-              assert(dimLabels.size()==dims.size());
-              for (int j=dims.size()-1; j>=0; --j)
-                {
-                  assert(dimLabels[j].count(i.first[j]));
-                  hIdx = (hIdx*dims[j]) + dimLabels[j][i.first[j]];
-                }                   
-                auto idx = find(v.index.begin(),v.index.end(),hIdx);   
-                // For feature 47. Stores only non-Nan elements from data in tensorInit.data                 
-                if (idx != v.index.end() && *idx<v.tensorInit.data.size()) v.tensorInit.data[*idx]=tempData[*idx];  
-             }               
-            
-            tempData.clear();   
-            
-            assert(tempData.size()*v.tensorInit.data.size()==0);
-            						  
-            if (v.tensorInit.data.size()!=v.numSparseElements) throw std::runtime_error("the number of non-NaN elements should be "+str(v.numSparseElements)+" but is "+str(v.tensorInit.data.size())+ " in a hypercube of "+str(v.numDenseElements())+" elements");
-            v.setXVector(xVector);
-            
-     	}          
+            assert(v.index.size()==v.tensorInit.data.size());
+     	}                 
 
       }
     catch (const std::bad_alloc&)
--- a/engine/CSVParser.cc
+++ b/engine/CSVParser.cc
@@ -127,6 +127,7 @@ void DataSpec::setDataArea(size_t row, size_t col)
   dimensionCols.erase(dimensionCols.lower_bound(nColAxes()), dimensionCols.end());
 }
 
+
 template <class TokenizerFunction>
 void DataSpec::givenTFguessRemainder(std::istream& input, const TokenizerFunction& tf)
 {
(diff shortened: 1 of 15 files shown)


engine/CSVParser.cc
```diff
@@ -402,15 +403,28 @@ namespace minsky
...
-                        s+=c;
+                        s+=c;                    
+                    
+                    // for feature 47, stores all values, duplicates included in a temporary map
+                    auto j=tmpAll.find(key);
+                    try
+                     {
+                      double vv=stod(s);    
+                                          
+                      if (j==tmpAll.end()) tmpAll.emplace(pair<Key,double>(key,vv));
```
highperformancecoder: I don't understand this code. If you are only inserting data if the key doesn't already exist, then why use a multimap?

engine/CSVParser.cc
```diff
@@ -447,26 +461,85 @@ namespace minsky
...
+                  idx = (idx*dims[j]) + dimLabels[j][i.first[j]];
+                }
+              v.tensorInit.data[idx]=i.second;  
+            }
+        }    
+        else 
+	    {	
+		  v.numSparseElements=tmpData.size();	
+          v.setXVector(xVector);          
+          if (!cminsky().checkMemAllocation(v.numSparseElements*sizeof(double)))
```
highperformancecoder: Argument probably needs to be doubled to take into account memory occupied by index

engine/CSVParser.cc
```diff
@@ -447,26 +461,85 @@ namespace minsky
...
+              assert(dimLabels.size()==dims.size());
+              for (int j=dims.size()-1; j>=0; --j)
+                {
+                  assert(dimLabels[j].count(i.first[j]));
+                  idx = (idx*dims[j]) + dimLabels[j][i.first[j]];
+                }
+               if (!isnan(i.second)) {
+                 v.index.push_back(idx);
+                 tempData.push_back(i.second);
+               }
```
highperformancecoder: OK so far. there is some common code with the true part of the conditional that could be refactored, but lets leave that until we've got things working.

engine/CSVParser.cc
```diff
@@ -447,26 +461,85 @@ namespace minsky
...
+        }    
+        else 
+	    {	
+		  v.numSparseElements=tmpData.size();	
+          v.setXVector(xVector);          
+          if (!cminsky().checkMemAllocation(v.numSparseElements*sizeof(double)))
+            throw runtime_error("memory threshold exceeded");		  
+		  v.index.clear();			
+		  v.tensorInit.data.clear();	
+		  v.tensorInit.data.resize(v.numSparseElements, spec.missingValue);
```
highperformancecoder: No need to allocate an initialise the memory here. Just push_back directly into v.tensorInit.data on line 514.

engine/CSVParser.cc
```diff
@@ -447,26 +461,85 @@ namespace minsky
...
+                  idx = (idx*dims[j]) + dimLabels[j][i.first[j]];
+                }
+               if (!isnan(i.second)) {
+                 v.index.push_back(idx);
+                 tempData.push_back(i.second);
+               }                   
+            }
+           
+            // for feature 47           
+            for (auto& i: tmpAll)
```
highperformancecoder: What is the purpose of this loop?

engine/CSVParser.cc
```diff
@@ -447,26 +461,85 @@ namespace minsky
...
+                  hIdx = (hIdx*dims[j]) + dimLabels[j][i.first[j]];
+                }                   
+                auto idx = find(v.index.begin(),v.index.end(),hIdx);   
+                // For feature 47. Stores only non-Nan elements from data in tensorInit.data                 
+                if (idx != v.index.end() && *idx<v.tensorInit.data.size()) v.tensorInit.data[*idx]=tempData[*idx];  
+             }               
+            
+            tempData.clear();   
+            
+            assert(tempData.size()*v.tensorInit.data.size()==0);
```
highperformancecoder: This is guaranteed by the previous statement.

engine/CSVParser.cc
```diff
@@ -447,26 +461,85 @@ namespace minsky
...
+            						  
+            if (v.tensorInit.data.size()!=v.numSparseElements) throw std::runtime_error("the number of non-NaN elements should be "+str(v.numSparseElements)+" but is "+str(v.tensorInit.data.size())+ " in a hypercube of "+str(v.numDenseElements())+" elements");
```
highperformancecoder: This is also guaranteed by the code as you have it - the exception will never be thrown.

engine/evalOp.cc
```diff
@@ -831,11 +849,12 @@ namespace minsky
                 if (&to==&from1) checkAllEntriesPresent(to.xVector, from2.xVector);
                 if (&to==&from2) checkAllEntriesPresent(to.xVector, from1.xVector);
             
-                if (from1.numElements()==1 && from2.numElements()==1)
+                // For feature 47            
+                if ((from1.numDenseElements()==1 && from2.numDenseElements()==1) || (from1.numSparseElements==1 && from2.numSparseElements==1))
```
highperformancecoder: You can pretty much guarantee that the single element tensor will be stored densely. Add an assertion to the fact.

engine/evalOp.cc
```diff
@@ -959,8 +978,14 @@ namespace minsky
...
                   to.setXVector(from1.xVector);
                 if (to.xVector==from1.xVector)
-                    for (size_t i=0; i<from1.numElements(); ++i)
-                      t->in1.push_back(i+from1.idx());
+                  { 
+					// For feature 47  
+                    if (from1.index.empty())   
+                      for (size_t i=0; i<from1.numDenseElements(); ++i)                          
+                        t->in1.push_back(i+from1.idx());
+                    else for (size_t i=0; i<from1.numSparseElements; ++i)
```
highperformancecoder: This loop can be written more neatly using C++-11 for loops:
for (size_t i: from.index)  t->in1.push_back(i+from1.idx())

Also it highlights that numSparseElements is just equal to index.size() - you should read up on the dangers of unnormalised data

engine/evalOp.cc
```diff
@@ -975,8 +1000,15 @@ namespace minsky
...
-                for (size_t i=0; i<from1.numElements(); ++i)
-                  t->in1.push_back(i*stride+from1.idx());
+                      //for (size_t i=0; i<from1.numElements(); ++i)
+                      //  t->in1.push_back(i*stride+from1.idx());
+                      
+                      // For feature 47
+                     if (from1.index.empty())   
+                       for (size_t i=0; i<from1.numDenseElements(); ++i)                          
+                         t->in1.push_back(i*stride+from1.idx());
+                     else for (size_t i=0; i<from1.numSparseElements; ++i)
```
highperformancecoder: ditto

engine/evalOp.cc
```diff
@@ -985,11 +1017,23 @@ namespace minsky
...
+                    
+                    //for (size_t i=0; i < targetDims[1]; ++i)     // from1.numElements()
+                    //  t->in1.push_back(i+from1.idx());
+                    
+                    
+                    // For feature 47
+                    if (from1.index.empty())   
+                      for (size_t i=0; i<from1.numDenseElements(); ++i)                          
+                        t->in1.push_back(i+from1.idx());
+                    else for (size_t i=0; i<from1.numSparseElements; ++i)
```
highperformancecoder: ditto re C++11 for loop

engine/variableValue.cc
```diff
@@ -32,14 +32,17 @@ namespace minsky
...
-    if (realloc) allocValue();
-    memcpy(&valRef(), &x.data[0], x.data.size()*sizeof(x.data[0]));
-    return *this;
+	 bool realloc;
+	 // For feature 47
+     if (numSparseElements==0) realloc=numDenseElements()!=x.data.size();  
+     else realloc=numSparseElements!=x.data.size();
+     if (dims()!=x.dims) dims(x.dims);
+     if (realloc) allocValue(); 
+     memcpy(&valRef(), &x.data[0], x.data.size()*sizeof(x.data[0]));
```
highperformancecoder: You also need to copy the index vector.

engine/variableValue.cc
```diff
@@ -51,20 +54,25 @@ namespace minsky
       case constant:
       case parameter:
         m_idx=ValueVector::flowVars.size();
-        ValueVector::flowVars.resize
-          (ValueVector::flowVars.size()+numElements());
+        //ValueVector::flowVars.resize(ValueVector::flowVars.size()+numElements());
+        // For feature 47
+		if (index.empty()) ValueVector::flowVars.resize(ValueVector::flowVars.size()+numDenseElements());
+		else ValueVector::flowVars.resize(ValueVector::flowVars.size()+numSparseElements);
```
highperformancecoder: Could be improved by have a dedicated method that returns the number actual finite elements. Could be implemented as:

return index.size()? index.size(): numDenseElements();

engine/variableValue.h
```diff
@@ -139,19 +141,28 @@ namespace minsky
       setXVector(std::move(xv));
       return d;
     }
-    size_t numElements() const {
+    // For feature 47
+    std::vector<size_t> index;
+    
+    double sparsityRatio=0.0;
```
highperformancecoder: Do we need this data member?

engine/variableValue.h
```diff
@@ -139,19 +141,28 @@ namespace minsky
...
-
-    template <class T>
-    void setXVector_(T x) {
-      size_t prevNumElems=numElements();
-      m_xVector=x;
-      if (idx()==-1 || prevNumElems<numElements())
-        allocValue();
-    }
+    
+    size_t numSparseElements = 0.0;
```
highperformancecoder: As mentioned, it would be better to carry this information as index.size(). You can add an accessor method if desired for stylistic reasons, though.

engine/variableValue.h
```diff
@@ -139,19 +141,28 @@ namespace minsky
...
+    
+    template <class T>                                            
+    void setXVector_(T x) {    
+      size_t prevNumElems;
+      if (sparsityRatio <= 0.5) prevNumElems=numDenseElements();
+      else prevNumElems=numSparseElements;    
+      m_xVector=x;    
+      if (idx()==-1 || (prevNumElems<numDenseElements() || prevNumElems<numSparseElements))    
+        allocValue();
```
highperformancecoder: I'm not sure about this. Now that the number of elements allocated is uncoupled from xVector, it may no longer make sense allocValue here. Alternatively, we assume that the index vector has been set correctly. This entails that the index vector needs to be encapsulated in the same way as xvector, such that setting the index vector also forces an allocValue if the index vector changes length.

model/plotWidget.cc
```diff
@@ -372,7 +372,7 @@ namespace minsky
   void PlotWidget::addPlotPt(double t)
   {
     for (size_t pen=0; pen<2*numLines; ++pen)
-      if (yvars[pen].numElements()==1 && yvars[pen].idx()>=0)
+      if ((yvars[pen].numDenseElements()==1 || yvars[pen].numSparseElements==1) && yvars[pen].idx()>=0)
```
highperformancecoder: In this client of VariableValue's code, we should be unaware of whether the storage is dense or sparse. In this case, a method that returns the actual number of elements suffices.

model/plotWidget.cc
```diff
@@ -488,21 +489,39 @@ namespace minsky
...
+                for (size_t i=1; i<yv.xVector.size(); ++i)
+                  {
+                    label+=str(yv.xVector[i][(j/stride)%d[i]])+" ";
+                    stride*=d[i];
+                  }
+                labelPen(extraPen,label);
+                extraPen++;
+              }
+           else
+            for (size_t j=0 /*d[0]*/; j<std::min(size_t(10)*d[0], yv.numSparseElements); j+=d[0])
```
highperformancecoder: In this branch, we need to unroll the index vector entry to pick up the pen label.

But we should aim to make this consideration as invisible as possible in this client code, so think about how to write a method that returns the hypercube index for a given offset into the actual data.

model/ravelWrap.cc
```diff
@@ -462,7 +462,10 @@ namespace minsky
           {
             vector<size_t> outHandles(dims.size());
             ravel_outputHandleIds(ravel, &outHandles[0]);
-            size_t prevNumElem=v.numElements();
+            size_t prevNumElem;
```
highperformancecoder: Once we get the VariableValue interface right, much of this code can be simplified such as to not care whether the vv is dense or sparse.

model/variable.cc
```diff
@@ -411,7 +412,8 @@ void VariableBase::draw(cairo_t *cairo) const
   if (VariableValue::isValueId(valueId()))
     vv=minsky::cminsky().variableValues[valueId()];
   
-  if (type()!=constant && !ioVar() && vv.numElements()==1 )
+  // For feature 47
+  if (type()!=constant && !ioVar() && (vv.numDenseElements()==1 || vv.numSparseElements==1) )
```
highperformancecoder: I think we can assume that numSparseElements==1 => numDenseElements()==1, ie basically a scalar variable.

highperformancecoder:
Please fix Travis errors, and also think about adding some test cases that exercise the sparse representation code.

engine/CSVParser.cc
wdednam: Yes, that was unnecessary. I've removed it. So is the use of tmpAll. tmpData contains the data that we want to import.

engine/CSVParser.cc
wdednam: Makes sense. I didn't double it because index is a separate vector. But it still gets loaded into memory, so that doesn't matter. Thanks.

engine/CSVParser.cc
wdednam: Thanks. I directly load i.second into v.tensorInit.data now, since there is no reason not to use the same data structure for the sparse data. The only difference is that its size is v.numSparseElements (or better yet v.index.size()) instead of v.numDenseElements(). Its indices run from 0 to v.index.size()-1, whereas v.index contains the original flattened/serialized hypercube indices which, as I understand it, will be required anywhere in the code the hypercube axes (dimensions) are looped over and the data is used in that context.

engine/CSVParser.cc
wdednam: Thanks. I've removed it. I have to look carefully at the size and capacity of this and the index vector. I get segmentation faults when I try to import certain CSV files.

engine/CSVParser.cc
wdednam: I thought I had to run through all the serialized hypercube indices (including those of missing data) to pick out the onse corresponding to actual data. Of course this is not necessary because the loop over tmpData already does this. I've removed it.

engine/CSVParser.cc
wdednam: Thanks, I've removed it.

engine/CSVParser.cc
wdednam: I have removed this code. However, I have had problems (get segmentation faults) with CSV files that on import result in index vectors whose index.capacity() != index.size(), or rather when tensorInit.data's size and capacity are not the same (in number) as the index's.

engine/evalOp.cc
wdednam: I think this is what I'm really struggling with: that the normalized data structure in the xVector is not preserved upon import of sparse data files the way it is done at present. Storing the data in sparse format is easy. I can use two vectors, a vector of pairs, a map, etc. However, the really tricky part is serializing the data in the valuevectors for use in the Runge Kutta routines. The way I do it at present does not preserve the normalized data structure in the xVector.

engine/variableValue.cc
wdednam: Okay, I thought that &valRef() only needed index.size(). Do I thus have to replace m_idx in 

          if (size_t(m_idx+index.size())<=ValueVector::stockVars.size())
             return ValueVector::stockVars[m_idx];  

by index[0]?

engine/variableValue.cc
wdednam: Thanks, I will implement this method today.

engine/variableValue.h
wdednam: No, we don't. I can just have a local variable in CSVParser.cc to decide whether or not the data is sparse. I will remove it.

engine/variableValue.cc
wdednam: Done. The segmentation faults have been fixed with this, thank you very much for the suggestion. The code now also works with the throw on line ~135 of variableValue.cc reinstated.

engine/variableValue.h
wdednam: Done, there is now a new member in variableValue called dataSize():

    size_t dataSize() const {
		return index.size() ? index.size(): numDenseElements();
	}

engine/variableValue.h
wdednam: Would an appropriately defined template<> to set the index vector do the job?

template<T>
void setDataSize(T t) {
}

with a protected member m_dataSize set equal to t=index.size() or numDenseElement.size()?

model/plotWidget.cc
wdednam: Done, thanks for the suggestion.

model/plotWidget.cc
wdednam: The index vector at present contains the serialized or flattened hypercube indices corresponding the actual data, so it is immediately available, but I'm not sure that is correct. As alluded to above, this is where my greatest confusion about the sparse data implementation lies.

model/ravelWrap.cc
wdednam: I've done so through the common dataSize() member that substitutes index.size() and numDenseElements() everywhere.

model/variable.cc
wdednam: The check is now vv.dataSize()==1 only, but I can remove it if necessary.

wdednam:
The new changes I've made should also take care of the compile problems in equations.cc I had, and I have modified two tests that previously called numElements() and now call dataSize(). I'll think about adding tests to testTensorOps.cc. I might have to call a small but very sparse CSV file for this? I will also commit the new code today.

engine/CSVParser.cc
```diff
@@ -342,15 +342,16 @@ namespace minsky
     string buf;
     typedef vector<string> Key;
     map<Key,double> tmpData;
+    multimap<Key,double> tmpAll;
```
highperformancecoder: multiMap still here

engine/CSVParser.cc
```diff
@@ -447,26 +448,62 @@ namespace minsky
...
+            }
+        }    
+        else 
+	    {	
+          v.index.clear();			
+          v.setXVector(xVector);          
+          if (!cminsky().checkMemAllocation(tmpData.size()*sizeof(double)))
+            throw runtime_error("memory threshold exceeded");	  	  		
+          auto dims=v.tensorInit.dims=v.dims();	
+          v.tensorInit.data.clear();
```
highperformancecoder: Common code like this should be moved outside the conditional

engine/equations.cc
```diff
@@ -180,7 +180,8 @@ namespace MathDAG
                   OperationType::Type op, OperationType::Type accum, double groupIdentity)
     {
       // check if any arguments have x-vectors, and if so, initialise r.xVector
-      size_t oldNumElems=r.numElements();
+      // For feature 47
```
highperformancecoder: At some point "feature 47" is not a helpful comment. "Sparse data representation" might be more useful, if needed.

engine/evalOp.cc
```diff
@@ -767,15 +767,21 @@ namespace minsky
                 e.shape=from1.dims();
                 if (to.xVector!=from1.xVector)
                   to.setXVector(from1.xVector);
-                for (size_t i=0; i<from1.numElements(); ++i)
-                  t->in1.push_back(i+from1.idx());
+                  
+                // For feature 47
+                  for (size_t i=0; i<from1.dataSize(); ++i)                          
+                    t->in1.push_back(i+from1.idx());
```
highperformancecoder: For these operations, we probably need to transfer to index vector to to as well.

engine/evalOp.cc
```diff
@@ -985,11 +996,15 @@ namespace minsky
                 {
                 case index:
                   {
-                    vector<unsigned> targetDims{unsigned(from1.rank()),unsigned(from1.numElements())};
+					vector<unsigned> targetDims;  
+                    targetDims={unsigned(from1.rank()),unsigned(from1.dataSize())};
```
highperformancecoder: Any particular reason why you changed construction to assignment? Construction is (marginally) more efficient.

engine/variableValue.cc
highperformancecoder: No - vaRef() return the reference to the first data element.

engine/variableValue.h
highperformancecoder: Probably better to have a setIndex() setter, and a index() getter that returns a const reference to the m_index member variable. That way, the allocated vector is the correct length.

model/plotWidget.cc
```diff
@@ -488,21 +489,22 @@ namespace minsky
...
-              for (size_t i=1; i<yv.xVector.size(); ++i)
-                {
-                  label+=str(yv.xVector[i][(j/stride)%d[i]])+" ";
-                  stride*=d[i];
-                }
-              labelPen(extraPen,label);
-              extraPen++;
-            }
+          // For feature 47
+            for (size_t j=0 /*d[0]*/; j<std::min(size_t(10)*d[0], yv.dataSize()); j+=d[0])
```
highperformancecoder: In this case, we need to iterate j over the first 10 elements of the index vector.

test/runtests.sh
```diff
@@ -21,6 +21,8 @@ for i in test/00/*.sh; do
     # t0029a.sh tests that saving over a readonly file is prevented. Test doesn't work as root
     elif [ `whoami` = 'root' -a $i = test/00/t0029a.sh ]; then
         su minsky $i &> /dev/null
+    elif [ $i = test/00/findDefinition.sh ]; then
+        sh $i
```
highperformancecoder: This is temporary - I;m trying to chase down a mysterious regression failure on Travis.

highperformancecoder:
Thanks - there's a few suggestions for improvement above, but basically this is good enough to merge. The failure in Travis is a preexisting problem, that just suddenly appeared, not due to any code changes. It is not showing up on either of my OpenSUSE machines, so it is just mysterious why it suddenly started failing.

engine/variableValue.h
wdednam: Does the following look okay (in variableValue.h)?

  class VariableValue: public VariableType, public XVectorMixin
  {
    CLASSDESC_ACCESS(VariableValue);
  private:
    std::vector<size_t> m_index;

  public:
    std::vector<size_t> getIndex() **const** { return m_index;}
    **const** void setIndex(std::vector<size_t> index) {	m_index=index;}

    size_t dataSize() **const** {
		return getIndex().size() ? getIndex().size(): numDenseElements();
	}    

I've put the "const" in boldface because my studies of C++ have still not got to the point where I can understand the meaning of constify in different positions in member declarations.

engine/CSVParser.cc
wdednam: I've removed it and will shortly commit the latest fixes you have suggested.

engine/CSVParser.cc
wdednam: Done. Unfortunately I cannot declare a local index vector conditionally here, so it is declared in the common code whether the data is sparse or not (to be committed shortly).

engine/equations.cc
wdednam: Thanks, I'll improve the description before I commit the latest changes.

engine/evalOp.cc
wdednam: I still need to do this.

engine/evalOp.cc
wdednam: No, I forgot to revert it to construction. This is code I had with a conditional for sparse or not. I will fix it before the next recommit.

model/plotWidget.cc
wdednam: At present, I'm using the index vector just to propagate the size of the sparse data everywhere. I still need to modify the parts of the code which require the actual hypercube indices of the sparse data (e.g., CSV export and the code in plotWidget.cc you point out here).

engine/variableValue.h
highperformancecoder: In the first case, you're declaring the return type const (const void - which doesn't make much sense as you cannot mutate void anyway - but maybe the compiler allows it). In the second case, you're declaring the method const, which means you can call it on a const object.

WRT types, you can place the const before or after the type being qualified. I prefer before, because I speak English, and const is an adjective and that's where English puts adjectives. Other people seem to have the opposite choice. Finally, beloved of C++ job interviews, not so useful in real life,
T * const   -  the pointer is const, but the target is not
const T* const - both the pointer and target is const
T const * const  - ditto.

wdednam: > In the first case, you're declaring the return type const (const void - which doesn't make much sense as you cannot mutate void anyway - but maybe the compiler allows it).

Thanks, now I understand this. I didn't know the const could go before or after the type. I've mostly seen it after in type declarations, but it makes sense that it go first because it is akin to an adjective.

Indeed, the compiler allows the second case with or without the const on the type. It does not allow the const on the method, however. I get the error:

error: passing ‘const std::vector<long unsigned int>’ as ‘this’ argument discards qualifiers [-fpermissive]
   m_index=index;

I guess this is because I modify m_index, so the object is not const.

> Finally, beloved of C++ job interviews, not so useful in real life,
T * const - the pointer is const, but the target is not
const T* const - both the pointer and target is const
T const * const - ditto.

Thanks, very useful advice for interviews, on this, as well as the linked lists previously :)

PR #95, opened 2019-12-11
Fix for library dependency problem with items during Travis build

wdednam:
This change is [Reviewable](https://reviewable.io/reviews/highperformancecoder/minsky/95)

### DIFF
--- a/model/canvas.cc
+++ b/model/canvas.cc
@@ -395,26 +395,24 @@ namespace minsky
   
   ItemPtr Canvas::itemAt(float x, float y)
   {
-	// Fix for library dependency problem with items during Travis build    
-    ItemPtr item;
+	// Fix for library dependency problem with items during Travis build  
+    ItemPtr closestItem;
     auto minD=numeric_limits<float>::max();
     model->recursiveDo(&GroupItems::items,
                        [&](const Items&, Items::const_iterator i)
                        {
-                          float d=sqr((*i)->x()-x)+sqr((*i)->y()-y);
-                          if (d<minD && (*i)->visible() && (*i)->contains(x,y))
-                            {
-                              minD=d;
-                              item=*i;
-                            }
-                          return false;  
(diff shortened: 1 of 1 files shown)


model/canvas.cc
```diff
@@ -395,8 +395,20 @@ namespace minsky
...
+                       {
+                          float d=sqr((*i)->x()-x)+sqr((*i)->y()-y);
+                          if (d<minD && (*i)->visible() && (*i)->contains(x,y))
+                            {
+                              minD=d;
+                              item=*i;
+                            }
+                          return false;  
+                       });
     if (!item)
```
highperformancecoder: This statement is now redundant

model/canvas.cc
```diff
@@ -395,8 +395,20 @@ namespace minsky
   
   ItemPtr Canvas::itemAt(float x, float y)
   {
-    auto item=model->findAny(&Group::items,
-                        [&](const ItemPtr& i){return i->visible() && i->contains(x,y);});
+    // Fix for library dependency problem with items during Travis build     
+    ItemPtr item;                    
+    auto minD=numeric_limits<float>::max();
+    model->recursiveDo(&GroupItems::items,
```
highperformancecoder: Great - almost exactly how I would have done it! An it looks like it has fixed the problem - yay!

model/canvas.cc
wdednam: Thanks, I didn't understand why I still needed it, but on a previous iteration the code seemed to find the item everywhere on the canvas, so I added it back in, but I must have had some other bit of code that caused that. I will remove it and recommit for a pull request.

model/canvas.cc
wdednam: Thanks. I just followed your instructions, which were crystal clear, and repurposed your existing code in canvas.cc for finding the closest "in" port. I wish all bugs were this easy to fix.:)

PR #96, opened 2019-12-11
Update differentiate and Trigonometric functions

halcrichtonstandish:
Update differentiate and Trigonometric functions

---
This change is [Reviewable](https://reviewable.io/reviews/highperformancecoder/minsky/96)

### DIFF
--- a/doc/Reference.tex
+++ b/doc/Reference.tex
@@ -61,7 +61,9 @@ ports are labelled $x$ and $y$, referring the the formula $x^y$.
 \subsection{time $t$}\label{Operation:time}  Returns the current value of system time.
 
 \subsection{differentiate $d/dt$}\label{Operation:differentiate}
-Symbolically differentiates its input.
+Symbolically differentiates its input, producing d/dx[input]. For example, if you input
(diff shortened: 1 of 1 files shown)


doc/Reference.tex
```diff
@@ -61,7 +61,9 @@ \subsection{not $\neg$}\label{Operation:not_} The output is 1 or 0, depending
 \subsection{time $t$}\label{Operation:time}  Returns the current value of system time.
 
 \subsection{differentiate $d/dt$}\label{Operation:differentiate}
-Symbolically differentiates its input.
+Symbolically differentiates its input, producing d/dx[input]. For example, if you input
```
highperformancecoder: Well it actually differentiates with respect to time, not with respect to its input. That's why it is d/dt.

As to audience assumed technical ability, I think it is reasonable to assume they have HSC 2 unit knowledge, ie trig functions, logs etc, and single variable differential and integral calculus and so - albeit maybe a dim distant memory. It is fine to reference wikipedia articles (using \htmladdnormallink), but not to explain high school maths.

PR #98, opened 2019-12-14
refactor curved wires and recover delete handle on double click

wdednam:
This change is [Reviewable](https://reviewable.io/reviews/highperformancecoder/minsky/98)

### DIFF
--- a/gui-tk/wiring.tcl
+++ b/gui-tk/wiring.tcl
@@ -524,9 +524,10 @@ proc doubleButton {x y} {
         selectVar $x $y
         editItem
     }
-    
-    # For ticket 1092. Reinstate delete handle user interaction
-	canvas.delHandle $x $y    
+    # For ticket 1092. Reinstate delete handle user interaction    
+    } else {
+    canvas.delHandle $x $y
+    }    
 }
 # for ticket 1062, new hierarchy of context menu access on mouse right click: wires, items and background canvas.
 bind .wiring.canvas <<contextMenu>> {
--- a/gui-tk/wiring.tcl
+++ b/gui-tk/wiring.tcl
@@ -524,10 +524,9 @@ proc doubleButton {x y} {
         selectVar $x $y
         editItem
     }
-    # For ticket 1092. Reinstate delete handle user interaction    
-    } else {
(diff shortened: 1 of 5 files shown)


gui-tk/wiring.tcl
```diff
@@ -524,6 +524,9 @@ proc doubleButton {x y} {
         selectVar $x $y
         editItem
     }
+    
+    # For ticket 1092. Reinstate delete handle user interaction
```
highperformancecoder: Can we be sure that you won't select an item and a handle at the same time? It might be safer to have delHandle con ditioned in the else part of this conditional.

model/wire.cc
```diff
@@ -543,6 +537,26 @@ namespace
...
+      {
+        float d=d2(c[i],c[i+1],x,y);
+        if (d<closestD)
+          {
+            closestD=d;
+            n=i;
+          }
+      }	  
+    assert(n<c.size()-1);
+    c.erase(c.begin()+n);
```
highperformancecoder: Don't you have to erase the y coordinate as well? c.erase(c.begin()+n, c.erase(c.begin()+n+2) should erase both the x and y coordinate values.

gui-tk/wiring.tcl
wdednam: Thanks, I've put it in the else part.

model/wire.cc
wdednam: Thanks, I've made it c.erase(c.begin()+n, c.erase(c.begin()+n+1));

With n+2 it deletes more than one handle on double-click.

wdednam:
After making more changes to this pull request, many tests failed and I don't know why. I've reverted the latest changes but the build would not finish as you can see above.

model/wire.cc
highperformancecoder: Sorry I meant to say c.erase(c.begin()+n, c.begin()+n+2), which erases two items starting at c.begin()+n in one hit. What you ended up with is pretty much equivalent, although terribly obscure, and with two calls to erase, so the trailing data will be copied twice. I'll approve this, and fix it in master branch.

PR #107, opened 2019-12-19
Definitive fix for ticket 1080 requires two modes of pasting: 

wdednam:
pasteItems() and pasteGroup(). Thus, canvas context menu now has two pasting options: paste group and paste items to allow the user to select which they prefer.

---
This change is [Reviewable](https://reviewable.io/reviews/highperformancecoder/minsky/107)

### DIFF
--- a/gui-tk/wiring.tcl
+++ b/gui-tk/wiring.tcl
@@ -499,8 +499,7 @@ proc canvasContext {x y X Y} {
     .wiring.context add command -label "Cut" -command cut
     .wiring.context add command -label "Copy" -command minsky.copy
     .wiring.context add command -label "Save selection as" -command saveSelection
-    .wiring.context add command -label "Paste Group" -command minsky.pasteGroup
-    .wiring.context add command -label "Paste Item(s)" -command minsky.pasteItems
+    .wiring.context add command -label "Paste" -command minsky.paste
     .wiring.context add command -label "Bookmark here" -command "bookmarkAt $x $y $X $Y"
     .wiring.context add command -label "Group" -command "minsky.createGroup"
     .wiring.context add command -label "Lock selected Ravels" -command "minsky.canvas.lockRavelsInSelection"
--- a/gui-tk/wiring.tcl
+++ b/gui-tk/wiring.tcl
@@ -499,7 +499,8 @@ proc canvasContext {x y X Y} {
     .wiring.context add command -label "Cut" -command cut
     .wiring.context add command -label "Copy" -command minsky.copy
     .wiring.context add command -label "Save selection as" -command saveSelection
-    .wiring.context add command -label "Paste" -command {paste}
+    .wiring.context add command -label "Paste Group" -command minsky.pasteGroup
+    .wiring.context add command -label "Paste Items" -command minsky.pasteItems
     .wiring.context add command -label "Bookmark here" -command "bookmarkAt $x $y $X $Y"
     .wiring.context add command -label "Group" -command "minsky.createGroup"
     .wiring.context add command -label "Lock selected Ravels" -command "minsky.canvas.lockRavelsInSelection"
--- a/gui-tk/wiring.tcl
+++ b/gui-tk/wiring.tcl
@@ -500,7 +500,7 @@ proc canvasContext {x y X Y} {
     .wiring.context add command -label "Copy" -command minsky.copy
     .wiring.context add command -label "Save selection as" -command saveSelection
     .wiring.context add command -label "Paste Group" -command minsky.pasteGroup
-    .wiring.context add command -label "Paste Items" -command minsky.pasteItems
+    .wiring.context add command -label "Paste Item(s)" -command minsky.pasteItems
     .wiring.context add command -label "Bookmark here" -command "bookmarkAt $x $y $X $Y"
     .wiring.context add command -label "Group" -command "minsky.createGroup"
     .wiring.context add command -label "Lock selected Ravels" -command "minsky.canvas.lockRavelsInSelection"
--- a/model/minsky.cc
+++ b/model/minsky.cc
@@ -255,8 +255,7 @@ namespace minsky
 //      throw runtime_error("cannot save to "+fileName);
   }
 
-    // Fix for ticket 1080
-  void Minsky::pasteGroup()
+  void Minsky::paste()
   {
     istringstream is(getClipboard());
     xml_unpack_t unpacker(is);
@@ -265,36 +264,15 @@ namespace minsky
     GroupPtr g(new Group);
     canvas.setItemFocus(model->addGroup(g));
     m.populateGroup(*g);
-    g->resizeOnContents();	  
-  }
-
-    // Fix for ticket 1080
-  void Minsky::pasteItems()
-  {
-    istringstream is(getClipboard());
-    xml_unpack_t unpacker(is);
-    schema2::Minsky m;
-    xml_unpack(unpacker, "Minsky", m);
-    GroupPtr g(new Group);
-    canvas.setItemFocus(model->addGroup(g));
-    m.populateGroup(*g);
-    if (auto p=g->group.lock())
-    {
-	   p->moveContents(*g); 
-	   canvas.item.reset();
-       model->recursiveDo(&GroupItems::items,
-                              [&](const Items&, Items::const_iterator i)
-                              {
-                                 canvas.item=*i;
-                                 canvas.item->selected=true;
-                                 canvas.selection.ensureItemInserted(canvas.item);
-                                 return false;  
-                              });       
-       canvas.setItemFocus(model->addItem(canvas.item)); 
-       model->normaliseGroupRefs(model);                                  
-       g->splitBoundaryCrossingWires();   
-    }
-    g->resizeOnContents();
+    // Fix for ticket 1080    
+    auto p = model;
+    p->moveContents(*g);
+    for (auto& i: p->items) {   
+        canvas.selection.ensureItemInserted(i);  
+        canvas.setItemFocus(model->addItem(i)); 
+	}
+    g->resizeOnContents();   
+    model->removeGroup(*g);
   }
 
   void Minsky::toggleSelected(ItemType itemType, int item)
--- a/model/minsky.cc
+++ b/model/minsky.cc
@@ -255,7 +255,8 @@ namespace minsky
 //      throw runtime_error("cannot save to "+fileName);
   }
 
-  void Minsky::paste()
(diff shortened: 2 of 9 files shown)


model/minsky.cc
```diff
@@ -264,6 +265,35 @@ namespace minsky
...
+  void Minsky::pasteItems()
+  {
+    istringstream is(getClipboard());
+    xml_unpack_t unpacker(is);
+    schema2::Minsky m;
+    xml_unpack(unpacker, "Minsky", m);
+    GroupPtr g(new Group);
+    canvas.setItemFocus(model->addGroup(g));
+    m.populateGroup(*g);
+    if (auto p=g->group.lock())
```
highperformancecoder: p should equal model at this stage.

model/minsky.cc
```diff
@@ -264,6 +265,35 @@ namespace minsky
...
+    {
+	   p->moveContents(*g); 
+	   canvas.item.reset();
+       model->recursiveDo(&GroupItems::items,
```
highperformancecoder: No need to do a recursiveDo. It suffices to loop over items and groups of g, and call insertItem and insertGroup on that. I don't think there is any scenario where any items could be duplicated.

gui-tk/wiring.tcl
```diff
@@ -499,7 +499,8 @@ proc canvasContext {x y X Y} {
     .wiring.context add command -label "Cut" -command cut
     .wiring.context add command -label "Copy" -command minsky.copy
```
highperformancecoder: Needs to call your new TCL procedure.

gui-tk/wiring.tcl
```diff
@@ -499,7 +499,8 @@ proc canvasContext {x y X Y} {
...
     .wiring.context add command -label "Save selection as" -command saveSelection
-    .wiring.context add command -label "Paste" -command {paste}
+    .wiring.context add command -label "Paste Group" -command minsky.pasteGroup
+    .wiring.context add command -label "Paste Item(s)" -command minsky.pasteItems
```
highperformancecoder: I would think we just want pastItems, and remove pasteGroup altogether. What do you think?

test/00/t0033a.sh
```diff
@@ -48,7 +48,7 @@ proc afterMinskyStarted {} {uplevel #0 {
   }
  minsky.copy
  saveSelectionAsFile xxx.mky
- group.get [paste]
+ group.get [pasteGroup]
```
highperformancecoder: pasteItems should leave all the items in the selection, so if followed by group, we should end up in the same state as tested by these tests.

model/minsky.cc
wdednam: Thanks, that seems much simpler than what I have.

model/minsky.cc
wdednam: Okay, I've done this, but I don't know if it is correct. Test t0036a.sh fails with this new change:

  void Minsky::pasteItems()
  {
    istringstream is(getClipboard());
    xml_unpack_t unpacker(is);
    schema2::Minsky m;
    xml_unpack(unpacker, "Minsky", m);
    GroupPtr g(new Group);
    m.populateGroup(*g);
    **auto p = model;
    p->moveContents(*g);
    for (auto& i: p->items)
      if (i->visible()) {
        canvas.selection.ensureItemInserted(i);
	}
    for (auto& i: p->groups)
      if (i->visible()) {
        canvas.selection.ensureGroupInserted(i);
	}**
    g->resizeOnContents();
  }

gui-tk/wiring.tcl
wdednam: How about making "PasteItems" just plain "Paste" and if I can't get one function to paste both groups and separate items correctly, I can maybe add a "Paste Special" with a popup menu for pasting groups which were groups before they were copied?

test/00/t0033a.sh
wdednam: Okay, I've tried this. When I use PasteItems followed by CreateGroup in t0036a.sh, I get the error:

assertion: [model.numGroups]==2 failed: 
type eval line 11 cmd {assert {[model.numGroups]==2}} proc ::afterMinskyStarted
FAILED

wdednam:
I'll need to work a little more on this one. What I've tried so far does not quite do the job.

test/00/t0033a.sh
highperformancecoder: You can add a "puts [model.numGroups]" to see what minsky thinks it is.

wdednam: Thanks, now It finds the correct number of groups, but there is still an problem in these two lines:

  # retrieve inner group
  findObject Group
  \$item.groups.@elem 0

So, when an inner group has to be fetched. The error is:

2   (THIS 2 IS RETURNED BY puts [model.numGroups])
index 0 out of boundsindex 0 out of bounds
index 0 out of bounds
    while executing
"$item.groups.@elem 0"
    ("uplevel" body line 16)
    invoked from within
"uplevel #0 {
  openNamedFile /home/wd/minsky_dev/examples/GoodwinLinear02.mky
  assert {[findObject Group]}
  set item minsky.canvas.item
  canvas.mou..."
    (procedure "afterMinskyStarted" line 1)
    invoked from within
"afterMinskyStarted"
    invoked from within
"if {[llength [info commands afterMinskyStarted]]>0} {
    afterMinskyStarted
}"
    (file "/home/wd/minsky_dev/gui-tk/minsky.tcl" line 1649)
FAILED

Kris31415:
Get me off this list....please....I have asked more than once, next time I will not ask nicely. 



On 20 Dec 2019, at 11:25, Wynand <notifications@github.com> wrote:

@wdednam commented on this pull request.

In test/00/t0033a.sh <https://github.com/highperformancecoder/minsky/pull/107#discussion_r360312176>:

> @@ -48,7 +48,7 @@ proc afterMinskyStarted {} {uplevel #0 {
   }
  minsky.copy
  saveSelectionAsFile xxx.mky
- group.get [paste]
+ group.get [pasteGroup]
Thanks, now It finds the correct number of groups, but there is still an problem in these two lines:

retrieve inner group

findObject Group
$item.groups.@elem <https://github.com/elem> 0

So, when an inner group has to be fetched. The error is:

2 (THIS 2 IS RETURNED BY puts [model.numGroups])
index 0 out of boundsindex 0 out of bounds
index 0 out of bounds
while executing
"$item.groups.@elem <https://github.com/elem> 0"
("uplevel" body line 16)
invoked from within
"uplevel #0 {
openNamedFile /home/wd/minsky_dev/examples/GoodwinLinear02.mky
assert {[findObject Group]}
set item minsky.canvas.item
canvas.mou..."
(procedure "afterMinskyStarted" line 1)
invoked from within
"afterMinskyStarted"
invoked from within
"if {[llength [info commands afterMinskyStarted]]>0} {
afterMinskyStarted
}"
(file "/home/wd/minsky_dev/gui-tk/minsky.tcl" line 1649)
FAILED

test/00/t0033a.sh
highperformancecoder: This is saying that the first group found has no groups contained in it.

I'd suggest adding a minsky.save call to save the state of the model to a file, then opening the save file in Minsky to get a picture of what's going on.

highperformancecoder: NB - note the tests work within a temporary directory, which gets deleted once the test is done, so save to an absolute known path. Otherwise do "here=`pwd` in the top level Minsky directory, then copy past the cat command that creates the input file, and run that directly:
gui-tk/minsky input.tcl

highperformancecoder:
Which of these two pull requests is the definitive one?

highperformancecoder:
On Fri, Dec 20, 2019 at 04:47:46AM -0800, Kris31415 wrote:
> Get me off this list....please....I have asked more than once, next time I will
> not ask nicely.

Please take this up with Github. This is not an email list over which
we have control, but rather correspondence between people and entities
on particular project issues.


-- 

Principal, High Performance Coders
Visiting Senior Research Fellow        hpcoder@hpcoders.com.au
Economics, Kingston University         http://www.hpcoders.com.au

wdednam:
> Which of these two pull requests is the definitive one?

I think we should delete this pull request, it is messing up others like the one for ticket 1043 which could have been merged already.

test/00/t0033a.sh
wdednam: Thanks. I've done what you suggested in your first message. When I perform pasteItems with the changes I made to paste loose items as ungrouped on a new Canvas, not only the original "Philips Curve" group appears, but also the small phantom group that Hal also picked up in a recent ticket. I think the script then fails when it tries to retrieve the contents of the phantom group. I attach a screen cap of what is saved when I use pasteItems instead of pasteGroup (which is just the original minsky.paste()).
t0036a sh_screencap

wdednam: Since we don't want to copy/paste groups within groups anymore, i.e., an existing group should copied/pasted as the same group and not a group within a group, and ditto for items, I've modified t0036a.sh to take this into account:

##### insert ecolab script code here
##### use \$ in place of $ to refer to variable contents
##### exit 0 to indicate pass, and exit 1 to indicate failure
##### test copy/pasting of groups
cat >input.tcl <<EOF
source $here/test/assert.tcl
proc afterMinskyStarted {} {uplevel #0 {
  openNamedFile $here/examples/GoodwinLinear02.mky
  assert {[findObject Group]}
  set item minsky.canvas.item
  canvas.mouseDown [expr [\$item.x]-0.55*[\$item.width]] [expr [\$item.y]+0.55*[\$item.height]]
  canvas.mouseUp [expr [\$item.x]+0.55*[\$item.width]] [expr [\$item.y]-0.55*[\$item.height]] 
  assert {[canvas.selection.groups.size]==1}
  newSystem
  paste
  **###### For ticket 1080. There is no longer an outer group when existing groups or items are pasted between canvasses or in the same canvas
  assert {[model.numGroups]==1}
  assert {[model.numItems]==8}
  assert {[model.numWires]==8}**
  tcl_exit
}}
EOF

model/minsky.cc
wdednam: Here is the latest solution (insideminsky::paste()) I have committed:

    // Fix for ticket 1080    
    auto p = model;
    p->moveContents(*g);
    for (auto& i: p->items) {   
        canvas.selection.ensureItemInserted(i);  
        canvas.setItemFocus(model->addItem(i)); 
	}
    g->resizeOnContents();   
    model->removeGroup(*g);

gui-tk/wiring.tcl
wdednam: We are back to using just one paste() function.

PR #113, opened 2019-12-23
Proposed fix for ticket 1053. Disallow moving column headings between…

wdednam:
… different asset classes

---
This change is [Reviewable](https://reviewable.io/reviews/highperformancecoder/minsky/113)


model/godleyTableWindow.cc
```diff
@@ -485,8 +485,8 @@ namespace minsky
     if ((selectedCol==0 && selectedRow==1) || (c==0 && r==1) || size_t(selectedRow)>=(godleyIcon->table.rows()) || size_t(r)>=(godleyIcon->table.rows()) || size_t(c)>=(godleyIcon->table.cols()) || size_t(selectedCol)>=(godleyIcon->table.cols()))
       return;  
     else if (selectedRow==0)
-      {
-        if (c>0 && selectedCol!=0 && c!=selectedCol)      // Disallow moving flow labels column. For ticket 1064/1066
+      { // Disallow moving flow labels column, in paticular between different asset classes. For tickets 1053/1064/1066
+        if (c>0 && selectedCol!=0 && c!=selectedCol && godleyIcon->table._assetClass(c)==godleyIcon->table._assetClass(selectedCol))
```
highperformancecoder: I'm not at all sure about this change. I posted a comment on ticket #1053. Let's see what Steve says, as it's his ticket.

wdednam: I've undone the change. Other changes have been committed as a partial fix because the bug is hard to pin down. Basically, the problem manifests when the left mouse button is released and GodleyTable::moveCol() is called from within mouseUp() in godleyTableWindow.cc. I think something is up with how noAssetClass columns are added and removed and turned into a target asset class in moveCol(). When I comment out _assetClass(col,n) in moveCol, the problem disappears but then the ability to move whole columns is disabled.

PR #114, opened 2019-12-26
Copying and pasting between two different canvasses with items remaining in selection but not attached to mouse cursor

wdednam:
This change is [Reviewable](https://reviewable.io/reviews/highperformancecoder/minsky/114)


model/minsky.cc
```diff
@@ -264,7 +264,12 @@ namespace minsky
     GroupPtr g(new Group);
     canvas.setItemFocus(model->addGroup(g));
```
highperformancecoder: I don't think this line is needed at all, although you undo it later on on line 272.

PR #115, opened 2019-12-29
Updated fix for ticket 1080. Only pasted items remain in selection

wdednam:
This change is [Reviewable](https://reviewable.io/reviews/highperformancecoder/minsky/115)

### DIFF
--- a/model/group.cc
+++ b/model/group.cc
@@ -385,6 +385,31 @@ namespace minsky
          source.clear();
        }
   }
+  
+  // Move contents out of source group but leave them in selection. Fix for tickets 1080
+  void Group::moveContentsInSelection(Group& source) {
+     if (&source!=this)
+       {
+         if (source.higher(*this))
+             throw error("attempt to move a group into itself");
+         // make temporary copies as addItem removes originals
+         auto copyOfItems=source.items;
+         for (auto& i: copyOfItems)
+           {
+             addItem(i);
+             minsky().canvas.selection.ensureItemInserted(i);
+             assert(!i->ioVar());
+           }
+         auto copyOfGroups=source.groups;
+         for (auto& i: copyOfGroups)
+         {
+           addGroup(i);
+           minsky().canvas.selection.ensureGroupInserted(i);
+	     }
+         /// no need to move wires, as these are handled above
+         source.clear();
+       }
+  }    
 
   VariablePtr Group::addIOVar()
   {
--- a/model/group.cc
+++ b/model/group.cc
@@ -385,31 +385,6 @@ namespace minsky
          source.clear();
        }
   }
-  
-  // Move contents out of source group but leave them in selection. Fix for tickets 1080
-  void Group::moveContentsInSelection(Group& source) {
-     if (&source!=this)
-       {
-         if (source.higher(*this))
-             throw error("attempt to move a group into itself");
-         // make temporary copies as addItem removes originals
-         auto copyOfItems=source.items;
-         for (auto& i: copyOfItems)
-           {
-             addItem(i);
-             minsky().canvas.selection.ensureItemInserted(i);
-             assert(!i->ioVar());
-           }
-         auto copyOfGroups=source.groups;
-         for (auto& i: copyOfGroups)
-         {
-           addGroup(i);
-           minsky().canvas.selection.ensureGroupInserted(i);
-	     }
-         /// no need to move wires, as these are handled above
-         source.clear();
-       }
-  }    
 
   VariablePtr Group::addIOVar()
   {
--- a/model/minsky.cc
+++ b/model/minsky.cc
@@ -266,10 +266,8 @@ namespace minsky
     m.populateGroup(*g);
     // Default pasting no longer occurs as grouped items or as a group within a group. Fix for tickets 1080/1098    
     auto p = model;
-	p->moveContents(*g);
-    for (auto& i: p->items)   
-        canvas.selection.ensureItemInserted(i);  
-    model->removeGroup(*g);
+	p->moveContentsInSelection(*g);
+    p->removeGroup(*g);
   }
 
   void Minsky::toggleSelected(ItemType itemType, int item)
--- a/model/minsky.cc
+++ b/model/minsky.cc
@@ -265,15 +265,18 @@ namespace minsky
     canvas.setItemFocus(model->addGroup(g));
     m.populateGroup(*g);
     // Default pasting no longer occurs as grouped items or as a group within a group. Fix for tickets 1080/1098    
-    auto p = minsky().model;
-	for (auto i: p->items)	
-	    if (!dynamic_cast<IntOp*>(i.get())) canvas.selection.toggleItemMembership(i);   
-	p->moveContents(*g);
-	auto q = model;
-	for (auto i: q->items)
-	  canvas.selection.toggleItemMembership(i); 	
-	g->splitBoundaryCrossingWires(); 
-    p->removeGroup(*g);
+    auto copyOfItems=g->items;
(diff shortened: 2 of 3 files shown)


model/minsky.cc
```diff
@@ -266,10 +266,8 @@ namespace minsky
     m.populateGroup(*g);
     // Default pasting no longer occurs as grouped items or as a group within a group. Fix for tickets 1080/1098    
     auto p = model;
-	p->moveContents(*g);
-    for (auto& i: p->items)   
-        canvas.selection.ensureItemInserted(i);  
-    model->removeGroup(*g);
+	p->moveContentsInSelection(*g);
+    p->removeGroup(*g);
```
highperformancecoder: I don't think this line nor 265 is needed.

model/group.cc
```diff
@@ -385,6 +385,31 @@ namespace minsky
          source.clear();
        }
   }
+  
+  // Move contents out of source group but leave them in selection. Fix for tickets 1080
+  void Group::moveContentsInSelection(Group& source) {
```
highperformancecoder: Why do we need this method? Surely it is enough to loop over the temporary pasted group's contents call ensureItemInserted on that and then addItem on model.

model/minsky.cc
wdednam: When I removed them, I got a segmentation fault during tests t0026a.sh and t0036a.sh, so I put them back in.

highperformancecoder: What was the cause of the segfault?

highperformancecoder: Does removeGroup also clear itemFocus?

model/group.cc
wdednam: When I set "auto p = model", and move g's contents to p, all the items on the canvas, not only the temporary pasted group's, are looped over on using "for (auto i:p->items)" to put the items in selection. I don't know why if only g's contents get moved to p, all items are selected when I loop over those in p? This is what caused the bug you caught and pointed out on sourceforge. So, the only way I could ensure that only the temporary group's contents are looped over and remain in selection, was to do so inside Group::moveContents(). Obviously, when we ungroup items, we don't necessarily want them to be in selection upon ungrouping, so I added a second method to deal with the case where they need to remain in selection.

model/minsky.cc
wdednam: > What was the cause of the segfault?

I followed the steps in t0036a.sh in gdb and the following backtrace:

Screenshot_segfault_ticket1080

wdednam: > Does removeGroup also clear itemFocus?

No, it doesn't look like it does to me:

  GroupPtr GroupItems::removeGroup(const Group& group)
  {
    for (auto i=groups.begin(); i!=groups.end(); ++i)
      if (i->get()==&group)
        {
          GroupPtr r=*i;
          groups.erase(i);
          return r;
        }

    for (auto& g: groups)
      if (GroupPtr r=g->removeGroup(group))
        return r;
    return GroupPtr();
  }

model/group.cc
highperformancecoder: OK - but its a lot of extra code that does stuff, then undoes most of the stuff to leave what you want done. What you want done is two things:

1) insertion of all items and groups of the temporary group into canvas.model (which is not necessarily the same as minsky().model). Wires will automatically be carried through if the addItem method is used.
2) Insertion of all items and groups into the selection. You will need to do this step prior to step 1), because it is not possible for an item to belong to two groups at once, with the exception of the selection.

I think the key is to resolve why the simplest approach of not inserting the temporary group (mentioned below) causes a segfault. Time to break out gdb.

model/minsky.cc
wdednam: > What was the cause of the segfault?

I attach another screenshot with more details on frames 5-7 of the backtrace at the bottom:

Screenshot_segfault_backtrace_frames_ticket1080

model/group.cc
wdednam: 1. Thanks. The fact that minsky().model and canvas.model are not necessarily the same has helped me to find a way to perform the pasting operation by looping separately over the two types of items and toggling their selection status. That way only the pasted items remain in selection:

`      auto p = minsky().model;
	for (auto i: p->items)
	    if (!dynamic_cast<IntOp*>(i.get())) canvas.selection.toggleItemMembership(i);   
	p->moveContents(*g);
	auto q = canvas.model;
	for (auto i: q->items)
	    canvas.selection.toggleItemMembership(i);   	`

For some reason an IntOp item that has not been selected also remains in selection if I don't add " if (!dynamic_cast<IntOp*>(i.get()))" when looping over the minsky().model items.

2. Doing the above, I now get a more clearly defined crash too, which has something to do with wires:

minsky: model/group.cc:281: void minsky::GroupItems::adjustWiresGroup(minsky::Wire&): Assertion `p1 && p2' failed.

The error again disappears when I reinstate canvas.setItemFocus(model->addGroup(g)); a few lines up. Very strange. I will try to finish sorting it out when I get back from my trip tomorrow evening and continue working on the other bugs on Steve's list.

highperformancecoder: This also strikes me as a bit bizarre. The usual scenario is that canvas.model == minsky.model. So this will basically be the identity operation. When canvas.mode != minsky,model, it ends up filling the selection with the contents of both groups.

The assertion failure indicates that one of the wire ends is not owned by a group.

wdednam: Will I need to pass the coordinates as arguments to Minsky::paste() to add all the wire ends via  the ports they are attached to? I have tried several approaches to remove canvas.itemFocus(model->addGroup(g)) and p->removeGroup(*g), but the best I could get also deleted the wire during the process...

highperformancecoder: No - the actual wired coordinate data is relative to the end points, which is decided by the port coordinates, which in turn is relative to the coordinates of the items to which they belong.

Group::adjustWires will move relevant wires to the groups that should own them.

model/minsky.cc
```diff
@@ -265,8 +265,14 @@ namespace minsky
     canvas.setItemFocus(model->addGroup(g));
     m.populateGroup(*g);
     // Default pasting no longer occurs as grouped items or as a group within a group. Fix for tickets 1080/1098    
-    auto p = model;
-	p->moveContentsInSelection(*g);
+    auto p = minsky().model;
+	for (auto i: p->items)	
+	    if (!dynamic_cast<IntOp*>(i.get())) canvas.selection.toggleItemMembership(i);   
+	p->moveContents(*g);
```
highperformancecoder: This empties g, so line 275 becomes a nop. In any case, g should not have any boundary crossing wires.

model/minsky.cc
```diff
@@ -265,8 +265,14 @@ namespace minsky
...
+	auto q = model;
+	for (auto i: q->items)
+	  canvas.selection.toggleItemMembership(i);
```
highperformancecoder: Assuming selection is empty, this adds all model items to the selection. Again not what we want.

model/group.cc
wdednam: Thanks. 

I see Group::adjustWires is called in Group::moveContents, so that should be taken care of.

model/minsky.cc
wdednam: Mmmmm, I'm seeing really bizarre behaviour. When I use only "auto p=canvas.model" or "auto p=minsky().model", a subset of items remain selected after "p->moveContents(*g)" in most of the examples I've tested. But they are not the items I have selected for copying/cutting and pasting. The only time I get the items I've selected (and very weirdly also the two first variables on the bottom of a Godley icon???) after pasting, is when I first loop over items in minsky().model and afterwards canvas.model. I agree that is bizarre, but I don't understand why it happens.

PR #116, opened 2020-01-04
definitive fix for ticket 1098

wdednam:
Certain items still remained attached to the tip of the mouse cursor upon copying and pasting, e.g., constants, parameters, variables and integral operators. This fix hopefully solves that.

---
This change is [Reviewable](https://reviewable.io/reviews/highperformancecoder/minsky/116)

### DIFF
--- a/gui-tk/wiring.tcl
+++ b/gui-tk/wiring.tcl
@@ -497,9 +497,9 @@ proc canvasContext {x y X Y} {
     .wiring.context delete 0 end
     .wiring.context add command -label Help -command {help DesignCanvas}
     .wiring.context add command -label "Cut" -command cut
-    .wiring.context add command -label "Copy" -command minsky.copy
+    .wiring.context add command -label "Copy" -command "minsky.copy"
     .wiring.context add command -label "Save selection as" -command saveSelection
-    .wiring.context add command -label "Paste" -command minsky.paste
+    .wiring.context add command -label "Paste" -command "minsky.paste"
     .wiring.context add command -label "Bookmark here" -command "bookmarkAt $x $y $X $Y"
     .wiring.context add command -label "Group" -command "minsky.createGroup"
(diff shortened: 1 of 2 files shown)


gui-tk/wiring.tcl
```diff
@@ -497,9 +497,9 @@ proc canvasContext {x y X Y} {
     .wiring.context delete 0 end
     .wiring.context add command -label Help -command {help DesignCanvas}
     .wiring.context add command -label "Cut" -command cut
-    .wiring.context add command -label "Copy" -command minsky.copy
+    .wiring.context add command -label "Copy" -command "minsky.copy"
```
highperformancecoder: Just for future reference, quotes are only needed when the command contains spaces.

gui-tk/wiring.tcl
```diff
@@ -650,7 +650,7 @@ proc contextMenu {x y X Y} {
                .wiring.context add command -label "Initialise Random" \
                     -command "initRandom" 
             }
-            .wiring.context add command -label "Copy" -command "canvas.copyItem"
+            .wiring.context add command -label "Copy" -command "minsky.copy"
```
highperformancecoder: This changes the semantics. In the context menu, the copy command means "copy this item". You've changed it to "copy the selection", which is rather confusing, I would think.

gui-tk/wiring.tcl
```diff
@@ -629,7 +629,7 @@ proc contextMenu {x y X Y} {
                 renameVariableInstances
             }
             .wiring.context add command -label "Edit" -command "editItem"
-            .wiring.context add command -label "Copy" -command "canvas.copyItem"
+            .wiring.context add command -label "Copy" -command "minsky.copy"
```
highperformancecoder: You've change the semantics here. In an item context menu, copy mean "copy this item", and you;ve changed it to mean "copy the selection". Rather confusing, I would think.

highperformancecoder:
Rejecting this PR, as it only contains neutral or incorrect changes.

PR #117, opened 2020-01-05
Back to using copy.canvasItem. Pating now occurs at position x y on the canvas.

wdednam:
This change is [Reviewable](https://reviewable.io/reviews/highperformancecoder/minsky/117)

### DIFF
--- a/gui-tk/minsky.tcl
+++ b/gui-tk/minsky.tcl
@@ -203,7 +203,23 @@ wm title . "$progName: $fname"
 setBackgroundColour $backgroundColour
 proc tk_focusPrev {win} {return $win}
 proc tk_focusNext {win} {return $win}
-if $preferences(focusFollowsMouse) tk_focusFollowsMouse
+if {$preferences(focusFollowsMouse)} {
+	tk_focusFollowsMouse
+# Make tab traversal possible within a window that is given focus by only clicking on it (no focusFollowsMouse). For ticket 901.	
+} else {
+    set old [bind all <Enter>]
+    set script {
+	if {"%d" eq "NotifyAncestor" || "%d" eq "NotifyNonlinear" \
+		|| "%d" eq "NotifyInferior"} {
+	       tk::FocusOK %W	
+	    }
+    }
+    if {$old ne ""} {
+	bind all <Enter> "$old; $script"
+    } else {
+	bind all <Enter> $script
+    }
+}
 proc setCursor {cur} {. configure -cursor $cur; update idletasks}
 
 #source $minskyHome/library/htmllib.tcl
@@ -1360,7 +1376,23 @@ proc setPreferenceParms {} {
     } else {
         place forget .wiring.panopticon
     }
-    if {$preferences(focusFollowsMouse)} tk_focusFollowsMouse
+    if {$preferences(focusFollowsMouse)} {
+		tk_focusFollowsMouse
+	# Make tab traversal possible within a window that is given focus by only clicking on it (no focusFollowsMouse). For ticket 901.
+    } else {
+       set old [bind all <Enter>]
+       set script {
+	   if {"%d" eq "NotifyAncestor" || "%d" eq "NotifyNonlinear" \
+	   	|| "%d" eq "NotifyInferior"} {
+	          tk::FocusOK %W	
+	       }
+       }
+       if {$old ne ""} {
+	   bind all <Enter> "$old; $script"
+       } else {
+	   bind all <Enter> $script
+       }
+    }  
 }
 
 setPreferenceParms
--- a/gui-tk/minsky.tcl
+++ b/gui-tk/minsky.tcl
@@ -686,8 +686,8 @@ proc logVarsOK {} {
 .menubar.edit add command -label "Undo" -command "undo 1" -accelerator $meta_menu-Z
 .menubar.edit add command -label "Redo" -command "undo -1" -accelerator $meta_menu-Y
 .menubar.edit add command -label "Cut" -command cut -accelerator $meta_menu-X
-.menubar.edit add command -label "Copy" -command minsky.copy -accelerator $meta_menu-C
-.menubar.edit add command -label "Paste" -command minsky.paste -accelerator $meta_menu-V
+.menubar.edit add command -label "Copy" -command "minsky.copy" -accelerator $meta_menu-C
+.menubar.edit add command -label "Paste" -command "minsky.paste" -accelerator $meta_menu-V
 .menubar.edit add command -label "Group selection" -command "minsky.createGroup" -accelerator $meta_menu-G
 .menubar.edit add command -label "Dimensions" -command dimensionsDialog
 
--- a/gui-tk/minsky.tcl
+++ b/gui-tk/minsky.tcl
@@ -687,7 +687,7 @@ proc logVarsOK {} {
 .menubar.edit add command -label "Redo" -command "undo -1" -accelerator $meta_menu-Y
 .menubar.edit add command -label "Cut" -command cut -accelerator $meta_menu-X
 .menubar.edit add command -label "Copy" -command "minsky.copy" -accelerator $meta_menu-C
-.menubar.edit add command -label "Paste" -command pasteAt -accelerator $meta_menu-V
+.menubar.edit add command -label "Paste" -command "minsky.paste" -accelerator $meta_menu-V
 .menubar.edit add command -label "Group selection" -command "minsky.createGroup" -accelerator $meta_menu-G
 .menubar.edit add command -label "Dimensions" -command dimensionsDialog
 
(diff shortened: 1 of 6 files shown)


gui-tk/minsky.tcl
```diff
@@ -703,6 +703,12 @@ proc cut {} {
     minsky.cut
 }
 
+# Paste item(s) at position x and y on the canvas upon click down. For ticket 1098
+proc pasteAt {} {
+	canvas.mouseDown [get_pointer_x .wiring.canvas] [get_pointer_y .wiring.canvas]
```
highperformancecoder: The problem with this approach is that the mouse is currently over the "Paste" menu item, particularly with the defintion on line 690

gui-tk/minsky.tcl
```diff
@@ -786,8 +792,8 @@ bind . <$meta-x> {minsky.cut}
 bind . <$meta-X> {minsky.cut}
 bind . <$meta-c> {minsky.copy}
 bind . <$meta-C> {minsky.copy}
-bind . <$meta-v> {minsky.paste}
-bind . <$meta-V> {minsky.paste}
+bind . <$meta-v> {pasteAt}
+bind . <$meta-V> {pasteAt}
```
highperformancecoder: This would work...

model/minsky.cc
```diff
@@ -265,6 +265,7 @@ namespace minsky
     canvas.setItemFocus(model->addGroup(g));
```
highperformancecoder: OK for the moveTo to work on line 268, yes g needs to be added to model. But you can ditch the call to setItemFocus.

highperformancecoder: Also solves the segfault problem in adjustWires when the two groups don't share an ancester (that should probably be fixed too, but not so urgent).

highperformancecoder:
The problem I see with this approach is that it pastes the clipboard where the mouse pointer currently is - which will be wherever the menu cascade leaves it. My approach has been to "attach" the pasted items to the mouse, requiring a further click to indicate where the items should be positioned. This is a consistent protocol with the copy item operation and the insert item operation.

The alternative is to introduce a concept of "the cursor", which you'd have to move around with the mouse, and where copyitem, paste and insert (and incidentally text items) would place their payloads. This is the wordprocessor model.

I just checked what a drawing program might do, eg LibreOffice. It appears to paste over the top of the previously selected item. Ouch! I like my way better than that.

wdednam:
Please don't merge yet. On copy of an exiting item or selection of items the new copy still ends up on top of the original selection after pasting, and then both remain in selection, not just the new copy, which one wants to move off the original. This is obviously not a problem when cutting and pasting...

model/minsky.cc
```diff
@@ -255,27 +255,47 @@ namespace minsky
...
+    if (copyOfItems.size()%2==0) numPaste = copyOfItems.size()/2-g->numWires();
+    else numPaste = copyOfItems.size()/2-g->numWires()-1;    
     for (auto& i: copyOfItems)
-      {
-        model->addItem(i);
-        canvas.selection.toggleItemMembership(i);
-        assert(!i->ioVar());
+      {		
+		 // Ensure only a single copy of a selection of items exists at time 
+		 if (count<numPaste && model->uniqueItems())
```
highperformancecoder: I don't follow at all what you're doing with numPaste, an uniqueItems is a quite expensive operation, so shouldn't be done in a loop like this.

model/minsky.cc
```diff
@@ -255,27 +255,47 @@ namespace minsky
...
-        model->addItem(i);
-        canvas.selection.toggleItemMembership(i);
-        assert(!i->ioVar());
+      {		
+		 // Ensure only a single copy of a selection of items exists at time 
+		 if (count<numPaste && model->uniqueItems()) 
+		 {	 
+           canvas.model->addItem(i);			  
+           canvas.selection.ensureItemInserted(i);			 
+           canvas.setItemFocus(i);
```
highperformancecoder: Why are we setting itemFocus here? The next iteration undoes it.

model/minsky.cc
```diff
@@ -255,27 +255,47 @@ namespace minsky
 //      throw runtime_error("cannot save to "+fileName);
   }
 
-  void Minsky::paste(float x, float y)
```
highperformancecoder: Aww - I quite liked the idea of passing in coordinates for where to paste the contents...

model/minsky.cc
wdednam: Thanks, your suggestion of clearing the selection first did the trick.

model/minsky.cc
wdednam: I tried to select just one of the items after the loop, via findAny:

    auto oneItem=g->findAny(&Group::items,[&](const ItemPtr& i)
                         {return false;});          
    if (oneItem) canvas.setItemFocus(oneItem);

This just pastes the copy of the selection on top of the original one and then I can't separate them.

I don't know if there is a better method?

model/minsky.cc
wdednam: I also thought it was a good idea, and think I came close to getting it right but feel my understanding of how the coordinates are handled between TCL and Minsky is still on the low side.

gui-tk/wiring.tcl
```diff
@@ -629,7 +629,7 @@ proc contextMenu {x y X Y} {
                 renameVariableInstances
             }
             .wiring.context add command -label "Edit" -command "editItem"
-            .wiring.context add command -label "Copy" -command "canvas.copyItem"
+            .wiring.context add command -label "Copy" -command "copy.canvasItem"
```
highperformancecoder: Where is copy.canvasItem defined?

gui-tk/wiring.tcl
```diff
@@ -650,7 +650,7 @@ proc contextMenu {x y X Y} {
                .wiring.context add command -label "Initialise Random" \
                     -command "initRandom" 
             }
-            .wiring.context add command -label "Copy" -command "canvas.copyItem"
+            .wiring.context add command -label "Copy" -command "copy.canvasItem"
```
highperformancecoder: Where is copy.canvasItem defined?

gui-tk/minsky.tcl
```diff
@@ -703,6 +703,12 @@ proc cut {} {
     minsky.cut
 }
 
+# Paste item(s) at position x and y on the canvas upon click down. For ticket 1098
+proc pasteAt {} {
+	canvas.mouseDown [get_pointer_x .wiring.canvas] [get_pointer_y .wiring.canvas]
+	minsky.paste
```
highperformancecoder: Missing the new coordinate arguments

model/minsky.cc
```diff
@@ -255,7 +255,7 @@ namespace minsky
 //      throw runtime_error("cannot save to "+fileName);
   }
 
-  void Minsky::paste()
+  void Minsky::paste(float x, float y)
```
highperformancecoder: Probably also need a boolean argument to indicate that paste occurred from the edit menu. If so, then the newly pasted items need to be left in the selection, and one item of the selection needs to be left in itemFocus (effectively attaching the selection to the mouse tip). If pasted from ^V or the context menu, the itemFocus should be cleared, and I'm not sure about the selection.

highperformancecoder:
On Tue, Jan 07, 2020 at 04:27:38AM -0800, Wynand wrote:
> 
> I tried to select just one of the items after the loop, via findAny:
> 
> auto oneItem=g->findAny(&Group::items,[&](const ItemPtr& i)
>                      {return false;});
> if (oneItem) canvas.setItemFocus(oneItem);
> 
> This just pastes the copy of the selection on top of the original one and then
> I can't separate them.
> 
> I don't know if there is a better method?
> 

Instead of using findAny, which in this case will not return anything
since you are returning the constant false, just check if g->items is
empty, and if not, just grab the first element (g->items[0]).

Cheers

-- 

                      http://www.hpcoders.com.au

gui-tk/wiring.tcl
wdednam: Thanks for catching this. It must have occurred when I tested minsky.paste here instead of copy.canvasItem which was originally there.

gui-tk/minsky.tcl
highperformancecoder: It seems to me you still want pasteAt for ^V handling.

wdednam: Why is that there? I don't understand. They don't show up in my local bugfix-1098 branch?? I have also added the bugfix-901 changes because you have just merged that PR into your master branch. maybe I should revert and git pull your master into my local bugfix-1098 again? It has become a bit messy...

test/00/t0036a.sh
```diff
@@ -42,7 +42,8 @@ proc afterMinskyStarted {} {uplevel #0 {
   canvas.mouseUp [expr [\$item.x]+0.55*[\$item.width]] [expr [\$item.y]-0.55*[\$item.height]] 
   assert {[canvas.selection.groups.size]==1}
   newSystem
-  paste
+  # For ticket 1098. Paste items at mouse position
+  paste [expr [\$item.x]+0.55*[\$item.width]] [expr [\$item.y]-0.55*[\$item.height]]
```
highperformancecoder: Should revert this change too

PR #118, opened 2020-01-07
Ensure value of variable is preserved after rename. For ticket 1106.

wdednam:
This change is [Reviewable](https://reviewable.io/reviews/highperformancecoder/minsky/118)

### DIFF
--- a/engine/equations.h
+++ b/engine/equations.h
@@ -342,7 +342,7 @@ namespace MathDAG
     /// create a variable DAG. returns cached value if previously called
     NodePtr makeDAG(const string& valueId, const string& name, VariableType::Type type);
     NodePtr makeDAG(VariableBase& v)
-    {v.ensureValueExists(); return makeDAG(v.valueId(),VariableValue::uqName(v.name()),v.type());}
+    {v.ensureValueExists(v.vValue()); return makeDAG(v.valueId(),VariableValue::uqName(v.name()),v.type());}
     /// create an operation DAG. returns cached value if previously called
     NodePtr makeDAG(const OperationBase& op);
     NodePtr makeDAG(const SwitchIcon& op);
--- a/model/canvas.cc
+++ b/model/canvas.cc
@@ -554,10 +554,13 @@ namespace minsky
              if (auto v=(*i)->variableCast())
                if (v->valueId()==valueId)
                  {
+				   // Ensure value of variable is preserved after rename. For ticket 1106.	 
+				   auto tempVal=v->value();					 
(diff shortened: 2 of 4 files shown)


model/canvas.cc
```diff
@@ -554,10 +554,13 @@ namespace minsky
              if (auto v=(*i)->variableCast())
                if (v->valueId()==valueId)
                  {
+				   // Ensure value of variable is preserved after rename. For ticket 1106.	 
+				   auto tempVal=v->value();					 
                    if (auto g=dynamic_cast<GodleyIcon*>(v->controller.lock().get()))
                        // fix up internal references in Godley table
                        g->table.rename(v->rawName(), (v->name()[0]==':'?":":"")+newName);
                    v->name(newName);
+                   v->value(tempVal);
```
highperformancecoder: Whilst this does technically solve the bug as described, I think there are bigger things here. The problem is that when a variable is given a new name, it gets assigned a default constructed variableValue. A variableValue is more than just a floating point number, particularly in the tensor-valued world. I would say the more general solution lies in modifying VariableBase::name():
1. Before assigning m_name, stash a copy of the pointer return by vValue(). 
2. After assigning m_name, if vValue() returns nullptr, rather than calling ensureValueExists, insert a copy constructed VariableValue (from the previously stashed vValue()) into minsky().variableValues. This will look a bit like the ensureValueExists code. If the previously vValue pointer is null, we want to do a regular ensureValueExists(() anyway. So maybe just modify ensureValueExists to take a pointer to a VariableValue and if the argment is not null, use it to copy construct the variableValue

wdednam: Thanks for the very clear instructions. I've just tested it on Steve's example and ran "make sure" and it looks like it works. Will commit the new code soon.

engine/equations.h
```diff
@@ -342,7 +342,7 @@ namespace MathDAG
     /// create a variable DAG. returns cached value if previously called
     NodePtr makeDAG(const string& valueId, const string& name, VariableType::Type type);
     NodePtr makeDAG(VariableBase& v)
-    {v.ensureValueExists(); return makeDAG(v.valueId(),VariableValue::uqName(v.name()),v.type());}
+    {v.ensureValueExists(v.vValue()); return makeDAG(v.valueId(),VariableValue::uqName(v.name()),v.type());}
```
highperformancecoder: Strictly unnecessary to pass anything other than nullptr here and in other places below, except when variables are being renamed. I would have probably made the parameter default to nullptr. But harmless nevertheless (if v.vValue is not null, ensureValueExists is a no operation), so I'll merge the PR.

PR #121, opened 2020-01-11
Stock and flow variables on Godley icons should not be copied as groups.

wdednam:
This change is [Reviewable](https://reviewable.io/reviews/highperformancecoder/minsky/121)


model/canvas.cc
```diff
@@ -629,7 +629,6 @@ namespace minsky
   void Canvas::copyVars(const std::vector<VariablePtr>& v)
   {
     auto group=model->addGroup(new Group);
-    setItemFocus(group);
```
highperformancecoder: I suspect this code can be refactored to not create a new group. Unless we desire to share code between this and paste.

highperformancecoder:
Need to fix the regression test.

model/canvas.cc
```diff
@@ -628,44 +628,40 @@ namespace minsky
...
+          heights.push_back(rv.height());
+          maxWidth=max(maxWidth, widths.back());
+          totalHeight+=heights.back();
+        }
+      float y=v[0]->y() - totalHeight;
+      for (size_t i=0; i<v.size(); ++i)
+        {
+          auto ni=v[i]->clone();
+          ni->rotation=0;
+          //ni->zoomFactor=group->zoomFactor;
```
highperformancecoder: dead code, should be removed

model/canvas.cc
```diff
@@ -628,44 +628,40 @@ namespace minsky
...
+          ni->moveTo(v[0]->x()+maxWidth-v[i]->zoomFactor()*widths[i],
+                     y+heights[i]);
+          // variables need to refer to outer scope
+          if (ni->name()[0]!=':')
+            ni->name(':'+ni->name());
+          y+=2*v[i]->zoomFactor()*heights[i];
+          // Stock and flow variables on Godley icons should not be copied as groups. For ticket 1039
+          ItemPtr newI(ni);
```
highperformancecoder: Not actually necessary to create the shared_ptr here.

model/canvas.cc
```diff
@@ -628,44 +628,40 @@ namespace minsky
...
+          selection.insertItem(model->addItem(newI));		 
+        }
+        // Item focus put on one of the copied vars that are still in selection. For ticket 1039
+        setItemFocus(model->findAny
```
highperformancecoder: What about using selection.items[0]? Obviously check for empty selection first, and clear the itemFocus in that case.

test/testModel.cc
```diff
@@ -776,32 +776,15 @@ SUITE(Canvas)
         godley->table.cell(2,2)="y";
         godley->update();
 
-        
+        // Unit test greatly simplified since flow/stock vars no longer copied via temporary group. For ticket 1039
```
highperformancecoder: This comment is not really helpful for down the track. Add it to the ticket discussion if you want to comment on it.

test/testModel.cc
```diff
@@ -776,32 +776,15 @@ SUITE(Canvas)
...
         unsigned originalNumItems=model->numItems();
-        unsigned originalNumGroups=model->numGroups();
         copyAllFlowVars();
         CHECK_EQUAL(originalNumItems+godley->flowVars().size(),model->numItems());
```
highperformancecoder: This is a good check. But should also check that the relevant items have been inserted into selection, and also that  (I'm guessing in this case) that there are now two copies of the flowVars in model. The way I would approach this is do a loop of model->items, and increment a map<string,int> where the key is the valueId, then afterwards, loop over the flowVars and check that the count is 2. eg

map<string,int> idCnt;
for (auto& i: model->items)
  if (auto v=i.variableCast())
     idCnt[v->valueId()]++;
for (auto v: godley->flowVars())
    CHECK_EQUAL(2, idCnt[v->valueId()])

model/canvas.cc
wdednam: Thanks. I've removed it.

model/canvas.cc
wdednam: Also fixed this. I've made ni an ItemPtr directly as you suggested, and to access members of variablePtr I use variableCast(), e.g., (ni->variableCast())->name().

model/canvas.cc
wdednam: Thanks, done. I forgot about that member of selection.

test/testModel.cc
wdednam: I've removed it, thanks.

test/testModel.cc
wdednam: Thank you for this, the test is much more robust now. I will soon commit the new changes.

model/canvas.cc
```diff
@@ -628,32 +628,37 @@ namespace minsky
...
+                     y+heights[i]);
+          // variables need to refer to outer scope
+          if ((ni->variableCast())->name()[0]!=':')
+            (ni->variableCast())->name(':'+(ni->variableCast())->name());
+          y+=2*v[i]->zoomFactor()*heights[i];
+          selection.insertItem(model->addItem(ni));		 
+        }
+        // Item focus put on one of the copied vars that are still in selection. For ticket 1039
+        if (!selection.empty()) setItemFocus(selection.items[0]);
+        else selection.clear();
```
highperformancecoder: Actually, selection is empty, so doesn't need to be cleared. However, itemFocus should probably be cleared too (setItemFocus(nullptr) perhaps?).

wdednam: Thanks, I will fix this after your current commit finishes and I can merge the most up to date version into this branch.

highperformancecoder: OK - I've finished with #1113 now. There won't be any merge conflicts with this PR.

wdednam: I don't know why test t0001a.sh fails, it doesn't on my machine.

model/canvas.cc
```diff
@@ -677,32 +677,37 @@ namespace minsky
...
+          widths.push_back(rv.width());
+          heights.push_back(rv.height());
+          maxWidth=max(maxWidth, widths.back());
+          totalHeight+=heights.back();
+        }
+      float y=v[0]->y() - totalHeight;
+      for (size_t i=0; i<v.size(); ++i)
+        {
+		  // Stock and flow variables on Godley icons should not be copied as groups. For ticket 1039	
+          ItemPtr ni(v[i]->clone());
```
highperformancecoder: I would add an assert(ni->variableCast()) here as a protection. Yes the current code guarantees that ni is actually a VariableBase, but things have a habit of changing :).

Also variableCast() is not without cost. It is cheaper than dynamic_cast. In this case, you could use static_cast, which is free. Mind you, this is not really performance critical code.

PR #122, opened 2020-01-11
Make text dialog boxes wider to facilitate grabbing them in Windows.

wdednam:
This change is [Reviewable](https://reviewable.io/reviews/highperformancecoder/minsky/122)

### DIFF
--- a/gui-tk/minsky.tcl
+++ b/gui-tk/minsky.tcl
@@ -849,7 +849,7 @@ proc textEntryPopup {win init okproc} {
     if {![winfo exists $win]} {
         toplevel $win
         entry $win.entry
-        pack $win.entry -side top
+        pack $win.entry -side top -ipadx 50
         buttonBar $win $okproc
     } else {
         wm deiconify $win
--- a/gui-tk/minsky.tcl
+++ b/gui-tk/minsky.tcl
@@ -782,11 +782,6 @@ proc dimFormatPopdown {comboBox type} {
     }
 }
 
-proc pasteAt {} {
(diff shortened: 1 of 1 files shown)


gui-tk/minsky.tcl
```diff
@@ -782,11 +782,6 @@ proc dimFormatPopdown {comboBox type} {
     }
 }
 
-proc pasteAt {} {
```
highperformancecoder: You've got merge issues again. Try remerging master into this branch

wdednam: As I mentioned on sourceforge, I'm confused about why pasteAt is still here. I thought I had removed any trace of it in a previous PR. I know why my PRs get mixed up now, because I was creating new bugfix branches from existing ones instead of from the most up-to-date HEAD/master branch. I assumed all new branches were created from the master by default. Won't do that again.

highperformancecoder: pasteAt is back. As I mentioned on a previous ticket, the requestRedraw() at the end of patse didn't do the trick, as the pasted items were off screen. They need to be moved on screen, and the simplest way of doing that is to call mouseMove(), which I ended up putting in a revised pasteAt.

wdednam: I understand now, sorry. I didn't when you first mentioned it to me in the other ticket, because I hadn't experienced that behaviour during the limited testing I did. Thanks for fixing that up. I will merge the most recent master branch into all my other bugfix branches before committing new code.

PR #124, opened 2020-01-14
Grey out (disable) paste option when clipboard is empty.

wdednam:
This change is [Reviewable](https://reviewable.io/reviews/highperformancecoder/minsky/124)

### DIFF
--- a/gui-tk/godley.tcl
+++ b/gui-tk/godley.tcl
@@ -88,19 +88,14 @@ proc openGodley {id} {
         menu .$id.menubar.file
         .$id.menubar.file add command -label "Export" -command "exportGodley $id"
         
-        menu .$id.menubar.edit
+        menu .$id.menubar.edit -postcommand "toggleGodleyPaste $id"
         .$id configure -menu .$id.menubar
         .$id.menubar.edit add command -label Undo -command "$id.undo 1" -accelerator $meta_menu-Z
         .$id.menubar.edit add command -label Redo -command "$id.undo -1" -accelerator $meta_menu-Y
         .$id.menubar.edit add command -label Title -command "textEntryPopup .godleyTitle {[$id.godleyIcon.table.title]} {setGodleyTitleOK $id}"
         .$id.menubar.edit add command -label Cut -command "$id.cut" -accelerator $meta_menu-X
         .$id.menubar.edit add command -label Copy -command "$id.copy" -accelerator $meta_menu-C
-        .$id.menubar.edit add command -label Paste -command "$id.paste" -accelerator $meta_menu-V
-	     if {![catch {clipboard get -displayof .$id.context -type UTF8_STRING}]} {
-			.$id.menubar.edit entryconfigure end -state normal
-	     } else {
-	     	.$id.menubar.edit entryconfigure end -state disabled
-	     }                
+        .$id.menubar.edit add command -label Paste -command "$id.paste" -accelerator $meta_menu-V               
 
         menu .$id.menubar.view
         .$id.menubar.view add command -label "Zoom in" -command "zoomIn $id" -accelerator $meta_menu-+
@@ -126,6 +121,12 @@ proc openGodley {id} {
     }
     wm deiconify .$id
     raise .$id .
+}
+
+proc toggleGodleyPaste id {
+	if {[catch {clipboard get -type UTF8_STRING}]} {
+	.$id.menubar.edit entryconfigure end -state disabled
+	}
 }
 
 proc zoomOut id {
@@ -242,12 +243,11 @@ proc godleyContext {id x y X Y} {
             .$id.context add command -label "Copy" -command "$id.copy"
         }
     }
-    if {![catch {clipboard get -type UTF8_STRING}]  && ($r!=1 || $c!=0)} {   # Cannot Paste into cell(1,0). For ticket 1064
+    if {($r!=1 || $c!=0)} {   # Cannot Paste into cell(1,0). For ticket 1064
         .$id.context add command -label "Paste" -command "$id.paste"
-        
-    } else {
-		.$id.context add command -label "Paste" -command "$id.paste"
-		.$id.context entryconfigure end -state disabled
+        if {[catch {clipboard get -type UTF8_STRING}]} {
+			.$id.context entryconfigure end -state disabled 
+           }
 	}
     tk_popup .$id.context $X $Y
 }
--- a/gui-tk/godley.tcl
+++ b/gui-tk/godley.tcl
@@ -96,6 +96,11 @@ proc openGodley {id} {
         .$id.menubar.edit add command -label Cut -command "$id.cut" -accelerator $meta_menu-X
         .$id.menubar.edit add command -label Copy -command "$id.copy" -accelerator $meta_menu-C
         .$id.menubar.edit add command -label Paste -command "$id.paste" -accelerator $meta_menu-V
+	     if {![catch {clipboard get -displayof .$id.context -type UTF8_STRING}]} {
+			.$id.menubar.edit entryconfigure end -state normal
+	     } else {
+	     	.$id.menubar.edit entryconfigure end -state disabled
+	     }                
 
         menu .$id.menubar.view
         .$id.menubar.view add command -label "Zoom in" -command "zoomIn $id" -accelerator $meta_menu-+
@@ -239,7 +244,11 @@ proc godleyContext {id x y X Y} {
     }
     if {![catch {clipboard get -type UTF8_STRING}]  && ($r!=1 || $c!=0)} {   # Cannot Paste into cell(1,0). For ticket 1064
         .$id.context add command -label "Paste" -command "$id.paste"
-    }
+        
+    } else {
+		.$id.context add command -label "Paste" -command "$id.paste"
+		.$id.context entryconfigure end -state disabled
+	}
     tk_popup .$id.context $X $Y
 }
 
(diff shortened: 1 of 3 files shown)


gui-tk/godley.tcl
```diff
@@ -96,6 +96,11 @@ proc openGodley {id} {
         .$id.menubar.edit add command -label Cut -command "$id.cut" -accelerator $meta_menu-X
         .$id.menubar.edit add command -label Copy -command "$id.copy" -accelerator $meta_menu-C
         .$id.menubar.edit add command -label Paste -command "$id.paste" -accelerator $meta_menu-V
+	     if {![catch {clipboard get -displayof .$id.context -type UTF8_STRING}]} {
```
highperformancecoder: The problem with this approach is that your setting the menu item state according to the state of the clipboard at window creation. What you need to do is set the state at menu post time.

gui-tk/godley.tcl
```diff
@@ -239,7 +244,11 @@ proc godleyContext {id x y X Y} {
     }
     if {![catch {clipboard get -type UTF8_STRING}]  && ($r!=1 || $c!=0)} {   # Cannot Paste into cell(1,0). For ticket 1064
         .$id.context add command -label "Paste" -command "$id.paste"
-    }
+
```
highperformancecoder: It works here, because this method constructs and posts the menu immediately after.

However, it would be simpler to create the menu item (via the add command) outside of the conditional, then just have the entryconfigure command inside the conditional (more DRY).

gui-tk/minsky.tcl
```diff
@@ -704,6 +704,11 @@ proc logVarsOK {} {
 .menubar.edit add command -label "Cut" -command cut -accelerator $meta_menu-X
 .menubar.edit add command -label "Copy" -command "minsky.copy" -accelerator $meta_menu-C
 .menubar.edit add command -label "Paste" -command "minsky.paste" -accelerator $meta_menu-V
+if {![catch {clipboard get -displayof .menubar.edit -type UTF8_STRING}]} {
```
highperformancecoder: This has the same problem as I earlier mentioned. You need to do this at post time. There's no examples of doing this in the codebase, but if you consult the menu man page, you'll see that you should be able to attach a handler to run at post time via the -postcommand menu creation option.

gui-tk/wiring.tcl
```diff
@@ -500,6 +500,11 @@ proc canvasContext {x y X Y} {
     .wiring.context add command -label "Copy selection" -command "minsky.copy"
     .wiring.context add command -label "Save selection as" -command saveSelection
     .wiring.context add command -label "Paste selection" -command pasteAt
+	if {![catch {clipboard get -type UTF8_STRING}]} {
+		.wiring.context entryconfigure end -state normal
```
highperformancecoder: This statement is redundant (by default, the previous line creates the menu item in normal state).

gui-tk/godley.tcl
wdednam: I've been struggling with this one most of today. The online [documentation](https://www.tcl.tk/man/tcl8.4/TkCmd/menu.htm#M7)  is not clear. The furthest I've got is:

In godley.tcl:90

`        menu .$id.menubar.edit -postcommand {
	     if {[catch {clipboard get -displayof .$id.context -type UTF8_STRING}]} {
			.$id.menubar.edit entryconfigure end -state normal
			}
		}             
        .$id configure -menu .$id.menubar  
        .$id.menubar.edit add command -label Undo -command "$id.undo 1" -accelerator $meta_menu-Z
        .$id.menubar.edit add command -label Redo -command "$id.undo -1" -accelerator $meta_menu-Y
        .$id.menubar.edit add command -label Title -command "textEntryPopup .godleyTitle {[$id.godleyIcon.table.title]} {setGodleyTitleOK $id}"
        .$id.menubar.edit add command -label Cut -command "$id.cut" -accelerator $meta_menu-X
        .$id.menubar.edit add command -label Copy -command "$id.copy" -accelerator $meta_menu-C
        .$id.menubar.edit post 5     
        .$id.menubar.edit add command -label Paste -command "$id.paste" -accelerator $meta_menu-V`

which gives me the following error when I click on Edit in the Godley window menubar:

can't read "id": no such variable
    while executing
".$id.menubar.edit entryconfigure end -state normal"

gui-tk/godley.tcl
wdednam: This was originally here. If I do  it the other way round, first add command and then if conditional

    if {($r!=1 || $c!=0)} {   # Cannot Paste into cell(1,0). For ticket 1064
        .$id.context add command -label "Paste" -command "$id.paste"
        if {[catch {clipboard get -type UTF8_STRING}]} {
			$id.context entryconfigure end -state disabled 
           }
	}


 I get an error when the clipboard is empty:

invalid command name "minsky.godleyWindow94383534427296.context"
    while executing
"tclunknown minsky.godleyWindow94383534427296.context entryconfigure end -state disabled"

gui-tk/minsky.tcl
wdednam: > you should be able to attach a handler to run at post time via the -postcommand menu creation option

This is where I'm stuck at present, the man page unfortunately does not specify what type of handler to pass to -postcommand.

gui-tk/wiring.tcl
wdednam: Thanks, I'm leaving it simply as:

    .wiring.context add command -label "Paste selection" -command pasteAt
	if {[catch {clipboard get -type UTF8_STRING}]} {
		.wiring.context entryconfigure end -state disabled
	}

gui-tk/godley.tcl
highperformancecoder: This is one of the most irritating things about TCL - understanding how expansion works, ie the interaction between $, {}, "" and [].

In this case, $id needs to be evaluated at menu instantiation time, not at menu post time. Wrapping a $id in a {} defers expansion until  the code with the braces is run, ie at post time.

The best bet is to create another proc, say editProc, eg:
proc editProc id {
   if {[catch ...
}

and then set menu .$id.menubar.edit -postcommand "editProc $id". Note the use of quotes, which causes id to be immediate expanded, ie at menu instantiation time.

gui-tk/godley.tcl
highperformancecoder: That's because of the $id in braces thing again.

gui-tk/minsky.tcl
highperformancecoder: Any TCL command

gui-tk/godley.tcl
highperformancecoder: I don't understand why you're getting this message when the previous line (with add) is working just fine.

gui-tk/minsky.tcl
wdednam: Understood, thanks.

gui-tk/godley.tcl
wdednam: Thanks, this has helped a lot. I've also now seen the -postcommand example applied to getBookMarkMenu in minsky.tcl, which is where I decided to start since it is easier than the godley.tcl case. So, now I don't get any error messages, but the last entry on the cascade menu gets greyed out and not the one I want. Here's what I did and thought would work for me:

menu .menubar.edit -postcommand "editProc window"
.menubar add cascade -menu .menubar.edit -label Edit -underline 0

.menubar.edit add command -label "Undo" -command "undo 1" -accelerator $meta_menu-Z
.menubar.edit add command -label "Redo" -command "undo -1" -accelerator $meta_menu-Y
.menubar.edit add command -label "Cut" -command cut -accelerator $meta_menu-X
.menubar.edit add command -label "Copy" -command "minsky.copy" -accelerator $meta_menu-C
.menubar.edit add command -label "Paste" -command "minsky.paste" -accelerator $meta_menu-V
.menubar.edit add command -label "Group selection" -command "minsky.createGroup" -accelerator $meta_menu-G
.menubar.edit add command -label "Dimensions" -command dimensionsDialog

proc editProc {window} {
	if {[catch {clipboard get -displayof $window -type UTF8_STRING}]} {
	.menubar.edit entryconfigure 4 -state disabled
	}
}

I thought that placing the index 4 (counting from 0) after entryconfigure would grey out the corresponding entry (Paste) of the edit cascade menu? I think I'm finally on the right track and hope to have it working next time you take a look at it.

gui-tk/godley.tcl
wdednam: I fixed this, there was a . missing in front of $id.context entryconfigure end -state disabled.

gui-tk/godley.tcl
wdednam: I got this functionality working in godley.tcl. Still having the same problem in the minsky.tcl I mentioned in the previous post. Working on it.

gui-tk/godley.tcl
```diff
@@ -123,6 +123,12 @@ proc openGodley {id} {
     raise .$id .
 }
 
+proc toggleGodleyPaste id {
```
highperformancecoder: Is this dead code then?

gui-tk/minsky.tcl
```diff
@@ -707,6 +707,12 @@ proc logVarsOK {} {
 .menubar.edit add command -label "Group selection" -command "minsky.createGroup" -accelerator $meta_menu-G
 .menubar.edit add command -label "Dimensions" -command dimensionsDialog
 
+proc togglePaste {window} {
+	if {[catch {clipboard get -type UTF8_STRING}]} {
+	.menubar.edit entryconfigure "Paste" -state disabled
```
highperformancecoder: Excellent - maybe we should use "Paste" instead of end for the toggleGodleyPaste proc as well...

PR #125, opened 2020-01-15
Catch exception 'Unknown variable XX in initialisation of YY' at inpu…

wdednam:
…t time.

---
This change is [Reviewable](https://reviewable.io/reviews/highperformancecoder/minsky/125)

### DIFF
--- a/model/variable.cc
+++ b/model/variable.cc
@@ -204,8 +204,12 @@ string VariableBase::_init(const string& x)
               (type()==flow && !cminsky().definingVar(valueId()))) 
             val.reset(minsky().variableValues);
         }
-      catch (...)
-        {}
+      // Catch exception "Unknown variable XX in initialisation of YY" at input time. For ticket 1045.
+      catch (const std::exception& e)
+        {
+		    val.init.clear();
+			throw runtime_error(e.what());
+		}
     }
   return x;
 }


wdednam:
Not such an easy fix after all. I forgot that variables can also have symbolic values in, e.g., Godley tables. So, this fix is going into the bin... 

Is it perhaps better that the exception get caught at reset time, i.e., leave things as they were?

highperformancecoder:
Which bug are you trying to fix?

On Wed, Jan 15, 2020 at 03:04:36PM -0800, Wynand wrote:
> Not such an easy fix after all. I forgot that variables can also have symbolic
> values in, e.g., Godley tables. So, this fix is going into the bin...
> 
> Is it perhaps better that the exception get caught at reset time, i.e., leave
> things as they were?
> 
> —
> You are receiving this because you are subscribed to this thread.
> 

-- 

                      http://www.hpcoders.com.au

model/variable.cc
```diff
@@ -204,8 +204,12 @@ string VariableBase::_init(const string& x)
               (type()==flow && !cminsky().definingVar(valueId()))) 
             val.reset(minsky().variableValues);
         }
-      catch (...)
-        {}
+      // Catch exception "Unknown variable XX in initialisation of YY" at input time. For ticket 1045.
```
highperformancecoder: I think we do want this error message to propagate to the user, which is why I think this fix doesn't work.

wdednam: I also think it is better for the exception to be caught on variableValue reset, which, as you allude to in the same member, is related to an older ticket (#433). I guess the bug becomes less important, because as a result of some posterior fix, it is now possible to edit the variable whose value disappeared permanently and was previously uneditable. But it doesn't look good. So, would a requestRedraw() fix the disappearing values on the other variables?

wdednam:
I think we should close this PR, which does not solve the bug. Maybe we can lower the priority of the bug now that the variables remain editable even though the values on them disappear until model run and stop, or model reset?

PR #126, opened 2020-01-19
Update the bounding box of icons which have been rotated to permit dr…

wdednam:
…agging wires from their out ports. I tried to add a class ItemData from which Item (containing the accessor member lambda) inherits other member functions/variables, but could not manage to compile the code without errors where members from one class could not be converted to the other class type. I moved the complaining members around in attempt to try to fix the compile errors, but unfortunately did not succeed. Thus, I have stuck with the initial Item class we discussed by email.

---
This change is [Reviewable](https://reviewable.io/reviews/highperformancecoder/minsky/126)

### DIFF
--- a/model/item.h
+++ b/model/item.h
@@ -70,16 +70,16 @@ namespace minsky
       // extend each item by a portradius to solve ticket #903
       return left-portRadius<=x && right+portRadius>=x && bottom+portRadius>=y && top-portRadius<=y;
     }
-    bool valid() const {return left!=right;}
+    bool valid() const {return right-left!=2*portRadius;}
     float width() const {return right-left;}
     float height() const {return bottom-top;}
   };
 
   class Item: virtual public NoteBase
   {
+	double m_rotation=0; ///< rotation of icon, in degrees  
   public:
     float m_x=0, m_y=0; ///< position in canvas, or within group
-    double rotation=0; ///< rotation of icon, in degrees
     mutable bool onResizeHandles=false; ///< set to true to indicate mouse is over resize handles
     /// owning group of this item.
     classdesc::Exclude<std::weak_ptr<Group>> group; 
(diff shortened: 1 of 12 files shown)


model/item.h
```diff
@@ -70,16 +70,16 @@ namespace minsky
       // extend each item by a portradius to solve ticket #903
       return left-portRadius<=x && right+portRadius>=x && bottom+portRadius>=y && top-portRadius<=y;
     }
-    bool valid() const {return left!=right;}
+    bool valid() const {return right-left!=2*portRadius;}
```
highperformancecoder: What's the reasoning here? On initialisation, left=right=0, so with your change, initialliy the bounding box is valid. That doesn't seem right.

model/item.h
```diff
@@ -94,13 +94,29 @@ namespace minsky
...
+    virtual double _rotation() const;  
+    virtual double _rotation(double);
+    ecolab::Accessor<double> rotation {
+      [this]() {return _rotation();},
+        [this](double x){return _rotation(x);}};
+    /// @} 
+    
+    
+    Item() {}
+    Item(const Item& x): m_rotation(x.m_rotation) {}
```
highperformancecoder: You need to initialise all members and base classes to the values coming from x, not just m_rotation. It is the tediousness of that that I was trying to avoid with the ItemData trick we discussed.

model/item.h
```diff
@@ -94,13 +94,29 @@ namespace minsky
...
+    Item& operator=(const Item& x) {
```
highperformancecoder: Ditto with assignment, however since Accessor defines assignment, the default compiler generated version suffices.

model/operation.cc
```diff
@@ -82,7 +82,7 @@ namespace
 
     // puts a small symbol to identify port
     // x, y = position of symbol
-    void drawPort(void (DrawBinOp::*symbol)() const, float x, float y, float rotation)  const
+    void drawPort(void (DrawBinOp::*symbol)() const, float x, float y, double rotation)  const
```
highperformancecoder: Good catch.

model/item.h
wdednam: True. I became confused because of the addition of the portRadius on all sides of the bounding box. Reverting it to its previous definition.

model/item.h
wdednam: Thanks. I understand better now. I will try to implement the intermediate ItemData class again tomorrow. Unfortunately, all my attempts at it today resulted in failed compìlation.

schema/schema1.h
```diff
@@ -316,12 +316,12 @@ namespace schema1
...
                                 Join<PositionLayout, VisibilityLayout> >
   {
-    double rotation=0;
-
+	double rotation=0;
     ItemLayout() {}
-    template <class T> ItemLayout(int id, const T& item): 
-      Layout(id), PositionLayout(id, item), VisibilityLayout(item),
-      rotation(item.rotation) {}
+    template <class T> ItemLayout(const T& item): rotation(item.rotation){}
```
highperformancecoder: There shouldn't be any need to update anything in schema1.

wdednam: I guess once I get the ItemData class working, the problem here will go away.

highperformancecoder: The way the schema classes work is that you can construct a schema n from a schema n-1. That ensures backward compatibility. Finally, for the current schema, you can construct a minsky::Minsky object and construct a current schema object from a minsky::Minsky object. So this means that older schemas (eg schema1) should know nothing about anything in minsky namespace. If they do, it is probably some deprecated code that could and should be removed.

wdednam:
I'm going to close this pull request, as it is no longer current.

PR #129, opened 2020-01-22
Possible fix to 'non matching column' bug

wdednam:
The idea is to prevent balanceDuplicateColumn() in minsky.cc from checking for a match between a noAssetClass column and either a asset or liability column, which is what I think is causing the bug. But since I don't often see this particular bug and can't reproduce it at will, this fix may be a long shot.

---
This change is [Reviewable](https://reviewable.io/reviews/highperformancecoder/minsky/129)

### DIFF
--- a/model/godleyTableWindow.cc
+++ b/model/godleyTableWindow.cc
@@ -756,7 +756,7 @@ namespace minsky
   {
     x/=zoomFactor;
     int c=colX(x);
-    if (c>0)
+    if (c>0 && godleyIcon->table._assetClass(c)!=GodleyAssetClass::noAssetClass)
       {
         godleyIcon->table.cell(0,c)=name;
         minsky().importDuplicateColumn(godleyIcon->table, c);
@@ -882,7 +882,7 @@ namespace minsky
 
   void GodleyTableWindow::update()
   {
-    if (selectedCol>0 && selectedCol<int(godleyIcon->table.cols()))
+    if (selectedCol>0 && selectedCol<int(godleyIcon->table.cols()) && godleyIcon->table._assetClass(selectedCol)!=GodleyAssetClass::noAssetClass)
       {
         if (selectedRow==0)
           {
--- a/model/godleyTableWindow.cc
+++ b/model/godleyTableWindow.cc
@@ -825,9 +825,9 @@ namespace minsky
   void GodleyTableWindow::pushHistory()
   {
     while (history.size()>maxHistory) history.pop_front();
-    // Include asset class of each column in history at avoid spurious noAssetClass columns. For ticket 1118.
-    if (history.empty() || history.back().first!=godleyIcon->table.getData()) {
-      history.push_back(make_pair(godleyIcon->table.getData(), godleyIcon->table._assetClass()));
+    // Perform deep comparison of Godley tables in history to avoid spurious noAssetClass columns from arising during undo. For ticket 1118.
+    if (history.empty() || !(history.back()==godleyIcon->table)) {
+      history.push_back(godleyIcon->table);
     }
     historyPtr=history.size();
   }
@@ -839,16 +839,10 @@ namespace minsky
     historyPtr-=changes;
     if (historyPtr > 0 && historyPtr <= history.size())
       {
-        auto& d1=history[historyPtr-1].first;
-        auto& d2=history[historyPtr-1].second;
-        // Include asset class of each column in history at avoid spurious noAssetClass columns. For ticket 1118.
-        if (d1.empty() || d2.empty()) return; // should not happen
-        godleyIcon->table.resize(d1.size(), d1[0].size());
-        for (size_t r=0; r<godleyIcon->table.rows(); ++r)
-          for (size_t c=0; c<godleyIcon->table.cols(); ++c) {
-			godleyIcon->table.cell(r,c)=d1[r][c];
-			godleyIcon->table._assetClass(c,d2[c]);
-		}
+        auto& d=history[historyPtr-1];
+        // Perform deep comparison of Godley tables in history to avoid spurious noAssetClass columns from arising during undo. For ticket 1118.
+        if (d.getData().empty()) return; // should not happen
+		godleyIcon->table=d; 
         requestRedraw();
       }
   }
--- a/model/godleyTableWindow.cc
+++ b/model/godleyTableWindow.cc
@@ -466,7 +466,7 @@ namespace minsky
            {
              selectIdx=insertIdx = textIdx(x);
              auto& str=godleyIcon->table.cell(selectedRow,selectedCol);                         
-             savedText=str;
+             savedText=godleyIcon->table.cell(selectedRow, selectedCol);
            }
         else
           selectIdx=insertIdx=0;
--- a/model/godleyTableWindow.cc
+++ b/model/godleyTableWindow.cc
@@ -466,7 +466,7 @@ namespace minsky
            {
              selectIdx=insertIdx = textIdx(x);
              auto& str=godleyIcon->table.cell(selectedRow,selectedCol);                         
-             savedText=godleyIcon->table.cell(selectedRow, selectedCol);
+             savedText=str;
            }
         else
           selectIdx=insertIdx=0;
@@ -756,8 +756,7 @@ namespace minsky
   {
     x/=zoomFactor;
     int c=colX(x);
-    // Prevents a noAssetClass column to be matched with an asset or liability column. See ticket 1118.
-    if (c>0 && godleyIcon->table._assetClass(c)!=GodleyAssetClass::noAssetClass)
+    if (c>0)
       {
         godleyIcon->table.cell(0,c)=name;
         minsky().importDuplicateColumn(godleyIcon->table, c);
@@ -826,8 +825,10 @@ namespace minsky
   void GodleyTableWindow::pushHistory()
   {
     while (history.size()>maxHistory) history.pop_front();
-    if (history.empty() || history.back()!=godleyIcon->table.getData())
-      history.push_back(godleyIcon->table.getData());
+    // Include asset class of each column in history at avoid spurious noAssetClass columns. For ticket 1118.
+    if (history.empty() || (history.back().first!=godleyIcon->table.getData() && history.back().second!=godleyIcon->table._assetClass())) {
+      history.push_back(make_pair(godleyIcon->table.getData(), godleyIcon->table._assetClass()));
+    }
     historyPtr=history.size();
   }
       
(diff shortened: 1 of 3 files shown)


highperformancecoder:
Do we have a theory for how noAssetClass columns get created in the first place? They are really a legacy of an earlier design that didn't have the asset/liability/equity classification. Like we must continue to have a deprecated "constant" operation for backward compatibility purposes, the noAssetClass enumerator must exist, but never appear in active Minsky modesl.

model/godleyTableWindow.cc
```diff
@@ -825,8 +825,10 @@ namespace minsky
   void GodleyTableWindow::pushHistory()
   {
     while (history.size()>maxHistory) history.pop_front();
-    if (history.empty() || history.back()!=godleyIcon->table.getData())
-      history.push_back(godleyIcon->table.getData());
+    // Include asset class of each column in history at avoid spurious noAssetClass columns. For ticket 1118.
+    if (history.empty() || (history.back().first!=godleyIcon->table.getData() && history.back().second!=godleyIcon->table._assetClass())) {
+      history.push_back(make_pair(godleyIcon->table.getData(), godleyIcon->table._assetClass()));
```
highperformancecoder: This is a fair point, however I believe it would be better to push the entire table object onto history. Not only does it have these two items (data, asset classes) but also title and the deprecated doubleEntryCompliant flag (which we've still to get rid of).

model/godleyTableWindow.h
```diff
@@ -67,7 +68,7 @@ namespace minsky
       godleyIcon(godleyIcon), idx(idx) {}
   };
 
-  class GodleyTableWindow: public ecolab::CairoSurface, public ButtonWidgetEnums
+  class GodleyTableWindow: public ecolab::CairoSurface, public ButtonWidgetEnums, public GodleyAssetClass
```
highperformancecoder: The main reason for inheriting from a class enum is to gain access to enumerator labels. I don't think the small simplification of the type declaration on line 173 warrants this.

model/godleyTableWindow.cc
```diff
@@ -837,12 +839,16 @@ namespace minsky
     historyPtr-=changes;
     if (historyPtr > 0 && historyPtr <= history.size())
       {
-        auto& d=history[historyPtr-1];
+        auto& d=history[historyPtr-1].first;
```
highperformancecoder: If we make the change above to stashing full GodleyTable objects, then this bit of code becomes simple assignment.

model/godleyTableWindow.cc
wdednam: That's a very good suggestion, thanks. I will implement this today. The test t0001a.sh fails for some weird reason, which doesn't make sense, because all I've done is to stash the asset class of each column. In my development folder, the test is failing all over the place suddenly. I don't know what changes I made to cause it. I reverted the last few commits to see, but it didn't make a difference...

model/godleyTableWindow.h
wdednam: If I don't include it, I get the following error:

`model/godleyTableWindow.h:175:56: error: ‘AssetClass’ was not declared in this scope
     std::deque<std::pair<GodleyTable::Data,std::vector<AssetClass>>> history;
                                                        ^~~~~~~~~~
model/godleyTableWindow.h:175:56: note: suggested alternative: ‘Tk_SetClass’
     std::deque<std::pair<GodleyTable::Data,std::vector<AssetClass>>> history;
                                                        ^~~~~~~~~~
`

This is after having included the headers of godleyTable and assetClass in godleyTableWindow.h. Very weird...

model/godleyTableWindow.cc
wdednam: I took me quite a while to get the above to work, so again, thanks for the suggestion.  It will be a  boon to simplify the code to something similar to the original, in terms of keeping Minsky streamlined and efficient too.

model/godleyTableWindow.cc
wdednam: Implementing the above idea shortens the code dramatically, except that it does not do anything. Because it is very short I put it here:

godleyTableWindow.h:

`std::deque<GodleyTable> history;`

godleyTableWindow.cc:

`  void GodleyTableWindow::pushHistory()
  {
    while (history.size()>maxHistory) history.pop_front();
    if (history.empty() || (&history.back()!=&godleyIcon->table)) {
      history.push_back(godleyIcon->table);
    }
    historyPtr=history.size();
  }
      
  void GodleyTableWindow::undo(int changes)
  {
    if (historyPtr==history.size())
      pushHistory();
    historyPtr-=changes;
    if (historyPtr > 0 && historyPtr <= history.size())
      {
        auto& d=history[historyPtr-1];
        if (!&d) return; // should not happen
	godleyIcon->table=d;
        requestRedraw();
      }
  }`

Then, I have also tried to pass a "currTable" in pushHistory(), which also required editing minskyTCLObj.h on line 53:

`        if (!t || (!t->is_const && (!t->is_setterGetter || argc>1)))
            if (auto gtw=t->memberPtrCasted<GodleyTableWindow>()) {
              auto g=t->memberPtrCasted<GodleyTable>(); 
              gtw->pushHistory(*g);
            }`

Then, in godleyTableWindow.h:

`void pushHistory(const GodleyTable& currTable);`

and in godleyTableWindow.cc:

`  void GodleyTableWindow::pushHistory(const GodleyTable& currTable)
  {
    while (history.size()>maxHistory) history.pop_front();
    if (history.empty() || (&history.back()!=&currTable)) {
      history.push_back(currTable);
    }
    historyPtr=history.size();
  }
      
  void GodleyTableWindow::undo(int changes)
  {
    if (historyPtr==history.size())
      pushHistory(godleyIcon->table);
    historyPtr-=changes;
    if (historyPtr > 0 && historyPtr <= history.size())
      {
        auto& d=history[historyPtr-1];
        if (!&d) return; // should not happen
	godleyIcon->table=d;
        requestRedraw();
      }
  }`

This produces a segmentation fault that I think has to do with the fact that the objects in gtw->pushHistory(*g) in minskyTCLObj.h and pushHistory(godleyIcon->table) in godleyWindowTable::undo() are not the same? From backtrace in gdb:

(gdb) frame 7
#7  0x00005555556a962a in minsky::member_entry_hook<char const**> (argc=2, argv=0x55555656be90) at gui-tk/minskyTCLObj.h:53
**53	              gtw->pushHistory(*g);**
(gdb) frame 6
#6  0x0000555555b3d9a4 in minsky::GodleyTableWindow::pushHistory (this=0x555556b75400, currTable=...) at model/godleyTableWindow.cc:831
**831	      history.push_back(currTable);**

wdednam: I've tried several approaches, including using a weak_ptr for the history, and using either the GodleyTable or GodleyIcon objects to define the history, but when it successfully compiles, nothing happens when I press ctrl+Z, the changes cannot be undone. I think my lack of understanding of how these objects work prevents me from replacing the actual godleyIcon or godleyIcon->table objects with one of the elements of the history object. 

On a positive note, I've managed to fix my previous attempt (using a std::deque of GodleyTable::Data and vector<GodleyAssetClass::AssetClass> pairs) so that it now passes all the tests. I hope it fixes at least some of the Godley table bugs.

highperformancecoder: > Implementing the above idea shortens the code dramatically, except that it does not do anything. Because it is very short I put it here:
> 
> godleyTableWindow.h:
> 
> `std::deque<GodleyTable> history;`
> 
> godleyTableWindow.cc:
> 
> ` void GodleyTableWindow::pushHistory()
> {
> while (history.size()>maxHistory) history.pop_front();
> if (history.empty() || (&history.back()!=&godleyIcon->table)) {

It seems like mentally you think you are storing pointers to the Godley table. Whereas. you are actually making deep copies. Which you have to do - storing pointers won't work. At this point you need to test for deep equality, not identity. Now the compiler doesn't automatically generate an equality operator, (thats coming in C++-20) so you will need to write one yourself. Fortunately, its pretty easy. 

The net effect now is that history is probably full of the same state, as the top of the history stack will always be a different object to the current table, even though they may be in the same state.

> history.push_back(godleyIcon->table);
> }
> historyPtr=history.size();
> }
> 
> void GodleyTableWindow::undo(int changes)
> {
> if (historyPtr==history.size())
> pushHistory();
> historyPtr-=changes;
> if (historyPtr > 0 && historyPtr <= history.size())
> {
> auto& d=history[historyPtr-1];
> if (!&d) return; // should not happen

Again - because we're not storing pointers, this statement makes no sense whatsoever.

> godleyIcon->table=d;
> requestRedraw();
> }
> }`
> 
> Then, I have also tried to pass a "currTable" in pushHistory(), which also required editing minskyTCLObj.h on line 53:
> 
> ` if (!t || (!t->is_const && (!t->is_setterGetter || argc>1))) if (auto gtw=t->memberPtrCasted<GodleyTableWindow>()) { auto g=t->memberPtrCasted<GodleyTable>(); gtw->pushHistory(*g); }`
> 
> Then, in godleyTableWindow.h:
> 
> `void pushHistory(const GodleyTable& currTable);`
> 
> and in godleyTableWindow.cc:
> 
> ` void GodleyTableWindow::pushHistory(const GodleyTable& currTable)
> {
> while (history.size()>maxHistory) history.pop_front();
> if (history.empty() || (&history.back()!=&currTable)) {
> history.push_back(currTable);
> }
> historyPtr=history.size();
> }
> 
> void GodleyTableWindow::undo(int changes)
> {
> if (historyPtr==history.size())
> pushHistory(godleyIcon->table);
> historyPtr-=changes;
> if (historyPtr > 0 && historyPtr <= history.size())
> {
> auto& d=history[historyPtr-1];
> if (!&d) return; // should not happen
> godleyIcon->table=d;
> requestRedraw();
> }
> }`
> 
> This produces a segmentation fault that I think has to do with the fact that the objects in gtw->pushHistory(*g) in minskyTCLObj.h and pushHistory(godleyIcon->table) in godleyWindowTable::undo() are not the same? From backtrace in gdb:
> 
> (gdb) frame 7
> #7 0x00005555556a962a in minsky::member_entry_hook<char const**> (argc=2, argv=0x55555656be90) at gui-tk/minskyTCLObj.h:53
> *_53 gtw->pushHistory(_g);__
> (gdb) frame 6
> #6 0x0000555555b3d9a4 in minsky::GodleyTableWindow::pushHistory (this=0x555556b75400, currTable=...) at model/godleyTableWindow.cc:831
> **831 history.push_back(currTable);**

model/godleyTableWindow.h
highperformancecoder: Yes - that is why you have to refer to the type as GodleyAssetClass::AssetClass.

model/godleyTableWindow.cc
wdednam: Thank you, I really didn't understand why it wasn't working. 

I will add a bool operator==(const GodleyTable& x) const; in godleyTableWindow, test it and tomorrow morning commit the changes.

PR #130, opened 2020-01-22
Proposed fix for crash on latex export

wdednam:
This change is [Reviewable](https://reviewable.io/reviews/highperformancecoder/minsky/130)

### DIFF
--- a/engine/node_latex.cc
+++ b/engine/node_latex.cc
@@ -206,7 +206,7 @@ namespace MathDAG
           {
             checkArg(1,i);
             if (i>0) o<<"\\times ";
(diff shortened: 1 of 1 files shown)


engine/node_latex.cc
```diff
@@ -206,7 +206,7 @@ namespace MathDAG
           {
             checkArg(1,i);
             if (i>0) o<<"\\times ";
-            ParenIf p(o, i>0 && arguments[0][i]->BODMASlevel()>BODMASlevel());
+            ParenIf p(o, i>0 && arguments[1][i]->BODMASlevel()>BODMASlevel());
```
highperformancecoder: Good catch!

PR #131, opened 2020-01-24
Adjust bounding box for data import operation

wdednam:
Looks like I forgot to open a PR for this one.

---
This change is [Reviewable](https://reviewable.io/reviews/highperformancecoder/minsky/131)


model/operation.cc
```diff
@@ -697,6 +697,7 @@ namespace minsky
     description = "\\verb/"+
```
highperformancecoder: Yes - except that description can be altered from the UI. I think we'll need to change description into an Accessor much like we did with rotate, and attach a bb.update() to the setter.

PR #132, opened 2020-01-25
Adjust bounding box of data operation whenever its description is changed

wdednam:
This change is [Reviewable](https://reviewable.io/reviews/highperformancecoder/minsky/132)

### DIFF
--- a/model/operation.cc
+++ b/model/operation.cc
@@ -686,8 +686,8 @@ namespace minsky
    
   string DataOp::description_(const std::string& x)
   {
-    bb.update(*this); // adjust icon bounding box - see ticket #704
     m_description=x;
+    bb.update(*this); // adjust icon bounding box - see ticket #1121
     return m_description;
   }    
 
@@ -709,7 +709,6 @@ namespace minsky
     description("\\verb/"+
       ((p!=string::npos)? fileName.substr(p+1): fileName) + "/");
     //initXVector();
-    bb.update(*this); // adjust bounding box for data import operation - see ticket #1121 
   }
 
   void DataOp::initRandom(double xmin, double xmax, unsigned numSamples)
--- a/model/operation.cc
+++ b/model/operation.cc
@@ -148,7 +148,7 @@ namespace minsky
 
     auto t=type();
     // call the iconDraw method if data description is empty
-    if (t==OperationType::data && dynamic_cast<const DataOp&>(*this).description.empty())
+    if (t==OperationType::data && dynamic_cast<const DataOp&>(*this).description().empty())
       t=OperationType::numOps;
 
     switch (t)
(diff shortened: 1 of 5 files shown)


model/operation.cc
```diff
@@ -678,6 +678,18 @@ namespace minsky
...
+  
+  string DataOp::description_() const
+  {
+	return m_description;  
+  }
+   
+  string DataOp::description_(const std::string& x)
+  {
+    bb.update(*this); // adjust icon bounding box - see ticket #704
+    m_description=x;
```
highperformancecoder: These two statements are done in the wrong order

model/operation.cc
```diff
@@ -694,8 +706,8 @@ namespace minsky
     size_t p=fileName.rfind('/');
     // '/' is guaranteed not to be in fileName, so we can use that as
     // a delimiter
-    description = "\\verb/"+
-      ((p!=string::npos)? fileName.substr(p+1): fileName) + "/";
+    description("\\verb/"+
+      ((p!=string::npos)? fileName.substr(p+1): fileName) + "/");
     //initXVector();
     bb.update(*this); // adjust bounding box for data import operation - see ticket #1121
```
highperformancecoder: Can be removed now

model/operation.h
```diff
@@ -206,8 +206,27 @@ namespace minsky
   class DataOp: public ItemT<DataOp, Operation<minsky::OperationType::data>>
   {
     CLASSDESC_ACCESS(DataOp);
+    friend struct SchemaHelper;
+    string m_description;
   public:
-    string description;
+    DataOp() {description("");}
```
highperformancecoder: The default default constructor would be fine too (bb is updated next time it is used).

model/operation.h
```diff
@@ -206,8 +206,27 @@ namespace minsky
...
+    // ensure that copies create a new data operation
+    DataOp(const DataOp& x): 
+      OperationBase(x){group.reset(); description(x.description());}
```
highperformancecoder: Whilst probably harmless, I was wondering why you added group.reset() here. group.reset is already done in ItemT::clone. None of the other item copy constructors have this, except for IntOp. For consistency, I suggest removing the group.reset here and in IntOp. You can log this as a low priority backlog ticket, however, if your'e concerned we might not understand the implications of that change.

model/operation.cc
wdednam: Thanks for catching this.

model/operation.cc
wdednam: I have done so, thanks.

model/operation.h
wdednam: I have also made this the default constructor.

model/operation.h
wdednam: Sorry, I missed this and will commit the changes after the current Travis build has finished. It didn't occur to me that the group.reset() in ItemT:clone() would work for these objects too. I sometimes forget that these and other objects fall under the item object umbrella.

PR #133, opened 2020-01-25
Bugfix 1111 item only

wdednam:
This change is [Reviewable](https://reviewable.io/reviews/highperformancecoder/minsky/133)

### DIFF
--- a/test/00/groupResize.sh
+++ b/test/00/groupResize.sh
@@ -46,8 +46,9 @@ proc afterMinskyStarted {} {uplevel #0 {
  canvas.mouseDown [expr \$x+0.5*\$w] [expr \$y+0.5*\$h]
  canvas.mouseUp [expr \$x+\$w]  [expr \$y+\$h]
 
- assert "abs([expr 1.5*\$w]-[minsky.canvas.item.width])<5"
- assert "abs([expr 1.5*\$h]-[minsky.canvas.item.height])<5"
+ # Resize handles larger on groups, see ticket 1023, so needed to adjust the factors below
+ assert "abs([expr 1.3*\$w]-[minsky.canvas.item.width])<5"
+ assert "abs([expr 1.4*\$h]-[minsky.canvas.item.height])<5"
  tcl_exit
 }}
 EOF
--- a/gui-tk/wiring.tcl
+++ b/gui-tk/wiring.tcl
@@ -339,6 +339,7 @@ proc addConstantOrVariable {} {
     set varInput(Name) ""
     set varInput(Value) ""
     set varInput(Units) ""
+    set varInput(Rotation) [minsky.canvas.defaultRotation]
     set varInput(Type) $varType
     set "varInput(Short description)" ""
     set "varInput(Detailed description)" ""
--- a/minskyVersion.h
+++ b/minskyVersion.h
@@ -1,1 +1,1 @@
-#define MINSKY_VERSION "2.17.7"
+#define MINSKY_VERSION "2.17.8"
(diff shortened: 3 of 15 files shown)


test/00/groupResize.sh
```diff
@@ -46,8 +46,9 @@ proc afterMinskyStarted {} {uplevel #0 {
  canvas.mouseDown [expr \$x+0.5*\$w] [expr \$y+0.5*\$h]
  canvas.mouseUp [expr \$x+\$w]  [expr \$y+\$h]
 
- assert "abs([expr 1.5*\$w]-[minsky.canvas.item.width])<5"
- assert "abs([expr 1.5*\$h]-[minsky.canvas.item.height])<5"
+ # Resize handles larger on groups, see ticket 1023, so needed to adjust the factors below
```
highperformancecoder: Yes - this. I'm still unhappy with the GroupItem placement of resize handles.

highperformancecoder:
Given a) that I think there is still an issue with Group rendering, and b) that I think I have a better alternative to ecolab::Accessor that we can use in other cases as they arise, I might just hold off merging this until we've looked at those issues.

highperformancecoder:
Sorry - didn't mean to close this just yet.

test/00/groupResize.sh
wdednam: I realised after I requested the merge that I still hadn't fixed up the test, sorry. Once you have implemented the alternative to the Accessor, I will fix it if still required. I think the problem is caused by a combination of the larger resize handles and the left and right margins of the IO variables, which I still have to figure out how to call in the test TCL code, but haven't yet managed to.

highperformancecoder:
On Sun, Jan 26, 2020 at 02:55:43AM -0800, Wynand wrote:
> 
> 
> In test/00/groupResize.sh:
> 
> > @@ -46,8 +46,9 @@ proc afterMinskyStarted {} {uplevel #0 {
>   canvas.mouseDown [expr \$x+0.5*\$w] [expr \$y+0.5*\$h]
>   canvas.mouseUp [expr \$x+\$w]  [expr \$y+\$h]
> 
> - assert "abs([expr 1.5*\$w]-[minsky.canvas.item.width])<5"
> - assert "abs([expr 1.5*\$h]-[minsky.canvas.item.height])<5"
> + # Resize handles larger on groups, see ticket 1023, so needed to adjust the factors below
> 
> I realised after I requested the merge that I still hadn't fixed up the test,
> sorry. Once you have implemented the alternative to the Accessor, I will fix it
> if still required. I think the problem is caused by a combination of the larger
> resize handles and the left and right margins of the IO variables, which I
> still have to figure out how to call in the test TCL code, but haven't yet
> managed to.

My theory is that a bb.update() needs to be called after io vars have
been changed, or similar. Also, the way width/height is done in Groups
differs from the other items, and maybe needs to be brought into line
with the other Item types.

-- 

                      http://www.hpcoders.com.au

highperformancecoder:
I discovered the solution for the group resize issue. In schema2, bb
needs to be updated on loading the iconWidth/Height parameters. I
changed the name of these for good measure. Also fixed up a few other
places it makes sense to update the bounding box.


-- 

                      http://www.hpcoders.com.au

wdednam:
Awesome, thank you. The change you made in schema2.cc works like a charm. The test groupResize.sh works in its original form (no need to change the 1.5 coefficients to 1.3 and 1.4 in the asserts or to include the values of the margins either).

highperformancecoder:
This has now been obsoleted by other developments

PR #134, opened 2020-01-27
Enter/return key press in Godley table produces exit from highlighted…

wdednam:
… cell

---
This change is [Reviewable](https://reviewable.io/reviews/highperformancecoder/minsky/134)


highperformancecoder:
One way or other, I think this covers all cases. I wonder why it wasn't working before - the key marked "enter" on a PC keyboard actually generates "return", which we were handling.

PR #136, opened 2020-01-31
Fix for duplicate column bug on swapping of columns, accidental or ot…

wdednam:
…herwise

---
This change is [Reviewable](https://reviewable.io/reviews/highperformancecoder/minsky/136)

### DIFF
--- a/gui-tk/godley.tcl
+++ b/gui-tk/godley.tcl
@@ -43,8 +43,8 @@ proc openGodley {id} {
         bind .$id.table <<contextMenu>> "godleyContext $id %x %y %X %Y"
         bind .$id.table <KeyPress> "$id.keyPress %N [encoding convertto utf-8 %A]"
         # Pressing the return key deselects a cell. For ticket 1122.        
-        bind .$id.table <Key-Return> "enter $id"
-        bind .$id.table <Key-KP_Enter> "enter $id"        
+        bind .$id.table <Key-Return> "enter $id %x %y"
+        bind .$id.table <Key-KP_Enter> "enter $id %x %y"        
 
         global meta meta_menu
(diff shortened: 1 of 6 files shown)


gui-tk/godley.tcl
```diff
@@ -43,8 +43,8 @@ proc openGodley {id} {
         bind .$id.table <<contextMenu>> "godleyContext $id %x %y %X %Y"
         bind .$id.table <KeyPress> "$id.keyPress %N [encoding convertto utf-8 %A]"
         # Pressing the return key deselects a cell. For ticket 1122.        
-        bind .$id.table <Key-Return> "enter $id"
-        bind .$id.table <Key-KP_Enter> "enter $id"        
+        bind .$id.table <Key-Return> "enter $id %x %y"
+        bind .$id.table <Key-KP_Enter> "enter $id %x %y"
```
highperformancecoder: I'll accept this change for now, however since the mouse may be anywhere on the screen, the behaviour will probably still be surprising. Better, I would think, is to highlight the cell underneath the current selected cell, like what Excel does.

highperformancecoder: I think change has now been superseded by developments...

highperformancecoder:
I removed the enter proc stuff, as that was made obsolete by further development.

gui-tk/godley.tcl
wdednam: I found that with this change, the Godley tables were further stabilised and the stock variables would no longer be renamed to "1" on pressing enter. The appearance of these "1"s really baffled me. So, I was happy when I made this change and the effect was still that of the cell appearing deselected. Of course, as you point out, having enter work this way will surprise users.

PR #137, opened 2020-01-31
Group icon was not resizing properly on zoom in and out. Committing a…

wdednam:
… proposed fix.

---
This change is [Reviewable](https://reviewable.io/reviews/highperformancecoder/minsky/137)

### DIFF
--- a/model/group.cc
+++ b/model/group.cc
@@ -781,7 +781,7 @@ namespace minsky
       // display I/O region in grey
       drawIORegion(cairo);
 
-      cairo_translate(cairo, -0.5*iconWidth+leftMargin, -0.5*iconHeight);
+      cairo_translate(cairo, -0.5*width+leftMargin, -0.5*height);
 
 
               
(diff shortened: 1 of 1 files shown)


highperformancecoder:
This is effectively reverting a change I made. I don't understand how it even compiles, as now the "width" and "height" members are functions, not member variables.

The change I made width => iconWidth etc was out of an abundance of caution - the actual bug I was chasing was due to something else.

wdednam:
It is a pity the commit comparison doesn't show a few lines of code around the changes:

In particular, the following line (775 in group.cc) which I guess you added to improve readability of Group::draw(), clarifies the situation:

unsigned width=z*this->iconWidth, height=z*this->iconHeight;

The changes I made were essentially just a correction of a few typos, not a regression to the previous width() or height() members. Sorry I didn't include a reference to the above line in my commit.

highperformancecoder:
In that case I apologise. I was mislead by the way the code review tool presented the difference.

wdednam:
Thanks, but there's no need to :). I will resolve to explain my commits a little better from now on.

 I didn't want raise another ticket at sourceforge for this because the tickets are already ballooning and I thought this was a quick fix we could do under the radar. I'm still struggling a bit with ticket 1072, to access the colWidgets vectors in TCL. I need to add getters because colWidgets and rowWidgets  are protected members of the godleyTableWindow class.

highperformancecoder:
On Sat, Feb 01, 2020 at 02:17:35AM -0800, Wynand wrote:
> Thanks, but there's no need to :). I will resolve to explain my commits a
> little better from now on.

Usually, there is no need to. I can always ask questions!

Cheers


-- 

                      http://www.hpcoders.com.au

PR #138, opened 2020-02-02
warn user when a stock variable column is going to change asset class…

wdednam:
… on mouse click and/or drag

---
This change is [Reviewable](https://reviewable.io/reviews/highperformancecoder/minsky/138)

### DIFF
--- a/gui-tk/godley.tcl
+++ b/gui-tk/godley.tcl
@@ -158,11 +158,11 @@ proc mouseDown {id x y X Y} {
 
 # warn user when a stock variable column is going to be moved to a different asset class on pressing a column button widget. For ticket 1072.
 proc moveAssetClass {id x y X Y} {	
-   set c [$id.colXZoomed $x]
-   set oldAssetClass [$id.godleyIcon.table.getAssetClass $c]    	
-   set moveVar [$id.godleyIcon.table.getCell 0 $c]
    switch [$id.clickTypeZoomed $x $y] {    
    colWidget {
+      set c [$id.colXZoomed $x]   	
+      set moveVar [$id.godleyIcon.table.getCell 0 $c]	   
+	  set oldAssetClass [$id.godleyIcon.table.getAssetClass $c]  
 	  set colWId [$id.getColWidget $x] 
 		  if {$colWId==3} {     
              set targetAssetClass [$id.godleyIcon.table.getAssetClass [expr $c+1]]
--- a/gui-tk/godley.tcl
+++ b/gui-tk/godley.tcl
@@ -158,11 +158,11 @@ proc mouseDown {id x y X Y} {
 
 # warn user when a stock variable column is going to be moved to a different asset class on pressing a column button widget. For ticket 1072.
 proc moveAssetClass {id x y X Y} {	
+   set c [$id.colXZoomed $x]
+   set oldAssetClass [$id.godleyIcon.table.getAssetClass $c]    	
+   set moveVar [$id.godleyIcon.table.getCell 0 $c]
    switch [$id.clickTypeZoomed $x $y] {    
    colWidget {
-      set c [$id.colXZoomed $x]   	
-      set moveVar [$id.godleyIcon.table.getCell 0 $c]	   
-	  set oldAssetClass [$id.godleyIcon.table.getAssetClass $c]  
 	  set colWId [$id.getColWidget $x] 
 		  if {$colWId==3} {     
              set targetAssetClass [$id.godleyIcon.table.getAssetClass [expr $c+1]]
--- a/gui-tk/godley.tcl
+++ b/gui-tk/godley.tcl
@@ -34,7 +34,7 @@ proc openGodley {id} {
         bind .$id.table <Destroy> "$id.delete"
 
         bind .$id.table <ButtonPress-1> "moveAssetClass $id %x %y %X %Y"
-        bind .$id.table <ButtonRelease-1> "defaultCursor .$id.table; swapAssetClasses $id %x %y"
+        bind .$id.table <ButtonRelease-1> "defaultCursor .$id.table; swapAssetClass $id %x %y"
         bind .$id.table <B1-Motion> "motionCursor .$id.table; $id.mouseMoveB1 %x %y"
         bind .$id.table <Motion> "$id.mouseMove %x %y"
         bind .$id.table <Leave> "$id.mouseMove -1 -1; $id.update"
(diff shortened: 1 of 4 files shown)


model/godleyTable.h
```diff
@@ -106,6 +106,14 @@ namespace minsky
        sets if assetClass present, otherwise gets
     */
     string assetClass(ecolab::TCL_args args);
+    
+    // returns the asset class of column col. For ticket 1072.
+    AssetClass getAssetClass(unsigned col) const {
```
highperformancecoder: This does duplicate the method above. (assetClass).

gui-tk/godley.tcl
```diff
@@ -156,6 +156,90 @@ proc mouseDown {id x y X Y} {
     }
 }
 
+# warn user when a stock variable column is going to be moved to a different asset class on pressing a column button widget. For ticket 1072.
```
highperformancecoder: There is too much "business logic" wrapped up in TCL code for my liking. Given we're planning on replacing TCL code with Javascript code, it would be better to find a way of managing this process with as much business logic done in C++ as possible.

I can think of two approaches: 

1) a C++ method that detects whether the column move changes asset classes and if so returns back to TCL to post the dialog. If not then the column change method is called directly. In the case of the TCL dialog approving the column change, then call into the same C++ method from TCL to do it.

2) The alternative method is for C++ to call back into TCL to post the dialog. In order to do this, but not infect the main C++ code with too much TCL stuff is to use a virtual function override, much like the clipboard copy/paste operations.

model/godleyTable.h
wdednam: Sorry, I was thrown off by its type (string) and the appearance of its argument. So I went back to getAssetClass(). But I've now replaced getAssetClass() by assetClass(). Thanks.

gui-tk/godley.tcl
wdednam: I understand and should have had the Javascript code in mind when I did this... 

For the second option, would the code that I need to add to minskyTCL.cc be platform independent? In the case of the clipboard, I understand that this is not the case because different platforms handle clipboard differently, but I'm not sure in this case.

highperformancecoder:
On Sun, Feb 02, 2020 at 02:31:36PM -0800, Wynand wrote:
> 
> I understand and should have had the Javascript code in mind when I did this...
> 
> For the second option, would the code that I need to add to minskyTCL.cc be
> platform independent? In the case of the clipboard, I understand that this is
> not the case because different platforms handle clipboard differently, but I'm
> not sure in this case.

It should be platform independent in this case. To call back into TCL
from C++, instantiate a tclcmd object, and pass some TCL commands,
terminated by a newline, eg:

tclcmd() << "puts \"hello\"\n";

To separate TCL commands, use a semicolon, eg:

tclcmd cmd;
cmd << "puts \"hello\";"
cmd << "puts \"" << my_name << "\"\n";

Obviously wrap up complex TCL code into a proc to make it
easier. Return values from TCL can be found in cmd.result.

However, that being said, in the RESTService model, there are no
callbacks into Javascript (at least I can't think of a way of doing
it), so maybe the first option is better.


Cheers


-- 

                      http://www.hpcoders.com.au

wdednam:
Thanks for the hints on using TCL in C++ code. I will try the first approach and do my best to finish it today so that I can devote some time to striking off tickets before our workshop.

gui-tk/godley.tcl
wdednam: I took the first approach in the end. I tried to combine both swapAssetClass and moveAssetClass methods into one but hit a snag in TCL when I tried to use the variable buttonPressed (with values 0 or 1) to switch between (mouseUp and mouseDown). Obviously, mouseDown or buttonPressed=1 always gets activated first, so I spilt the method back into the two I originally had. 

As an aside, I tried to fix the right arrow/tab (navigateRight()) and left arrow/back tab traversal (navigateLeft()) methods to traverse all editable cells in a closed loop, but only succeeded in the first case. Back tab will only traverse cells from the last one up to the first in the initial conditions row and the left arrow will go all the way to the first editable cell but then stay in the first row upon continued navigateLeft(), i.e., it does not start from the last cell in the table.

model/godleyTableWindow.cc
```diff
@@ -798,6 +798,55 @@ namespace minsky
...
+	string tmpStr="";
+	if (clickType(x,y)==colWidget) {
+	    unsigned visibleCol=c-scrollColStart+1;
+        if (c<colWidgets.size() && visibleCol < colLeftMargin.size()) {
+		    auto& moveVar=godleyIcon->table.cell(0,c);
+		    auto oldAssetClass=classdesc::enumKey<GodleyAssetClass::AssetClass>(godleyIcon->table._assetClass(c));			
+            if (colWidgets[c].button(x-colLeftMargin[visibleCol])==3) {
+				auto targetAssetClass=classdesc::enumKey<GodleyAssetClass::AssetClass>(godleyIcon->table._assetClass(c+1));
+				if (targetAssetClass!=oldAssetClass && !moveVar.empty())
+					tmpStr="This will convert "+moveVar+" from "+oldAssetClass+" to "+targetAssetClass+". Are you sure?";
```
highperformancecoder: These two lines of code have effectively been repeated 4 times, and are a candidate for extraction into a function.

model/godleyTableWindow.cc
```diff
@@ -798,6 +798,55 @@ namespace minsky
...
+  {
+	x/=zoomFactor;
+	y/=zoomFactor;
+	unsigned c=colX(x);	
+	string tmpStr="";
+	if (clickType(x,y)==colWidget) {
+	    unsigned visibleCol=c-scrollColStart+1;
+        if (c<colWidgets.size() && visibleCol < colLeftMargin.size()) {
+		    auto& moveVar=godleyIcon->table.cell(0,c);
+		    auto oldAssetClass=classdesc::enumKey<GodleyAssetClass::AssetClass>(godleyIcon->table._assetClass(c));
```
highperformancecoder: I would leave the values as enums until the final construction of the string return value.

model/godleyTable.h
highperformancecoder: getAssetClass now dead code? Of would you prefer to rename it to assetClass, and delete the existing setter/getter on line 108? I believe the setter is not used anywhere in any TCL file (maybe it was once).

wdednam: Sorry, I forgot to commit this change. I removed it from my develop folder but not the one from which I do commits, so it didn't show up for the commit as the other changes did. I will remove it.

model/godleyTableWindow.cc
wdednam: Definitely agree. I wanted to do it "in place" but with the changing values of oldAssetClass and targetAssetClass that wouldn't work. But a function is the obvious alternative. Thanks for the suggestion :)

model/godleyTableWindow.cc
wdednam: Excellent suggestion! I will make that change today.

PR #141, opened 2020-02-05
Very minor fix: added zoomfactor in some parts of draw() in ravelWrap.cc

wdednam:
I noticed that ravel objects were not behaving like other canvas items upon zoom in or out and added the zoomfactor to a few places in ravelWrap::draw().

---
This change is [Reviewable](https://reviewable.io/reviews/highperformancecoder/minsky/141)

### DIFF
--- a/model/ravelWrap.cc
+++ b/model/ravelWrap.cc
@@ -314,19 +314,18 @@ namespace minsky
 
   void Ravel::draw(cairo_t* cairo) const
   {
-    double r=ravelDefaultRadius, z=zoomFactor();
+    double  z=zoomFactor(), r=ravelDefaultRadius*z;
     if (ravel) r=1.1*z*ravel_radius(ravel);
-    double s=r*z;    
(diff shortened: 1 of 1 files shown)


model/ravelWrap.cc
```diff
@@ -316,16 +316,17 @@ namespace minsky
   {
     double r=ravelDefaultRadius, z=zoomFactor();
     if (ravel) r=1.1*z*ravel_radius(ravel);
-    ports[0]->moveTo(x()+1.1*r, y());
-    ports[1]->moveTo(x()-1.1*r, y());
+    double s=r*z;
```
highperformancecoder: Why do you need to square the zoomFactor? It doesn't make sense.

highperformancecoder: Maybe the first r (ravelDefaultRadius) needs to be multiplied by the zoomFactor

wdednam: Thanks, your suggestion did the trick and is much simpler than what I proposed. :)

PR #143, opened 2020-02-06
Rename integrate operation now checks for attached input too

wdednam:
This change is [Reviewable](https://reviewable.io/reviews/highperformancecoder/minsky/143)

### DIFF
--- a/model/operation.cc
+++ b/model/operation.cc
@@ -574,7 +574,7 @@ namespace minsky
 
     string vid=VariableValue::valueId(group.lock(),desc);
     auto i=minsky().variableValues.find(vid);      
-    if (i!=minsky().variableValues.end())
+    if (i!=minsky().variableValues.end() &&  && minsky().definingVar(vid))
       {
         if (i->second.type()!=VariableType::integral) 
           try
--- a/model/operation.cc
+++ b/model/operation.cc
@@ -574,7 +574,7 @@ namespace minsky
 
     string vid=VariableValue::valueId(group.lock(),desc);
     auto i=minsky().variableValues.find(vid);      
-    if (i!=minsky().variableValues.end() && minsky().definingVar(vid)) 
+    if (i!=minsky().variableValues.end()) 
       {
         if (i->second.type()!=VariableType::integral) 
           try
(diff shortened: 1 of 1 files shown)


wdednam:
Test t0001a.sh fails on and off for some weird reason. I've just run it again on my machine and it has passed.
test_t0001a_passes

wdednam:
When it fails, it gives the error:

attempt to move a group into itselfunittests: model/canvas.cc:606: void minsky::Canvas::renameAllInstances(std::__cxx11::string): Assertion `model->numItems()==numItems' failed.
./t0001a.sh: line 16: 14154 Aborted                 (core dumped) $here/test/unittests

model/operation.cc
```diff
@@ -574,7 +574,7 @@ namespace minsky
 
     string vid=VariableValue::valueId(group.lock(),desc);
     auto i=minsky().variableValues.find(vid);      
-    if (i!=minsky().variableValues.end())
+    if (i!=minsky().variableValues.end() && minsky().definingVar(vid)) // Also check that integral has input. for ticket 1068.
```
highperformancecoder: Actually, the test for definingVar() should be put on the else clause on line 588. Think about it. In all other cases, we just need to check that the variable can be converted to an integral type, it need not be defined.

wdednam: Thank you, I've been struggling with this one. One of the unit tests keeps failing, the one on line ~1013 of testModel.cc.

testModel.cc:1013:1: error: Failure in description: intop2->description()!=intop->description()

 It looks like testing for definingVar() permits the possibility of more than one intVar with the same name... So, I tried to add a method in minsky.cc that checks for more than one intVar too, to call at the same time as definingVar() in the else statement in description() in operation.cc:

  bool Minsky::existingIntVar(const string& valueId) const 
  {
    bool r=false;
    model->recursiveDo
      (&Group::items,
       [&](Items&,Items::const_iterator i) {
         if (auto v=dynamic_cast<IntOp*>(i->get()))
          if (v->variableCast()->valueId()==valueId)
              return true;
        return false;
      });
    return r;    
    
   }  
    
but I haven't got it to work yet... I also tried adding coupled() to the else if condition in description() on line 588, e.g.,

        else
            if (coupled() && minsky().definingVar(vid))
                desc=minsky().variableValues.newName(vid);
        else 
            desc=vid;  

 but when it works, I can create two intVars with the same name on the canvas and the test fails... I'll keep trying :)

PR #144, opened 2020-02-10
Catch unknown exception when an incomplete model is saved as mky file…

wdednam:
… and opened in Minsky

---
This change is [Reviewable](https://reviewable.io/reviews/highperformancecoder/minsky/144)

### DIFF
--- a/engine/variableValue.cc
+++ b/engine/variableValue.cc
@@ -200,10 +200,14 @@ namespace minsky
 
   void VariableValue::reset(const VariableValues& v)
   {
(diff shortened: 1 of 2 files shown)


highperformancecoder:
I don't understand the purpose of this catch. There is already a catch in Minsky::load that absorbs exceptions.

wdednam:
To be honest, me either. But the problem is somewhere in here, because it goes away with this try..catch. I didn't have time to explore it further yesterday. I will look at it again while in transit today or tomorrow.

PR #158, opened 2020-02-25
several minor fixes of plot legend

wdednam:
This change is [Reviewable](https://reviewable.io/reviews/highperformancecoder/minsky/158)

### DIFF
--- a/model/plotWidget.cc
+++ b/model/plotWidget.cc
@@ -196,7 +196,8 @@ namespace minsky
 
     cairo_translate(cairo, portSpace, yoffs);
     cairo_set_line_width(cairo,1);
-    double gw=w-2*portSpace, gh=h-portSpace;
+    double gw=w-2*portSpace, gh=h-portSpace;;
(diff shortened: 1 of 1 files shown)


model/plotWidget.cc
```diff
@@ -196,7 +196,8 @@ namespace minsky
 
     cairo_translate(cairo, portSpace, yoffs);
     cairo_set_line_width(cairo,1);
-    double gw=w-2*portSpace, gh=h-portSpace;
+    double gw=w-2*portSpace, gh=h-portSpace;;
```
highperformancecoder: extraneous ;

model/plotWidget.cc
```diff
@@ -228,13 +231,12 @@ namespace minsky
             drawTriangle(cairo,x,y+0.5*height+arrowLength,{0,0,0,1},M_PI/2);
 
             cairo_rectangle(cairo,x-0.5*width,y-0.5*height,width,height);
-            cairo_stroke(cairo);
```
highperformancecoder: no longer drawin border around legend? cairo_rectangle not needed then...

PR #159, opened 2020-03-05
Make multiple equity columns a preference

wdednam:
This change is [Reviewable](https://reviewable.io/reviews/highperformancecoder/minsky/159)

### DIFF
--- a/gui-tk/godley.tcl
+++ b/gui-tk/godley.tcl
@@ -106,6 +106,7 @@ proc openGodley {id} {
         menu .$id.menubar.options
         .$id.menubar.options add checkbutton -label "Show Values" -variable preferences(godleyDisplay) -command setGodleyDisplay
         .$id.menubar.options add checkbutton -label "DR/CR style" -variable preferences(godleyDisplayStyle) -onvalue DRCR -offvalue sign -command setGodleyDisplay
+        .$id.menubar.options add checkbutton -label "Enable multiple equities" -variable preferences(multipleEquities) -command setGodleyDisplay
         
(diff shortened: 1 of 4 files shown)


highperformancecoder:
OK - I see you've implemented this as a global per user option, rather than as an actual attribute of the Godley table (which is persisted with the model file). Lets leave this as it is for now for Bentham, but I think we should reopen the discussion about whether this should be changed at some future point. ISTM it makes more sense to make the attribute per godley table, because it might look a little strange seeing multiple equity columns and no way of moving or deleting them.

wdednam:
Yes. I wasn't sure where in the schema files I had to make the change to make it persist with a model file. I looked, admittedly briefly, for other items that may have such attributes but could not see them. I'm wiliing to change it right away if there is an example of another item with such an attribute you could suggest I look at. Thanks.

highperformancecoder:
On Thu, Mar 05, 2020 at 01:14:50PM -0800, Wynand wrote:
> Yes. I wasn't sure where in the schema files I had to make the change to make
> it persist with a model file. I looked, admittedly briefly, for other items
> that may have such attributes but could not see them. I'm wiliing to change it
> right away if there is an example of another item with such an attribute you
> could suggest I look at. Thanks.

You would need to add another optional field in the schema3::Item
class, and modify the schema3::Item(minsky::GodleyIcon constructor),
as well as the code populating GodleyIcons in schema3.cc. About 3
lines of code needed in total. You may as well look at it to see how
to do it, but we should discuss if its the correct way of doing this
feature...


-- 

                      http://www.hpcoders.com.au

PR #160, opened 2020-03-05
Expose tensor ops to ravel. Not sure all are useful for ravel or even accessible

wdednam:
This change is [Reviewable](https://reviewable.io/reviews/highperformancecoder/minsky/160)


highperformancecoder:
I'm afraid you're barking up the wrong tree here. There are already tensor ops for the things that you are defining RavelTensor classes for.

What you need to do is to write code that parses the RavelState object, and generates a chain of tensorOps that implements that transformation. See the latest code on eg MultiWireBinOp to see how to compose TensorOps into an expression chain.

wdednam:
Thank you. I wasn't sure at all and a little confused by the "chain of tensor ops" we had previously discussed in Sydney too. I thought they had to be similar to the general tensor ops? Just for clarity, do I need to use a vector of tensor pointers as in the SwitchTensor, but unlike in that case. not perform operations element-wise on them?

wdednam:
> An example of a chain of tensorOps is the result of
TensorsFromPort::tensorsFromPort(). This produces a tree of tensorOps,
connected by their argument shared_ptrs.

I missed this part entirely, so thank you for suggesting I look at it.

> The things you will need: reduction ops and slice ops (they don't
currently exist, the current index operation is slightly like what you
want, though

At least I was partially on the right track at one point: I copied the bit of code that updates cachedResult in GeneralTensorOp<OperationType::index>: public civita::CachedTensorOp into the the RaveTensor class in computeTensor(), to overload it as you suggested. So, I'll come back to this approach now. Thanks:)

engine/minskyTensorOps.cc
```diff
@@ -162,6 +162,10 @@ namespace minsky
 //  };
 
   template <OperationType::Type op> struct GeneralTensorOp;
+
+
+  template <OperationType::Type ravel> struct RavelTensor;
```
highperformancecoder: Why does this need to be a template?

engine/minskyTensorOps.cc
```diff
@@ -191,6 +195,7 @@ namespace minsky
     registerOps<TensorBinOp, OperationType::log, OperationType::copy>(*this);
     registerOps<MinskyTensorOp, OperationType::copy, OperationType::sum>(*this);
     registerOps<GeneralTensorOp, OperationType::sum, OperationType::numOps>(*this);
+    registerOps<RavelTensor, OperationType::sum, OperationType::numOps>(*this);
```
highperformancecoder: It is a single type, so you should do registerType<RavelTensor>(OperationType::ravel);

Note the presence of the static variable tensorOpFactor on the TimeOp is a bit of a mistake (harmless one), but I've removed in a subsequent commit.

engine/minskyTensorOps.cc
```diff
@@ -446,24 +451,207 @@ namespace minsky
...
       const_cast<Ravel&>(ravel).loadDataCubeFromVariable(*arg);
       ravel.loadDataFromSlice(cachedResult);
       m_timestamp = Timestamp::clock::now();
-    }
+    }    
     
   public:
-    RavelTensor(const Ravel& ravel): ravel(ravel) {}
-    void setArgument(const TensorPtr& a,const std::string& d,double) override {arg=a;}
+    //RavelTensor(const Ravel& ravel): ravel(ravel) {}
```
highperformancecoder: Dead code...

engine/minskyTensorOps.cc
wdednam: This is still the part of the first commit. I'll commit the latest code with ravel as a registered single op in tensorOpFactory, but which still needs the actual operations. I'm going to spend the morning finishing working through Ravel's source code to get a complete picture of how the operations are invoked (with sorting and other ancillary operations performed on the data before the actual chain of operations in RawData::reduce()). I envision that I will start adding the operations in minskyTensorOps.cc this evening.

wdednam: The latest version also does not access the actual ravel state. I only managed to achieve that once today: by adding a RavelTensorOp struct to minskyTensorOps.cc, one deriving from EvalOp<OperationType::ravel>, to a sligthly modified version of yours and just before the RavelTensor class. In that case, it worked like the previous dense tensor version only for simple tensors such as iota(9). It would print out the value as the caliper was moved along the single handle and the sum of 36 (for 0, 1, 2, 3, 4, 5, 6, 7, 8) when the handle was collapsed. It would of course not work for a higher dimensional case (the dimensions would be all wrong). My idea was to have the dense tensor version working like before at least. But I now understand why we can't have this version fall back to 2.15.5 for even dense tensors...

engine/minskyTensorOps.cc
```diff
@@ -446,21 +449,45 @@ namespace minsky
...
     TensorPtr arg;
-    void computeTensor() const override
+    void computeTensor() const override  
     {
-      const_cast<Ravel&>(ravel).loadDataCubeFromVariable(*arg);
-      ravel.loadDataFromSlice(cachedResult);
-      m_timestamp = Timestamp::clock::now();
-    }
-    
+	   if (auto r=dynamic_cast<Ravel*>(arg.get())) {
```
highperformancecoder: This sounds like a confusion. arg points to an ITensor object, Ravel is an Item. Two different object heirarchies. Maybe your thinking of subclassing Ravel from ITensor, but I'd rather you didn't. I don't think it would make sense.

wdednam: Confusion cleared up. Registering RavelTensor as an Op in tensorFactoryOp like TimeOp caused problems with the Ravel object declared as a reference in the RavelTensor class. I got the uninitialised reference error.  So I made "const Ravel& ravel" just plain "const Ravel ravel". 

Of course, then, I didn't get the updated Ravel state...I tried making it a shared_ptr and also a unique_ptr, but Minsky would crash even just for simple tensors like iota(9)... Finally, it was also necessary to add the default constructor of RavelTensor in order to compile the code.

So, I gave up and in the end moved the Item-to-Ravel cast to the top of the if conditional in TensorOpFactory::create() and just updated cachedResult in setArguments in RavelTensor, which has got me the closest to version 2.15.5 functionality I've seen.

Now, cachedResult will have to be updated by a chain of operations in RavelTensor::ComputeTensor() if it is to work properly. 

But I'm not yet fully convinced I can't take advantage of what's already in ravelWrap. Are any of the other members in ravelWrap available to me, e.g., the slicing and sorting?

In this regard, I'll soon be committing code that uses the C interface in ravelWrap.cc in the hope that cachedResult is updated correctly. To use it with the tensor library wouldn't I just have to use the correct hypercube indices? At present, it compiles but then crashes when I try to test it on iota(9) or similar...

wdednam: It took duplicating within minskyTensorOps.cc a lot of code from ravelWrap.cc to figure out that the last line in Ravel::loadDataFromSlice() was returning an empty result in the original code. Now, if I understand correctly, I have to replace the "ravel_" functions in Ravel::loadDataCubeFromVariable() and Ravel::loadDataFromSlice() with versions that are consistent with the Civita library. This is what I've started doing today.

wdednam:
Okay, I've managed to detect and update the timestamp() when the Ravel handle currently selected is rotated. (i used the 2-D dense data CSV file: CPI_ByMonth.csv, attached as a zip file, for testing purposes).

Based on my understanding of the xvectors, the tick labels on the x-axis in a plot, or the horizontal and vertical labels in a sheet (for a 2-D Ravel), should update, or swap, when I rotate the handles. I calculate the correct number of labels when I pick the handles, and also feed them into the xvectors. So, I don't understand why the lables don't change when I rotate the handles, since I update the cachedResult hypercube with the new xvectors?


[CPI_ByMonth.zip](https://github.com/highperformancecoder/minsky/files/4344921/CPI_ByMonth.zip)

engine/minskyTensorOps.cc
```diff
@@ -447,27 +446,124 @@ namespace minsky
     }
   };
 
-  class RavelTensor: public CachedTensorOp
+  class RavelTensor: public civita::DimensionedArgCachedOp
```
highperformancecoder: RavelTensor is not a DimensionedArgCachedOp. A DimensionedArgCachedOp is one that applies to a specific dimension, as specified by the operation "axis" attribute. Eg reduction along a specific dimension (or axis).

model/ravelWrap.cc
```diff
@@ -494,7 +494,8 @@ namespace minsky
         else
           throw error(ravel_lastErr());
       }
-    v.hypercube({}); // ensure scalar data space allocated
+      // This line causes an empty result from ravel output port
+    //v.hypercube({}); // ensure scalar data space allocated
```
highperformancecoder: This is still needed, but only if the hypercube setting on line 488 hasn't taken place. Perhaps best handled by adding a return statement on line 493.

engine/minskyTensorOps.cc
```diff
@@ -491,13 +587,13 @@ namespace minsky
...
-        r->setArguments(tfp.tensorsFromPorts(it.ports));
-        return r;
-      }
+      } // Original code
+    //else if (auto ravel=dynamic_cast<const Ravel*>(&it))
+    //  {
+    //    auto r=make_shared<RavelTensor>(*ravel);
+    //    r->setArguments(tfp.tensorsFromPorts(it.ports));
+    //    return r;
+    //  }
```
highperformancecoder: Clean up dead code.

engine/minskyTensorOps.cc
wdednam: Fixed this. Thanks for the explanation.

model/ravelWrap.cc
wdednam: Okay, made these changes. Unfortunately did them in separate commits. Sorry.

engine/minskyTensorOps.cc
wdednam: Done.

wdednam:
Sorry, I'm about to commit a slight modification of this code, which checks for the sparsity of the data and throws an error. Just need to finish building and running the tests locally.

wdednam:
I've also removed the additional member I added to ravelWrap yesterday.

PR #161, opened 2020-03-07
Preference is now 'Enable multiple equity columns' and updated schema…

wdednam:
…3 to make the attribute per Godley table. Also added popup message to option menu in table to suggest user close and reopen table to bring feature into effect"

---
This change is [Reviewable](https://reviewable.io/reviews/highperformancecoder/minsky/161)

### DIFF
--- a/gui-tk/godley.tcl
+++ b/gui-tk/godley.tcl
@@ -101,10 +101,10 @@ proc openGodley {id} {
         .$id.menubar.view add command -label "Zoom out" -command "zoomOut $id" -accelerator $meta_menu--
         .$id.menubar.view add command -label "Reset zoom" -command "$id.zoomFactor 1; $id.requestRedraw"
         
-        menu .$id.menubar.options -postcommand "toggleEquityColumns $id" 
+        menu .$id.menubar.options  
+        ###-postcommand "toggleEquityColumns $id" 
         .$id.menubar.options add checkbutton -label "Show Values" -variable preferences(godleyDisplay) -command setGodleyDisplay
         .$id.menubar.options add checkbutton -label "DR/CR style" -variable preferences(godleyDisplayStyle) -onvalue DRCR -offvalue sign -command setGodleyDisplay
-        .$id.menubar.options add checkbutton -label "Enable multiple equity columns" -variable equityColumns -command "toggleEquityColumns $id"
         
         .$id.menubar add cascade -label File -menu .$id.menubar.file -underline 0
         .$id.menubar add cascade -label Edit -menu .$id.menubar.edit -underline 0
@@ -115,9 +115,6 @@ proc openGodley {id} {
         global preferences
         $id.displayValues $preferences(godleyDisplay)
         $id.displayStyle $preferences(godleyDisplayStyle)
-        global equityColumns
-        $id.godleyIcon.table.multipleEquities $equityColumns
-
         
     }
     wm deiconify .$id
@@ -304,23 +301,9 @@ proc setGodleyDisplay {} {
     }
     foreach c [info commands godleyWindow*.displayStyle] {
         $c $preferences(godleyDisplayStyle)
-    }
+    }    
     redrawAllGodleyTables
 }
-
-# sets each individual Godley table mutliple equity column preference
-proc toggleEquityColumns id {
-	global equityColumns
-    foreach c [info commands id.godleyIcon.table.multipleEquities] {
-        $c $equityColumns
-        if {$c==0} {
-	     .$id.menubar.edit entryconfigure end -state disabled
-        } else {
-	     .$id.menubar.edit entryconfigure end -state normal
-        }          
-    }        
-    redrawAllGodleyTables
-} 
 
 proc exportGodley {id} {
     global workDir type
--- a/gui-tk/godley.tcl
+++ b/gui-tk/godley.tcl
@@ -308,13 +308,12 @@ proc setGodleyDisplay {} {
     redrawAllGodleyTables
 }
 
+# sets each individual Godley table mutliple equity column preference
 proc toggleEquityColumns id {
 	global preferences
-	tk_messageBox -message "Close and reopen the Godley table to bring this feature into effect" -type ok -parent .$id.table
-    foreach c [info commands godleyIcon.table.multipleEquities] {
-        $c $preferences(multipleEquities)
-    }   
-    redrawAllGodleyTables
+	tk_messageBox -message "Close and reopen the Godley table to enable/disable this feature" -type ok -parent .$id.table
+    set $id.multipleEquities [info commands id.godleyIcon.table.multipleEquities]
+    $id.requestRedraw
 } 
 
 proc exportGodley {id} {
--- a/gui-tk/godley.tcl
+++ b/gui-tk/godley.tcl
@@ -106,6 +106,7 @@ proc openGodley {id} {
         menu .$id.menubar.options
         .$id.menubar.options add checkbutton -label "Show Values" -variable preferences(godleyDisplay) -command setGodleyDisplay
         .$id.menubar.options add checkbutton -label "DR/CR style" -variable preferences(godleyDisplayStyle) -onvalue DRCR -offvalue sign -command setGodleyDisplay
+        .$id.menubar.options add checkbutton -label "Enable multiple equities" -variable preferences(multipleEquities) -command setGodleyDisplay
         
         .$id.menubar add cascade -label File -menu .$id.menubar.file -underline 0
         .$id.menubar add cascade -label Edit -menu .$id.menubar.edit -underline 0
@@ -116,6 +117,7 @@ proc openGodley {id} {
         global preferences
         $id.displayValues $preferences(godleyDisplay)
         $id.displayStyle $preferences(godleyDisplayStyle)
+        $id.godleyIcon.table.multipleEquities $preferences(multipleEquities)
 
     }
     wm deiconify .$id
@@ -303,8 +305,12 @@ proc setGodleyDisplay {} {
     foreach c [info commands godleyWindow*.displayStyle] {
         $c $preferences(godleyDisplayStyle)
     }
+    foreach c [info commands godleyIcon.table.multipleEquities] {
+        $c $preferences(multipleEquities)
+    }    
     redrawAllGodleyTables
 }
+	
 
 proc exportGodley {id} {
     global workDir type
--- a/gui-tk/godley.tcl
+++ b/gui-tk/godley.tcl
@@ -103,10 +103,10 @@ proc openGodley {id} {
         
         
         
-        menu .$id.menubar.options
+        menu .$id.menubar.options  
         .$id.menubar.options add checkbutton -label "Show Values" -variable preferences(godleyDisplay) -command setGodleyDisplay
         .$id.menubar.options add checkbutton -label "DR/CR style" -variable preferences(godleyDisplayStyle) -onvalue DRCR -offvalue sign -command setGodleyDisplay
-        .$id.menubar.options add checkbutton -label "Enable multiple equities" -variable preferences(multipleEquities) -command setGodleyDisplay
+        .$id.menubar.options add checkbutton -label "Enable multiple equity columns" -variable preferences(multipleEquities) -command "toggleEquityColumns $id"
         
         .$id.menubar add cascade -label File -menu .$id.menubar.file -underline 0
         .$id.menubar add cascade -label Edit -menu .$id.menubar.edit -underline 0
@@ -305,12 +305,17 @@ proc setGodleyDisplay {} {
     foreach c [info commands godleyWindow*.displayStyle] {
         $c $preferences(godleyDisplayStyle)
     }
-    foreach c [info commands godleyIcon.table.multipleEquities] {
-        $c $preferences(multipleEquities)
-    }    
     redrawAllGodleyTables
 }
-	
+
+proc toggleEquityColumns id {
+	global preferences
+	tk_messageBox -message "Close and reopen the Godley table to bring this feature into effect" -type ok -parent .$id.table
+    foreach c [info commands godleyIcon.table.multipleEquities] {
+        $c $preferences(multipleEquities)
+    }   
+    redrawAllGodleyTables
+} 
 
 proc exportGodley {id} {
     global workDir type
--- a/gui-tk/godley.tcl
+++ b/gui-tk/godley.tcl
@@ -102,7 +102,6 @@ proc openGodley {id} {
         .$id.menubar.view add command -label "Reset zoom" -command "$id.zoomFactor 1; $id.requestRedraw"
         
         menu .$id.menubar.options  
-        ###-postcommand "toggleEquityColumns $id" 
         .$id.menubar.options add checkbutton -label "Show Values" -variable preferences(godleyDisplay) -command setGodleyDisplay
         .$id.menubar.options add checkbutton -label "DR/CR style" -variable preferences(godleyDisplayStyle) -onvalue DRCR -offvalue sign -command setGodleyDisplay
         
--- a/gui-tk/godley.tcl
+++ b/gui-tk/godley.tcl
@@ -104,6 +104,7 @@ proc openGodley {id} {
         menu .$id.menubar.options  
         .$id.menubar.options add checkbutton -label "Show Values" -variable preferences(godleyDisplay) -command setGodleyDisplay
         .$id.menubar.options add checkbutton -label "DR/CR style" -variable preferences(godleyDisplayStyle) -onvalue DRCR -offvalue sign -command setGodleyDisplay
+        .$id.menubar.options add checkbutton -label "Enable multiple equity columns" -variable preferences(multipleEquities) -command setGodleyDisplay
         
         .$id.menubar add cascade -label File -menu .$id.menubar.file -underline 0
         .$id.menubar add cascade -label Edit -menu .$id.menubar.edit -underline 0
@@ -300,7 +301,10 @@ proc setGodleyDisplay {} {
     }
     foreach c [info commands godleyWindow*.displayStyle] {
         $c $preferences(godleyDisplayStyle)
-    }    
+    }
+    foreach c [info commands multipleEquities] {
+        $c $preferences(multipleEquities)
+    }          
     redrawAllGodleyTables
 }
 
--- a/gui-tk/godley.tcl
+++ b/gui-tk/godley.tcl
@@ -101,12 +101,10 @@ proc openGodley {id} {
         .$id.menubar.view add command -label "Zoom out" -command "zoomOut $id" -accelerator $meta_menu--
         .$id.menubar.view add command -label "Reset zoom" -command "$id.zoomFactor 1; $id.requestRedraw"
         
-        
-        
-        menu .$id.menubar.options  
+        menu .$id.menubar.options -postcommand "toggleEquityColumns $id" 
         .$id.menubar.options add checkbutton -label "Show Values" -variable preferences(godleyDisplay) -command setGodleyDisplay
         .$id.menubar.options add checkbutton -label "DR/CR style" -variable preferences(godleyDisplayStyle) -onvalue DRCR -offvalue sign -command setGodleyDisplay
-        .$id.menubar.options add checkbutton -label "Enable multiple equity columns" -variable preferences(multipleEquities) -command "toggleEquityColumns $id"
+        .$id.menubar.options add checkbutton -label "Enable multiple equity columns" -variable equityColumns -command "toggleEquityColumns $id"
         
         .$id.menubar add cascade -label File -menu .$id.menubar.file -underline 0
         .$id.menubar add cascade -label Edit -menu .$id.menubar.edit -underline 0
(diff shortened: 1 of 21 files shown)


gui-tk/godley.tcl
```diff
@@ -103,10 +103,10 @@ proc openGodley {id} {
         
         
         
-        menu .$id.menubar.options
+        menu .$id.menubar.options  
         .$id.menubar.options add checkbutton -label "Show Values" -variable preferences(godleyDisplay) -command setGodleyDisplay
         .$id.menubar.options add checkbutton -label "DR/CR style" -variable preferences(godleyDisplayStyle) -onvalue DRCR -offvalue sign -command setGodleyDisplay
-        .$id.menubar.options add checkbutton -label "Enable multiple equities" -variable preferences(multipleEquities) -command setGodleyDisplay
+        .$id.menubar.options add checkbutton -label "Enable multiple equity columns" -variable preferences(multipleEquities) -command "toggleEquityColumns $id"
```
highperformancecoder: OK - with this change, you use a different variable rather than one in the preferences array. Or maybe even do away with the TCL variable.

You can use the menu -postcommand option to set the state of the check button variable every time this menu is posted.

gui-tk/godley.tcl
```diff
@@ -305,12 +305,16 @@ proc setGodleyDisplay {} {
...
-        $c $preferences(multipleEquities)
-    }    
     redrawAllGodleyTables
 }
-	
+
+# sets each individual Godley table mutliple equity column preference
+proc toggleEquityColumns id {
+	global preferences
+	tk_messageBox -message "Close and reopen the Godley table to enable/disable this feature" -type ok -parent .$id.table
```
highperformancecoder: This shouldn't be necessary if the change of state isn reflected in appearance.

gui-tk/godley.tcl
```diff
@@ -305,12 +305,16 @@ proc setGodleyDisplay {} {
...
+    set $id.multipleEquities [info commands id.godleyIcon.table.multipleEquities]
```
highperformancecoder: I don't understand what you're trying to do here. This sets a TCL variable with the name of the multipleEquities accessor function.

Surely what you really mean here is

$id.table.multipleEquities $preferences(multipleEquities)

gui-tk/godley.tcl
wdednam: Thanks. I've done away with the variable and can change the check button's appearance with the -postcommand option in a similar way we do with paste selection:

`
menu .$id.menubar.options -postcommand "toggleEquityColumns $id"
        .$id.menubar.options add checkbutton -label "Show Values" -variable preferences(godleyDisplay) -command setGodleyDisplay
        .$id.menubar.options add checkbutton -label "DR/CR style" -variable preferences(godleyDisplayStyle) -onvalue DRCR -offvalue sign -command setGodleyDisplay
        .$id.menubar.options add checkbutton -label "Enable multiple equity columns" -command "toggleEquityColumns $id"
`

I know I have to use the -textvariable option somewhere in toggleEquityColumns to change the value of $id.multipleEquities too, but haven't succeeded yet, so I have this in the proc now:

`
# sets each individual Godley table mutliple equity column preference
proc toggleEquityColumns id {
    set v $id.multipleEquities
    if {$v==0} {
	 .$id.menubar.edit entryconfigure end -state disabled
	  $id.requestRedraw
    } else {
	 .$id.menubar.edit entryconfigure end -state normal
	  $id.requestRedraw
    }      
}
`

wdednam: Sorry ,the code comment block does not work very well, so I'm repasting the first bit of code:

        menu .$id.menubar.options -postcommand "toggleEquityColumns $id" 
        .$id.menubar.options add checkbutton -label "Show Values" -variable preferences(godleyDisplay) -command setGodleyDisplay
        .$id.menubar.options add checkbutton -label "DR/CR style" -variable preferences(godleyDisplayStyle) -onvalue DRCR -offvalue sign -command setGodleyDisplay
        .$id.menubar.options add checkbutton -label "Enable multiple equity columns" -command "toggleEquityColumns $id"

gui-tk/godley.tcl
wdednam: I've removed it, thanks :)

gui-tk/godley.tcl
wdednam: Oops, that was a mistake, thanks for catching it.

gui-tk/godley.tcl
highperformancecoder: I couldn't find any obvious way of accessing the state of the checkbutton from the menu items interface. Hence, you probably do need the -variable option. But call it something else, not preferences(...).

wdednam: Thanks, I've fixed this but noticed that when there are more than one equity columns only the button widgets on the last column disappears, and not all of them. The number of columns should also automatically reduce to one when this preference is activated, so I'm busy fixing this too...

wdednam: Extra equity columns are now deleted when the preference is reverted from multiple to single equity column mode. I've taken a little longer to commit the code because I ran the test suite to make sure that all tests pass before the commit. I think everything works correctly now.

wdednam:
I don't understand why adding multipleEquities as a global preference causes the model to become "edited" and hence to always ask if the model is to be saved on exit, which leads the test suite to hang... I've tried several different approaches but I always end up with this outcome.

gui-tk/minsky.tcl
```diff
@@ -223,6 +224,14 @@ if {$preferences(focusFollowsMouse)} {
 }
 proc setCursor {cur} {. configure -cursor $cur; update idletasks}
 
+# Toggle single or multiple equity column mode
+multipleEquities $preferences(multipleEquities)
+if $preferences(multipleEquities) {
+   set multipleEquities 0
+} else {
+   set multipleEquities 1
```
highperformancecoder: Why do you need a separate TCL variable here?

gui-tk/minsky.tcl
```diff
@@ -1406,7 +1415,14 @@ proc setPreferenceParms {} {
...
        }
-    }  
+    }
+    # Toggle single or multiple equity column mode
+    multipleEquities $preferences(multipleEquities)
+    if $preferences(multipleEquities) {
+       set multipleEquities 0
+    } else {
+       set multipleEquities 1
+    }
```
highperformancecoder: Ditto.

model/godleyTableWindow.cc
```diff
@@ -201,7 +201,14 @@ namespace minsky
           {
             CairoSave cs(surface->cairo());
             cairo_move_to(surface->cairo(), x, columnButtonsOffset);
-            colWidgets[col].draw(surface->cairo());
+            // Prevent the existence of more than one equity column in single equity column mode
+            if (godleyIcon->table._assetClass(col)==GodleyAssetClass::equity && col!=godleyIcon->table.cols()-1 && !cminsky().multipleEquities) {
+               godleyIcon->table.deleteCol(col);
```
highperformancecoder: I don't think it is a good idea to abruptly delete columns. Rather, I think that the presence of more than one equity column should be treated identically to the case where multipleEquities is true (ie the old way).

gui-tk/minsky.tcl
wdednam: I don't know if I'm setting the variable correctly, but using the following does nothing at all, the preference remains single equity column:

if {$preferences(multipleEquities)} {
   set $preferences(multipleEquities) 0
} else {
   set $preferences(multipleEquities) 1
}

model/godleyTableWindow.cc
wdednam: Okay, I understand now. I should deal with the number of columns at column draw time and not button widget draw time. Thanks :)

wdednam:
I still have to fix the problem in the minsky.tcl script, but I don't understand why the preference remains unchanged when I use only the one TCL variable preferences(multipleEquities).

model/godleyTable.cc
```diff
@@ -169,6 +169,13 @@ GodleyTable::AssetClass GodleyTable::_assetClass
   return _assetClass(col);
 }
 
+bool GodleyTable::singleEquity() {
+	int count=0;
+	for (size_t c=1;c<cols();++c)
+	  if (m_assetClass[c]==GodleyAssetClass::equity) count++;
+	if (count > 1) return false;
```
highperformancecoder: You can do this more efficiently by just checking the rightmost two columns.

gui-tk/godley.tcl
```diff
@@ -101,12 +101,9 @@ proc openGodley {id} {
         .$id.menubar.view add command -label "Zoom out" -command "zoomOut $id" -accelerator $meta_menu--
         .$id.menubar.view add command -label "Reset zoom" -command "$id.zoomFactor 1; $id.requestRedraw"
         
-        
-        
-        menu .$id.menubar.options
+        menu .$id.menubar.options  
         .$id.menubar.options add checkbutton -label "Show Values" -variable preferences(godleyDisplay) -command setGodleyDisplay
         .$id.menubar.options add checkbutton -label "DR/CR style" -variable preferences(godleyDisplayStyle) -onvalue DRCR -offvalue sign -command setGodleyDisplay
-        .$id.menubar.options add checkbutton -label "Enable multiple equities" -variable preferences(multipleEquities) -command setGodleyDisplay
```
highperformancecoder: I still think we want to have this in the GodleyTable preferences menu.

gui-tk/minsky.tcl
```diff
@@ -223,6 +224,14 @@ if {$preferences(focusFollowsMouse)} {
 }
 proc setCursor {cur} {. configure -cursor $cur; update idletasks}
 
+# Toggle single or multiple equity column mode
+multipleEquities $preferences(multipleEquities)
```
highperformancecoder: This is the only place your setting Minsky::multipleEquities. You should set it within setPreferenceParms, which is called whenever the preference panel is closed.

gui-tk/minsky.tcl
highperformancecoder: You are setting a TCL variable correctly, but not using it in any way. You already have a TCL variable anyway, in the form of preferences(multipleEquities), so a multipleEquities TCL variable is redundant.

You need to set the Minksy::multipleEquities attribute for anything to happen.

gui-tk/minsky.tcl
wdednam: I only set this in setPreferenceParms now and have reinstated the code in godley.tcl, without actually setting the value there. I just apply it to all Godleys in proc setGodleyDisplay. The feature looks great now. The button widgets appear instantaneously when I deactivate and reactivate the preference. I've also fixed tests t0027a.sh, t0030a.sh, t0033a.sh and t0036a.sh to not save the model upon afterMinskyStarted.

gui-tk/godley.tcl
```diff
@@ -305,12 +302,11 @@ proc setGodleyDisplay {} {
     foreach c [info commands godleyWindow*.displayStyle] {
         $c $preferences(godleyDisplayStyle)
     }
-    foreach c [info commands godleyIcon.table.multipleEquities] {
+    foreach c [info commands multipleEquities] {
```
highperformancecoder: I think we can assume that the multipleEquities command always exists, so the foreach loop is not necessary

model/godleyTable.cc
```diff
@@ -169,6 +169,11 @@ GodleyTable::AssetClass GodleyTable::_assetClass
   return _assetClass(col);
 }
 
+bool GodleyTable::singleEquity() {
+	GodleyTable::AssetClass cls=GodleyAssetClass::equity;
+	if (m_assetClass[cols()-1]==cls && m_assetClass[cols()-2]==cls) return false;
```
highperformancecoder: Is OK, but I would write this as:
return m_assetClass[cols()-2]!=GodleyAssetClass::equity

If the second to right most column is equity, then so will the right most column.
The only other thing to consider is if the table could ever have fewer than 3 columns. I'd probably add that as an assertion.

model/godleyTable.h
```diff
@@ -72,21 +70,23 @@ namespace minsky
...
+		 doubleEntryCompliant==other.doubleEntryCompliant && title==other.title);
     }    
 
     /// class of each column (used in DE compliant mode)
     const vector<AssetClass>& _assetClass() const {return m_assetClass;}
     AssetClass _assetClass(size_t col) const;
     AssetClass _assetClass(size_t col, AssetClass cls);
+    
+    // Check whether more than one equity column is present irrespective of single or multiple equity column mode. For feature 111.
+    bool singleEquity();
```
highperformancecoder: This method should be const

model/godleyTableWindow.cc
```diff
@@ -1116,7 +1116,7 @@ namespace {
   {	    
     CairoSave cs(cairo);
     int idx=0;
-    if (!godleyIcon.table.multipleEquities || rowCol==row) {  // no column widgets on equity column in single equity column mode
+    if ((rowCol==row || !cminsky().multipleEquities) || (!cminsky().multipleEquities && godleyIcon.table.singleEquity())) {  // no column widgets on equity column in single equity column mode
```
highperformancecoder: The way this conditional is written, the final singleEquity() call is not done due to boolean shortcuts.
I suspect what is needed is:
rowCol==row || !cminsky().multipleEquities || godleyIcon.table.singleEquity()

test/00/t0027a.sh
```diff
@@ -35,6 +35,7 @@ cat >input.tcl <<EOF
 source assert.tcl
 proc afterMinskyStarted {} {
 uplevel #0 {
+minsky.save no
```
highperformancecoder: Are you planning on leaving this line in?

test/00/t0030a.sh
```diff
@@ -34,6 +34,7 @@ trap "fail" 1 2 3 15
 cat >input.tcl <<EOF
 source $here/test/assert.tcl
 proc afterMinskyStarted {} {
+  minsky.save no
```
highperformancecoder: ditto

test/00/t0033a.sh
```diff
@@ -34,6 +34,7 @@ trap "fail" 1 2 3 15
 cat >input.tcl <<EOF
 source $here/test/assert.tcl
 proc afterMinskyStarted {} {uplevel #0 {
+ minsky.save no
```
highperformancecoder: ditto

test/00/t0036a.sh
```diff
@@ -35,6 +35,7 @@ trap "fail" 1 2 3 15
 cat >input.tcl <<EOF
 source $here/test/assert.tcl
 proc afterMinskyStarted {} {uplevel #0 {
+  minsky.save no
```
highperformancecoder: ditto

test/00/t0027a.sh
wdednam: No, I hope that making the boolean variable multipleEquities const in minsky.h, as you have suggested, solves the save at exit warning that has caused the test suite to hang.

highperformancecoder:
On Thu, Mar 12, 2020 at 12:47:48AM -0700, Wynand wrote:
> 
> 
> In test/00/t0027a.sh:
> 
> > @@ -35,6 +35,7 @@ cat >input.tcl <<EOF
>  source assert.tcl
>  proc afterMinskyStarted {} {
>  uplevel #0 {
> +minsky.save no
> 
> No, I hope that making the boolean variable multipleEquities const in minsky.h,
> as you have suggested, solves the save at exit warning that has caused the test
> suite to hang.

minsky.save saves the model to a file called "no". Not what you want,
I would think.
 
There is a "minsky.resetEdited" method that does what want, I suspect.


-- 

                      http://www.hpcoders.com.au

wdednam:
Okay, still hitting a wall with this, sorry...

When I make the boolean multipleEquities const, then I can't set its value:

model/minsky.cc: In member function ‘void minsky::Minsky::load(const string&)’:
model/minsky.cc:939:35: error: use of deleted function ‘minsky::Minsky& minsky::Minsky::operator=(minsky::Minsky&&)’
     *this=schema3::Minsky(saveFile);
                                   ^
In file included from model/minsky.cc:21:0:
model/minsky.h:143:9: note: ‘minsky::Minsky& minsky::Minsky::operator=(minsky::Minsky&&)’ is implicitly deleted because the default definition would be ill-formed:
   class Minsky: public ValueVector, public Exclude<MinskyExclude>, public RungeKutta
         ^~~~~~
model/minsky.h:143:9: error: non-static const member ‘const bool minsky::Minsky::multipleEquities’, can’t use default assignment operator
/home/wd/minsky/ecolab/include/Makefile:616: recipe for target 'minsky.o' failed

Okay, I then have to add a getter as well as a setter:

    bool multipleEquities() const {return false;}
    bool multipleEquities(const bool me) {return me;}  

This compiles but then TCL can't find the setter?

invalid command name "multipleEquities"
invalid command name "multipleEquities"
    while executing
"tclunknown multipleEquities 0"
    ("eval" body line 1)
    invoked from within
"eval tclunknown $procname $args"
    (procedure "::unknown" line 18)
    invoked from within
"multipleEquities $preferences(multipleEquities)"
    (procedure "setPreferenceParms" line 8)
    invoked from within
"setPreferenceParms"

I don't understand, I thought that in TCL the difference between setter and getter wa just the presence or absence of a (list of) argument(s)? So, if I overload multipleEquities in minsky.h, then TCL should be smart enough to tell the difference between them, shouldn't it?

> minsky.save saves the model to a file called "no". Not what you want,
I would think.

There is a "minsky.resetEdited" method that does what want, I suspect.

Thank you, I will look there.

model/godleyTable.h
wdednam: Will I have to add a TCL accessor getter as in the case of assetClass in godleyTable.h?

wdednam: It looks like this will require a class_desc modification, so not sure about doing so. I tried:

    bool m_multipleEquities=false;

public:
    // Allow multiple equity columns.
    const bool& _multipleEquities() const {return m_multipleEquities;}
    bool _multipleEquities(const bool& me)
    {
		m_multipleEquities=me;
		return m_multipleEquities;
	}
    
    /**
       TCL accessor method 
       @param me - boolean value of multipleEquities member
       @return current value of multipleEquities
       sets if value is passed, otherwise gets
    */
    bool multipleEquities(ecolab::TCL_args args); 

But got an undefined reference at the linking stage:

minskyTCL.o: In function `void classdesc_access::access_isa<minsky::Minsky>::operator()<minsky::Minsky>(std::set<std::type_info const*, std::less<std::type_info const*>, std::allocator<std::type_info const*> >&, std::__cxx11::basic_string<char, std::char_traits<char>, std::allocator<char> > const&, minsky::Minsky&)':
/home/wd/minsky/./minsky.cd:1059: undefined reference to `minsky::Minsky::multipleEquities(ecolab::TCL_args)'
minskyTCL.o: In function `void classdesc_access::access_TCL_obj<minsky::Minsky>::operator()<minsky::Minsky>(classdesc::TCL_obj_t&, std::__cxx11::basic_string<char, std::char_traits<char>, std::allocator<char> > const&, minsky::Minsky&)':
/home/wd/minsky/./minsky.cd:784: undefined reference to `minsky::Minsky::multipleEquities(ecolab::TCL_args)'
minskyTCL.o: In function `void classdesc_access::access_pack<minsky::Minsky>::operator()<minsky::Minsky>(classdesc::pack_t&, std::__cxx11::basic_string<char, std::char_traits<char>, std::allocator<char> > const&, minsky::Minsky&)':
/home/wd/minsky/./minsky.cd:186: undefined reference to `minsky::Minsky::multipleEquities(ecolab::TCL_args)'
minskyTCL.o: In function `void classdesc_access::access_unpack<minsky::Minsky>::operator()<minsky::Minsky>(classdesc::pack_t&, std::__cxx11::basic_string<char, std::char_traits<char>, std::allocator<char> > const&, minsky::Minsky&)':
/home/wd/minsky/./minsky.cd:435: undefined reference to `minsky::Minsky::multipleEquities(ecolab::TCL_args)'
collect2: error: ld returned 1 exit status
Makefile:162: recipe for target 'gui-tk/minsky' failed
make: *** [gui-tk/minsky] Error 1

So, I'm going back to non const.

highperformancecoder:
On Thu, Mar 12, 2020 at 01:21:52AM -0700, Wynand wrote:
> 
> Okay, I then have to add a getter as well as a setter:
> 
> bool multipleEquities() const {return false;}
> bool multipleEquities(const bool me) {return me;}

Oh no, no need to do that. Take a look at the code in
minskyTCLObj.h. You'll see how to avoid setting the edit flag when
setting the multipleEquities attribute.

Cheers


-- 

                      http://www.hpcoders.com.au

model/godleyTableWindow.cc
wdednam: I've tried several combinations, and what I already have comes the closest to making the feature perfect. What I haven't considered is to separate the rows and cols, which may be the way to go...

model/godleyTable.h
wdednam: Could not achieve compilation with a const method. I think my lack of knowledge of and/or experience in C++ is the cause here...

model/godleyTable.cc
wdednam: Done, thanks :)

gui-tk/godley.tcl
wdednam: Removing this caused the feature to stop working, so I reinstated it. Again, I may not have enough knowldedge of TCL to do this properly.

wdednam:
Thank you, modifying minskyTCLObj.h worked :)

PR #163, opened 2020-03-17
Update 2.15.5

wdednam:
This change is [Reviewable](https://reviewable.io/reviews/highperformancecoder/minsky/163)


wdednam:
Updated CSVParser.cc to its latest version, excluding the sparse data code.

highperformancecoder:
I really don't know what you're trying to do here. 2.15 branch contains some minor customisation fixes to allow the code at that time to build on all systems.

wdednam:
I was trying to update version 2.15.5 of Minsky to have the latest CSVParser.cc code, excluidng the sparse tensor stuff.

highperformancecoder:
I think best bet would be to use the existing RavelTensor implementation, with its dense implementation:
    void computeTensor() const override
    {
      const_cast<Ravel&>(ravel).loadDataCubeFromVariable(*arg);
      ravel.loadDataFromSlice(cachedResult);
      m_timestamp = Timestamp::clock::now();
    }
The problem with why it wasn't working has to do with that issue in create, which you've now diagnosed and fixed (though not yet merged into top of tree).
We should also detect and throw on sparse data coming in, but at least we should be able to get dense datasets working just as well as in 2.15.5.

That way, we're moving forward, and its only a small set of changes to review and merge.

wdednam:
Okay, I've reintroduced the original computeTensor() implementation. I'm really just confused as to why the dimensions and their labels (i.e., xvectors) are not passed correctly to the output tensor. The data set I've been using to test the code changes has a sparsityRatio of 0.44, so it is dense. I think we might have to pass the itemPtr to TensorEval::eval() too in order to evaluate the rhs tensorPtr like it was done in the EvalOp class?

PR #164, opened 2020-03-19
code checks for sparsity of data on ravel output and throws an error …

wdednam:
…if it exceeds 0.5

---
This change is [Reviewable](https://reviewable.io/reviews/highperformancecoder/minsky/164)

### DIFF
--- a/engine/minskyTensorOps.cc
+++ b/engine/minskyTensorOps.cc
@@ -453,9 +453,8 @@ namespace minsky
    
     void computeTensor() const override  
     {
-	  double sparsityRatio =  static_cast<double>(1.0-static_cast<double>(arg->size())/arg->hypercube().numElements());  
 	  
-	  if (sparsityRatio <= 0.5) {	
+	  if (arg->index().empty()) {	
 	      const_cast<Ravel&>(ravel).loadDataCubeFromVariable(*arg);	
 	      ravel.loadDataFromSlice(cachedResult);	      
       } else throw runtime_error("ravel evaluation of sparse data not available");
--- a/engine/minskyTensorOps.cc
+++ b/engine/minskyTensorOps.cc
@@ -453,8 +453,13 @@ namespace minsky
    
     void computeTensor() const override  
     {
-      const_cast<Ravel&>(ravel).loadDataCubeFromVariable(*arg); 	 
-      ravel.loadDataFromSlice(cachedResult);
+	  double sparsityRatio =  static_cast<double>(1.0-static_cast<double>(arg->size())/arg->hypercube().numElements());  
+	  
+	  if (sparsityRatio <= 0.5) {	
+	      const_cast<Ravel&>(ravel).loadDataCubeFromVariable(*arg);	
+	      ravel.loadDataFromSlice(cachedResult);	      
+      } else throw runtime_error("ravel evaluation of sparse data not available");
+      
       m_timestamp = Timestamp::clock::now();
     }    
     CLASSDESC_ACCESS(Ravel);
--- a/engine/minskyTensorOps.cc
+++ b/engine/minskyTensorOps.cc
@@ -453,18 +453,9 @@ namespace minsky
    
     void computeTensor() const override  
     {
-	  
-	  double sparsityRatio =  static_cast<double>(1.0-static_cast<double>(arg->size())/arg->hypercube().numElements());  
-	  
-	  if (arg->index().empty()) {	
-	      const_cast<Ravel&>(ravel).loadDataCubeFromVariable(*arg);	
-	      ravel.loadDataFromSlice(cachedResult);	      
-      } else if (sparsityRatio <= 0.5) {
-		  const_cast<Ravel&>(ravel).loadDataCubeFromVariable(*arg);	
-		  ravel.loadDataFromSlice(cachedResult);	      
-	  } else throw runtime_error("ravel evaluation of sparse data not available");	
-      
-      m_timestamp = Timestamp::clock::now();
+      const_cast<Ravel&>(ravel).loadDataCubeFromVariable(*arg);	
+      ravel.loadDataFromSlice(cachedResult);	            
+      m_timestamp = Timestamp::clock::now(); 
     }    
     CLASSDESC_ACCESS(Ravel);
   public:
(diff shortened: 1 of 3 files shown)


engine/minskyTensorOps.cc
```diff
@@ -453,8 +453,13 @@ namespace minsky
    
     void computeTensor() const override  
     {
-      const_cast<Ravel&>(ravel).loadDataCubeFromVariable(*arg); 	 
-      ravel.loadDataFromSlice(cachedResult);
+	  double sparsityRatio =  static_cast<double>(1.0-static_cast<double>(arg->size())/arg->hypercube().numElements());  
+	  
+	  if (sparsityRatio <= 0.5) {
```
highperformancecoder: Kind of OK, but a better test is arg->index().empty().

wdednam: I've been testing the latest version and it works just like in version 2.15.5. Thanks for fixing it. The closest I could get to picking up the hypercube was not being set correctly was to try to set it to the output tensorVal's hypercube, cachedResult. But that caused havoc and only worked for 1-D examples like iota. So, I'm very happy it is working now :)

Could we consider keeping the sparsityRatio check? The current implementation may work in cases up to sparsity ratios up to, say, 0.6. For example, GDP_NationalCurrencies.csv file in the ravel tests folder does not get loaded with arg->index().empty(), but does with the ratio set as <= 0.5 and works as in version 2.15.5.

highperformancecoder: The code on line 537 of ravelWrap assumes that the input tensor is dense, ie that the index vector is not used. Even if the sparsity ratio is <= 0.5, if the index vector is used then the code will not work correctly.

Therefore, the correct test is that index() is empty.

For now - we'll work on handling sparse data promptly.

wdednam: Sorry, I have only tested outputting to plots, reducing (summing) along a handle, moving slicers and calipers. I haven't tested any of the tensor operations. But, with the current changes I've made, higher sparsity files load and the ravel behaves just like version 2.15.5 with respect to the above simple operations.

highperformancecoder: There should be subtle differences, even with just the plots test. These will probably show up as data points misaligned with the plot labels on the x-axis.

Anyway, I think we need to drop this change now, and work on getting sparse tensors to work with Ravel. Do you need me to do this?

model/ravelWrap.cc
```diff
@@ -536,7 +536,12 @@ namespace minsky
           }
 #endif
         vector<double> tmp(v.size());
-        for (size_t i=0; i<v.size(); ++i) tmp[i]=v[i];
+        if (v.index().empty()) 
+           for (size_t i=0; i<v.size(); ++i) tmp[i]=v[i];
+        else {
+            size_t inc=0; 
+		    for (auto i: v.index()) tmp[++inc]=v.atHCIndex(i);
```
highperformancecoder: This won't work. The problem is that Ravel::DataCube assumes the data is stored densely, and applies OLAP operations on it assuming that. Here, you're just storing the data sparsely again, in fact, just implementing the same storage structure as on line 540, just less efficiently :)

engine/minskyTensorOps.cc
wdednam: Yes, the differences are very subtle (the code in this commit seems to add only an extra x and y column in the CSV file exported from the plot attached to the ravel output port for the propertyPriceRealIndex.csv file, when I compare versions 2.15.5 and 2.18.2 of Ravel with all the slicers on the "home" position and "Quarter" as the output handle). Obviously, ravel evaluation is not available, only data import and basic analysis via the slicers and calipers on the handles, so this will not do.

model/ravelWrap.cc
wdednam: I understand that In terms of ravel evaluation, it does not work. However, copying via a counter over the size of the data, as on line 540, outputs completely differently to plots, for, say, the propertyPriceRealIndex.csv file, in Minsky 2.18.2 than in 2.15.5. I only recover very similar behaviour to version 2.15.5 when I do the above. I know I'm basically taking the sparse data in the Itensor v and packing it densely into tmp, and hence the dataCube. 

My question is then: not only loadDataFromVariable() and loadSliceFromData() need to be rewritten entirely in the full-on tensor implementation, ravelDC_loadData() also does? As a related question, will this require clearing everything in the ravelState and repopulating it in RavelTensor::computeTensor(), along with assigning the cachedResult TensorVal with values that have been appropriatedly transformed from the arg input tensor pointer?

highperformancecoder:
On Sun, Mar 22, 2020 at 09:04:13AM -0700, Wynand wrote:
> 
> 
> In model/ravelWrap.cc:
> 
> > @@ -536,7 +536,12 @@ namespace minsky
>            }
>  #endif
>          vector<double> tmp(v.size());
> -        for (size_t i=0; i<v.size(); ++i) tmp[i]=v[i];
> +        if (v.index().empty())
> +           for (size_t i=0; i<v.size(); ++i) tmp[i]=v[i];
> +        else {
> +            size_t inc=0;
> +                   for (auto i: v.index()) tmp[++inc]=v.atHCIndex(i);
> 
> I understand that In terms of ravel evaluation, it does not work. However,
> copying via a counter over the size of the data, as on line 540, outputs
> completely differently to plots, for, say, the propertyPriceRealIndex.csv file,
> in Minsky 2.18.2 than in 2.15.5. I only recover very similar behaviour to
> version 2.15.5 when I do the above. I know I'm basically taking the sparse data
> in the Itensor v and packing it densely into tmp, and hence the dataCube.
> 
> My question is then: not only loadDataFromVariable() and loadSliceFromData()
> need to be rewritten entirely in the full-on tensor implementation,
> ravelDC_loadData() also does?

For Ravel in Minsky, we don't rewrite, we bypass. We reimplement those
DataCube operations within Minsky as a chain of TensorOps.

Later on, we'll factor out the Civita library, and copy the Minsky
implementation of these datacube operations into Ravel proper. This
later will be needed to support other Ravel applications, such as the
Javascript widget. Just how much later on will depend on project
priorities, which are changing daily :).

. As a related question, will this require clearing
> everything in the ravelState and repopulating it in RavelTensor::computeTensor
> (), along with assigning the cachedResult TensorVal with values that have been
> appropriatedly transformed from the arg input tensor pointer?
>

No. RavelTensor should not be altering Ravel state at all.



-- 

                      http://www.hpcoders.com.au

PR #166, opened 2020-03-26
first pass at downloading CSV files from the web and loading them via…

wdednam:
… CSVDialog.cc. I used two sources for the present implementation. First, because I need to allow the user to input the web address of the CSV file, I've attempted [this](https://stackoverflow.com/questions/30981042/tcl-tk-user-input-window) and then because I need to download the file at the input web address, I've attempted [this](https://stackoverflow.com/questions/35673588/how-to-download-a-file-using-http-package), explained further in [this entry](https://wiki.tcl-lang.org/page/Download+file+via+HTTP) at the TCL wiki page. It doesn't seem to get the file or assign the values of the variables $url and $filename correctly, but it does not throw an error either... If you don't mind, please let me know if there is anything obvious I may have missed. If what I've done requires more substantial changes, I will continue to research downloading files in TCL tomorrow.

---
This change is [Reviewable](https://reviewable.io/reviews/highperformancecoder/minsky/166)


highperformancecoder:
I took it for a bit of a spin. I note that nothing actually happens at present, and that is because your implementation of inutURL doesn't successfully pass the user's input text back in return.

TCL doesn't natively provide a text entry dialog, sadly, but the good news is that Minsky does. It is called textEntryPopup - see the definition in minsky.tcl for usage documentation. It is slightly different to tk_getOpenFile in being event based rather than completely modal, so you'll need to rework the logic a bit.

Another comment, to create and open a temporary file (unique to the current process), the TCL command "file tempfile" can be used. It is better than using $workDir/download.tar. For one, $workDir may not be writable. Secondly, .tar is an extension already used for tar files, which this is unlikely to be. Thirdly, the user may just happen to care about a file called download.tar (or any other name you might happen to pick).

wdednam:
Thank you very much for testing the code and your suggestions. :) Eek, I should definitely have considered the write permissions on folders after my experience with permissions in Sydney...

wdednam:
As the two puts statements in the CSVWebImportDialog {} proc shows, the initial web file is recognised and a temporary file file10 is created, but the "???" appear on the parameter after import, so the CSV import procedure does not recognise the file properly, even though I've added the file extension in the current directory ./*.csv.

gui-tk/csvImport.tcl
```diff
@@ -214,15 +214,18 @@ proc CSVWebImportDialog {} {
...
     
-    #set filename [tk_getOpenFile -filetypes {{CSV {.csv}} {All {.*}}} -initialfile [getFile $urlfile] ]  
+    # Example CSV file that Minsky can import, but to be replaced by user input
+    set initUrl "http://samplecsvs.s3.amazonaws.com/TechCrunchcontinentalUSA.csv"   
+    set urlFile [inputUrl $initUrl] 
+    puts $urlFile
     
-    set filename [getFile $urlfile]
+    # Create temporary file for downloaded CSV file
+    set filename [file tempfile [getFile $urlFile] ./*.csv]
```
highperformancecoder: file tempfile returns an open channel, which can be used in puts and gets statements. The file is deleted when the channel is closed.

If you need the name of the file (which you probably do), then you can request it be placed in a TCL variable via an option to file tempfile.

highperformancecoder: "man file" has all the goss on using the this command. Select the (n) category, as there's an unrelated file system command as well.

gui-tk/csvImport.tcl
```diff
@@ -275,62 +278,37 @@ proc csvWebImportDialogOK {} {
...
-    bind .csvImportURL <Return> {.csvImportURL.b invoke}
-    focus .csvImportURL.e
+proc inputUrl {initUrl} {
+    set x [get_pointer_x .wiring.canvas]
+    set y [get_pointer_y .wiring.canvas]	
+	set url $initUrl
+	textEntryPopup .inputUrl $url {inputOK $url}
+	.inputUrl.entry configure -textvariable url -takefocus 1	
+    wm geometry .inputUrl "+[winfo pointerx .]+[winfo pointery .]"    
     return $url
```
highperformancecoder: You probably want to return the URL typed by the user, rather than the name of the url variable, which may require double indirection: [set [set url]] (I suspect $$url or ${$url} may not work, but perhaps it does).

Note, on line 286, the url text variable is a global variable, not the local url variable accessed elsewhere. You can declare a variable global with the global statement. TCL variable scoping is a bit of a mess, actually.

highperformancecoder:
We definitely need https support. If it requires a 3rd party TCL package, best to download it and add it to gui-tk/library directory, just like is done for the tooltip package.

gui-tk/csvImport.tcl
wdednam: Well, at least I know I can get the contents of a CSV file loaded into Minsky, but not via the filename itself (the latest commit dumps the contents of the CSV file into the error log). In the man page, were you referring to the following?

[https://www.manpagez.com/man/n/Tcl/](
https://www.manpagez.com/man/n/Tcl/)

I could not call it from a terminal on my BSD system, but found the above page online.

gui-tk/csvImport.tcl
wdednam: This wasn't working because I forgot to add a line "set url [$win.entry get]"

wdednam:
I've added a tls.tcl to the library folder under gui-tk, but am afraid, that unlike tooltip, it won't work out of the box and might have to be added as a submodule: [https://core.tcl-lang.org/tcltls/wiki/Download](https://core.tcl-lang.org/tcltls/wiki/Download)

I've tested it and when it gets to the point where it is going to import a https url, it gives an error:

"unknown command tls::import"

Will keep trying.

gui-tk/csvImport.tcl
highperformancecoder: Actually, this one.

http://www.manpagez.com/man/n/file/

highperformancecoder: You might also find this resource rather helpful: http://tcl.tk/

highperformancecoder: I wonder why your Ubuntu doesn't have the TCL man pages? Doesn't matter - the web resources work just as well.

wdednam:
Okay, this required installing boost 1.66 (the latest stable version available from [sourceforge](https://sourceforge.net/projects/boost/)) on my machine, which does not accept anything newer than v. 1.65.1 (Ubuntu 18.04 won't let me add repositories with more recent boost versions). So, I had to install it in my home folder and then replace /usr/include/boost and the libboost files in /usr/lib too for this to work. Additionally, it requires some files from the [beast github repository](https://github.com/boostorg/beast/tree/develop/example/common) to work: e.g.,  boost/beast/example/common/root_certificates.hpp.

highperformancecoder:
Looks to me like the maintainers of the source forge project forgot to update the default download links.

The latest boost is 1.72, available from boost.org, as well as SourceForge.

You're right, it looks like Ubuntu 18.04 is starting lag behind the coalface, particularly wrt Beast. Rapiscan uses Boost 1.68, but in that case, somebody provides a privately build copy of Boost in the source code repo.

WRT using Beast, go for it. I will need to provide a more recent Boost in my OBS repo to allow the building of Minsky on Ubuntu 18.04, but so be it. Actually, it looks like I already do, except it is called boost-python3, as I needed that for another package I support.

wdednam:
> You're right, it looks like Ubuntu 18.04 is starting lag behind the coalface, particularly wrt Beast. Rapiscan uses Boost 1.68, but in that case, somebody provides a privately build copy of Boost in the source code repo.

I've now Installed boost 1.68 in /usr/local and added /usr/local/include and /usr/local/lib to my .profile. I then copied the header folders for Beast and [Certify](https://github.com/djarek/certify/tree/master/include/boost/certify) as new folders to my Minsky top directory. I changed the #include headers for them in CSVDialog.cc to the local directories in minsky/ and recompiled. It works. I've committed the code to my develop branch and will look out for the Travis build report to see if it compiles online. If you are using the same (or later) version of boost, the code should build with the above modifications.

I will clean up the code a bit, though, before committing it to this branch and also fix the TCL issue. I foresee one problem: I haven't been able to load larger or more sparse csv files with the more recent version of boost/beast, I get a "stream truncated" error. Maybe we can increase the buffer size somewhere to deal with this.

I also haven't managed to get the tmpnam() implementation to work with the latest code, although it did work in boost::asio's pre-beast http version. I'll hopefully be able to sort out these issues relatively painlessly.

wdednam:
Okay, I've removed the beast folder I had added to my develop branch because the boost on the OBS repo already seems to have it, but we still need Certify to deal with ssl certificates (the example load_certificates.hpp file in beast is not good enough to handle all ssl connections). So, I've edited all the header files in the Certify folder I added to the minsky/ folder to be local, but now I'm getting a build error about a missing openssl conf.h file:

In file included from /usr/include/boost/asio/ssl/context_base.hpp:19,

                 from /usr/include/boost/asio/ssl/context.hpp:23,

                 from /usr/include/boost/asio/ssl/stream.hpp:26,

                 from ./certify/extensions.hpp:6,

                 from model/CSVDialog.cc:33:

/usr/include/boost/asio/ssl/detail/openssl_types.hpp:23:10: fatal error: openssl/conf.h: No such file or directory

   23 | #include <openssl/conf.h>

      |          ^~~~~~~~~~~~~~~~

compilation terminated.

. Isn't there already openssl on the OBS repo? A quick search brought this up:

[https://stackoverflow.com/questions/44394644/fatal-error-openssl-conf-h-file-not-found](https://stackoverflow.com/questions/44394644/fatal-error-openssl-conf-h-file-not-found)

The solution found by the person who asked the question:

> 
I have made a mistake, my lib was outdated and did not match dependencies, that's why it did not work in my scenario. An update helped.

highperformancecoder:
Let me know when you're ready to have this code reviewed. A brief glance at it, and it looked kind of experimental. You need to remove dead code, and ensure sensible reuse of existing method, rather than piling on unnecessary complexity.

wdednam:
I will hopefully have something better to commit tomorrow. I have encountered challenges on several fronts:

Creating a temporary file to load the data into Minsky. I spent a lot of time trying to make tmpnam() and/or tmpfile() from the C++ standard library to work, but it wouldn't. Luckily, the boost [filepath library](https://stackoverflow.com/questions/5586355/c-getting-a-temporary-file-cross-platform ) finally came to the rescue today. Thanks for suggesting it. 

Then, I have dispensed with using [Certify](https://github.com/djarek/certify) for SSL connections, because it will not compile online. So, I've gone back to the header file from the old beast [124 github tag](https://github.com/boostorg/beast/tree/v124/example/common) (load_certificates.hpp) that works with boost 1.65, which seems to permit loading data from most https URLs.

Finally, I still haven't finished working on the TCL part of the feature, because I first wanted to make sure the boost::asio and boost::beast implementation works as best as I get can get it to. I also attempted two Minsky tickets 1079 (which I opened last year and assigned to myself) and 362, with very limited success in the first case and almost complete success in the second (createicons.tcl won't run on my machine because of some "::unknown" proc business -- looking into it [here](https://www.tcl.tk/man/tcl8.5/TclCmd/unknown.htm)). So, fixing csvimport.tcl will be my focus tomorrow.

wdednam:
Travis still fails :( My openssl is version 1.1.1 from September 2018. Is the OBS version more recent?

highperformancecoder:
On Sun, Apr 05, 2020 at 01:52:36PM -0700, Wynand wrote:
> Travis still fails :( My openssl is version 1.1.1 from September 2018. Is the
> OBS version more recent?

It looks like I need to add the libopenssl-devel package to the
travisci docker image.

For your reference, the project is at
https://github.com/highperformancecoder/travisCIImage. But I'll
arrange to do that.


-- 

                      http://www.hpcoders.com.au

gui-tk/csvImport.tcl
```diff
@@ -130,6 +136,10 @@ proc csvImportDialogOK {} {
     } else {
         reset
         cancelWin .wiring.csvImport
+        # Delete tempoarray file created by web import of CSV data
```
highperformancecoder, line 139: typo temporary

gui-tk/csvImport.tcl
```diff
@@ -96,7 +96,12 @@ proc CSVImportDialog {} {
             csvDialog.requestRedraw
         }
     }
-    set filename [tk_getOpenFile -filetypes {{CSV {.csv}} {All {.*}}} -initialdir $workDir]
+    
+    if [string length $url] {
+	   set filename [csvDialog.loadWebFile $url]
```
highperformancecoder, line 101: I would have arranged loadWebFile to call loadFile, then delete the temporary file, all within C++, preferably in a destructor. I would eliminate the need to pass $url as a flag to delete the temp file within TCL.

gui-tk/library/tls.tcl
```diff
@@ -0,0 +1,398 @@
+#
+# Copyright (C) 1997-2000 Matt Newman <matt@novadigm.com> 
+#
+namespace eval tls {
```
highperformancecoder, line 4: Do we still need this file?
Also no clear statement of license.

gui-tk/wiring.tcl
```diff
@@ -658,7 +658,13 @@ proc contextMenu {x y X Y} {
             }
             .wiring.context add command -label "Flip" -command "$item.flip; flip_default"
             if {[$item.type]=="parameter"} {
-                .wiring.context add command -label "Import CSV" -command CSVImportDialog
+                .wiring.context add command -label "Import CSV" -command {CSVImportDialog {}}
+                .wiring.context add command -label "Import CSV from web" -command {
+					textEntryPopup .loadWebUrl {} {CSVImportDialog [.loadWebUrl.entry get]}
+                .loadWebUrl.entry configure -takefocus 1					
+				wm title .loadWebUrl "Insert URL:"		
+                wm geometry .loadWebUrl "+[winfo pointerx .]+[winfo pointery .]"
```
highperformancecoder, line 666: Indent the code to make this clear this is part of the -command script.  I found this misleading initially.

model/CSVDialog.cc
```diff
@@ -34,6 +53,111 @@ void CSVDialog::reportFromFile(const std::string& input, const std::string& outp
...
+        boost::regex ex("(http|https)://([^/ :]+):?([^/ ]*)(/?[^ #?]*)\\x3f?([^ #]*)#?([^ ]*)");
+        boost::cmatch what;
+        regex_match(url.c_str(), what, ex);
+        
+        // The web domain 
+        auto const host = what[2];
+        
+        // Set the port type
+        string port = string(what[1].first, what[1].second);
+        if (string(what[1].first, what[1].second)=="https") port = "https";
```
highperformancecoder, line 71: I believe what[1] is directly convertible/comparable to string. If not (sometimes its not), then what.str(1) certainly is.

Also, this code block is more simply written port=what[1], as it can only be http or https.

Also, you have tested the return value of regex_match for success.

A better term for "port" is "protocol". The actual port value is given by what[3]. If what[3] is empty, the port = 80 in the case of http and 443 in the case of https.

model/CSVDialog.cc
```diff
@@ -34,6 +53,111 @@ void CSVDialog::reportFromFile(const std::string& input, const std::string& outp
...
+        regex_match(url.c_str(), what, ex);
+        
+        // The web domain 
+        auto const host = what[2];
+        
+        // Set the port type
+        string port = string(what[1].first, what[1].second);
+        if (string(what[1].first, what[1].second)=="https") port = "https"; 
+        else port = "http";
+        auto const target = string(what[4].first, what[4].second);
```
highperformancecoder, line 73: target=what.str(4)

model/CSVDialog.cc
```diff
@@ -34,6 +53,111 @@ void CSVDialog::reportFromFile(const std::string& input, const std::string& outp
   reportFromCSVFile(is,of,spec);
 }
 
+// Return file name after downloading a CSV file from the web.
+std::string CSVDialog::loadWebFile(const std::string& url)
+{
+    try
+    {
+		// Parse input URL
+        boost::regex ex("(http|https)://([^/ :]+):?([^/ ]*)(/?[^ #?]*)\\x3f?([^ #]*)#?([^ ]*)");
```
highperformancecoder, line 62: I'd probably write it as \\?? rather than \\x3f?
Also, raw strings are very useful for regexes, to avoid "leaning toothpick syndrome" 
R"(...)"

everything between the parentheses is interpreted literally - ", ' and \ have no special meaning.

model/CSVDialog.cc
highperformancecoder: what[4] should be called path_info

highperformancecoder: what[5] is the query string and what[6] is the anchor.

highperformancecoder: See https://en.wikipedia.org/wiki/URL

model/CSVDialog.cc
```diff
@@ -34,6 +53,111 @@ void CSVDialog::reportFromFile(const std::string& input, const std::string& outp
...
+        // Set the port type
+        string port = string(what[1].first, what[1].second);
+        if (string(what[1].first, what[1].second)=="https") port = "https"; 
+        else port = "http";
+        auto const target = string(what[4].first, what[4].second);  
+        
+        // Protocol version
+        int version;
+        if (!string(what[6].first, what[6].second).empty() && !std::strcmp("1.0", what[6].second)) version= 11;
+        else version=10;
```
highperformancecoder, line 78: This bit of code makes no sense to me.

model/CSVDialog.cc
highperformancecoder: The regex does not correctly handle the case of a username and/or password being provided

model/CSVDialog.cc
```diff
@@ -34,6 +53,111 @@ void CSVDialog::reportFromFile(const std::string& input, const std::string& outp
...
+        boost::certify::enable_native_https_server_verification(ctx);
+        // end::ctx_setup_source[]        
+        		        
+        // These objects perform our I/O
+        tcp::resolver resolver{ioc};
+        ssl::stream<tcp::socket> stream{ioc, ctx};        
+        
+        // tag::stream_setup_source[]
+        boost::certify::set_server_hostname(stream, string(what[2].first, what[2].second));
+        boost::certify::sni_hostname(stream, host);
```
highperformancecoder, line 100: These are both the same. Is that intended?

model/CSVDialog.cc
```diff
@@ -34,6 +53,111 @@ void CSVDialog::reportFromFile(const std::string& input, const std::string& outp
...
+        // end::stream_setup_source[]         
+
+        // Look up the domain name
+        auto const results = resolver.resolve(host, what[1]);
```
highperformancecoder, line 104: My goodness, this is much lower level code than I thought it would be.

model/CSVDialog.cc
```diff
@@ -34,6 +53,111 @@ void CSVDialog::reportFromFile(const std::string& input, const std::string& outp
...
+
+        // This buffer is used for reading and must be persisted
+        boost::beast::flat_buffer buffer;
+
+        // Declare a container to hold the response
+        http::response_parser<http::string_body> res;
+        res.body_limit((std::numeric_limits<std::uint64_t>::max)());
+
+        // Receive the HTTP response
+        http::read(stream, buffer, res);
```
highperformancecoder, line 128: You need to check the response code. Anything other than 200 OK should result in an exception being thrown.

highperformancecoder: I think - need to check what happens in the case of a redirect.

model/CSVDialog.cc
```diff
@@ -34,6 +53,111 @@ void CSVDialog::reportFromFile(const std::string& input, const std::string& outp
...
+        }
+        if (ec)
+            throw boost::system::system_error{ec}; 
+            
+        // Return the file name for loading the in csvimport.tcl 
+        return tempStr;
+
+        // If we get here then the connection is closed gracefully
+    }
+    catch(std::exception const& e)
```
highperformancecoder, line 154: This try/catch can probably be eliminated. It looks like beast does not throw, so the only exception will be the one you're throwing on line 147, plus the one you will throw on checking the response code on line 128.

gui-tk/csvImport.tcl
wdednam: Thanks. Fixed it.

gui-tk/csvImport.tcl
wdednam: I couldn't accomplish this today and decided to move on to implementing the other suggestions for improving the code you made below.

gui-tk/library/tls.tcl
wdednam: No, I've removed it, thanks.

gui-tk/wiring.tcl
wdednam: Done.

model/CSVDialog.cc
wdednam: I think I've fixed all of this now. I found the regex expressions at these two pages:

[https://stackoverflow.com/questions/2616011/easy-way-to-parse-a-url-in-c-cross-platform](
https://stackoverflow.com/questions/2616011/easy-way-to-parse-a-url-in-c-cross-platform)

[http://www.zedwood.com/article/cpp-boost-url-regex](http://www.zedwood.com/article/cpp-boost-url-regex)

model/CSVDialog.cc
wdednam: See previous comment.

model/CSVDialog.cc
wdednam: See my response two comments up.

model/CSVDialog.cc
wdednam: It seems the web importer will only work with version=10. I don't understand why it doesn't work with version=11.

model/CSVDialog.cc
wdednam: I'm not sure. See [https://github.com/djarek/certify/blob/master/examples/get_page.cpp](https://github.com/djarek/certify/blob/master/examples/get_page.cpp)

model/CSVDialog.cc
wdednam: I don't really understand it, and had to mix and match different examples of code from boost::asio, boost::beast and certify.

model/CSVDialog.cc
wdednam: I think I've fixed this now.

model/CSVDialog.cc
wdednam: Try/catch gone and added a throw to test the response code.

certify/crlset_parser.hpp
```diff
@@ -0,0 +1,73 @@
+#ifndef BOOST_CERTIFY_CRLSET_PARSER_HPP
```
highperformancecoder, line 1: Are these certify files coming from a git repo? Perhaps we can arrange to insert them as a submodule.

model/CSVDialog.cc
```diff
@@ -34,6 +53,111 @@ void CSVDialog::reportFromFile(const std::string& input, const std::string& outp
...
+  tcp::resolver resolver{ioc};
+  ssl::stream<tcp::socket> stream{ioc, ctx};        
+  
+  // tag::stream_setup_source[]. See https://github.com/djarek/certify/blob/master/examples/get_page.cpp
+  boost::certify::set_server_hostname(stream, host.str());
+  boost::certify::sni_hostname(stream, host);
+  // end::stream_setup_source[]         
+
+  // Look up the domain name
+  auto const results = resolver.resolve(host, protocol);
```
highperformancecoder, line 103: Why is the protocol required to resolve the host?

highperformancecoder:
It's looking pretty good now. I'll merge this for now, but there's two things I'd like for a future PR:
a) Seeing if we can add the certify stuff as a submodule. Either that, or we pare it down to the minimum code needed to support what we want.
b) Moving the handling of deleting the temporary file into C++

certify/crlset_parser.hpp
wdednam: I just included the header folder of the repository: [https://github.com/djarek/certify/tree/master/include/boost/certify](https://github.com/djarek/certify/tree/master/include/boost/certify)

That was the reason I didn't add it as a submodule, because for our purposes it seems that we only need the headers.

model/CSVDialog.cc
wdednam: One could use either the protocol or the port, I've seen examples doing both online.

wdednam:
I forgot two lines in minsky.tcl which lead to cascade of test failures. They are related to sourcing tls.tcl. I've removed them and have checked that the tests pass now. I will recommit and open a new PR.

highperformancecoder:
> 
> I just included the header folder of the repository: https://github.com/djarek/
> certify/tree/master/include/boost/certify
> 
> That was the reason I didn't add it as a submodule, because for our purposes it
> seems that we only need the headers.

This makes it even better for use as a submodule. We don't need to
build anything, just check out the repo in place, and add the include
dir to our include path.

Being a submodule gives us the opportunities to pick up bugfixes and
improvements later on if needed, otherwise, we can just leave it fixed
at the git commit that works for us.


-- 

                      http://www.hpcoders.com.au

highperformancecoder:
On Tue, Apr 07, 2020 at 11:32:43PM -0700, Wynand wrote:
> I forgot two lines in minsky.tcl which lead to cascade of test failures. They
> are related to sourcing tls.tcl. I've removed them and have checked that the
> tests pass now. I will recommit and open a new PR.

I beat you to it :)



-- 

                      http://www.hpcoders.com.au

wdednam:
> I just included the header folder of the repository: https://github.com/djarek/ certify/tree/master/include/boost/certify That was the reason I didn't add it as a submodule, because for our purposes it seems that we only need the headers.
> This makes it even better for use as a submodule. We don't need to build anything, just check out the repo in place, and add the include dir to our include path. Being a submodule gives us the opportunities to pick up bugfixes and improvements later on if needed, otherwise, we can just leave it fixed at the git commit that works for us.
> […](#)
> --
> ---------------------------------------------------------------------------- Dr Russell Standish Phone 0425 253119 (mobile) Principal, High Performance Coders hpcoder@hpcoders.com.au http://www.hpcoders.com.au ----------------------------------------------------------------------------

I should not add the whole /certify repo, though, should I? I'm asking because I had some bad issues when I tried to add the libcurl C++ wrapper. In that case, although I only added a header file, the build would not work.

So, do I just add the subdirectory as explained [here](https://stackoverflow.com/questions/9035895/how-do-i-add-a-submodule-to-a-sub-directory)?

wdednam:
> On Tue, Apr 07, 2020 at 11:32:43PM -0700, Wynand wrote: I forgot two lines in minsky.tcl which lead to cascade of test failures. They are related to sourcing tls.tcl. I've removed them and have checked that the tests pass now. I will recommit and open a new PR.
> I beat you to it :)
> […](#)
> --
> ---------------------------------------------------------------------------- Dr Russell Standish Phone 0425 253119 (mobile) Principal, High Performance Coders hpcoder@hpcoders.com.au http://www.hpcoders.com.au ----------------------------------------------------------------------------

Thanks :)

wdednam:
> > ## I just included the header folder of the repository: https://github.com/djarek/ certify/tree/master/include/boost/certify That was the reason I didn't add it as a submodule, because for our purposes it seems that we only need the headers.
> > This makes it even better for use as a submodule. We don't need to build anything, just check out the repo in place, and add the include dir to our include path. Being a submodule gives us the opportunities to pick up bugfixes and improvements later on if needed, otherwise, we can just leave it fixed at the git commit that works for us.
> > […](#)
> > ---------------------------------------------------------------------------- Dr Russell Standish Phone 0425 253119 (mobile) Principal, High Performance Coders [hpcoder@hpcoders.com.au](mailto:hpcoder@hpcoders.com.au) http://www.hpcoders.com.au ----------------------------------------------------------------------------
> 
> I should not add the whole /certify repo, though, should I? I'm asking because I had some bad issues when I tried to add the libcurl C++ wrapper. In that case, although I only added a header file, the build would not work.
> 
> So, do I just add the subdirectory as explained [here](https://stackoverflow.com/questions/9035895/how-do-i-add-a-submodule-to-a-sub-directory)?

Okay, I see what I will have to clone the whole repo as a submodule:
[https://stackoverflow.com/questions/5303496/how-to-change-a-git-submodule-to-point-to-a-subfolder](
https://stackoverflow.com/questions/5303496/how-to-change-a-git-submodule-to-point-to-a-subfolder)

PR #170, opened 2020-04-08
Feature 50

wdednam:
This is close to being done. The tearoff menus seem not to work even though I've added the fundamental constants as "constant" operations just like the other operations. Tests will be needed once they are working properly.

---
This change is [Reviewable](https://reviewable.io/reviews/highperformancecoder/minsky/170)


wdednam:
Have to comment out line 149 in wiring.tcl:

tooltip .wiring.menubar.constops.menu -index 0 "constant operations toolbox"

as Minsky won't launch otherwise because tooltip can't add the "constant" operations I have added and I don't understand why... 

The error in question occurs on lines 185-188 of tooltip.tcl:

		if {[catch {$w entrycget 1 -label}]} {
		    return -code error "widget \"$w\" does not seem to be a\
			    menu, which is required for the -index switch"
		}

model/CSVDialog.cc
```diff
@@ -71,10 +71,10 @@ std::string CSVDialog::loadWebFile(const std::string& url)
          
   auto const protocol =what[1];
   auto const host = what[2];
-  auto const port = what[3];
```
highperformancecoder, line 74: What's this file doing here?

model/operation.cc
```diff
@@ -516,6 +516,9 @@ namespace minsky
   }
 
   Units Time::units(bool) const {return cminsky().timeUnit;}
+  //Units Eu::units(bool) const {}
+  //Units Pi::units(bool) const {}
+  //Units Fb::units(bool) const {}
```
highperformancecoder, line 521: Dead code

model/operation.cc
```diff
@@ -637,6 +640,9 @@ namespace minsky
     switch (type)
       {
       case time: return new Time;
+      //case euler: return new Eu;
+      //case pi: return new Pi;
+      //case feigenbaum: return new Fb;
```
highperformancecoder, line 645: Dead code

model/operation.h
```diff
@@ -139,6 +139,24 @@ namespace minsky
...
+  //public:
+  //  Units units(bool) const override;
+  //};        
+  //
+  //class Fb: public Operation<OperationType::feigenbaum>
+  //{
+  //public:
+  //  Units units(bool) const override;
+  //};    
+
```
highperformancecoder, line 159: Dead code

schema/operationType.h
```diff
@@ -27,7 +27,7 @@ namespace minsky
   struct OperationType
   {
     enum Type {constant, // deprecated - left to support legacy schemas
-               time, // zero input port ops
+               time, euler, pi, feigenbaum, // zero input port ops
```
highperformancecoder, line 30: You need to extend the classification Group enum, on line 56.

schema/operationType.cc
```diff
@@ -33,6 +33,7 @@ namespace minsky
   {
       if (t<add) return general;
       if (t<copy) return binop;
+      if (t<integrate) return constop;
```
highperformancecoder, line 36: This fails, as anything less than integrate will be classified as general on line 34.
Best would be to place the constops after Ravel, and before add, then you can use euler (for general) and add (for constops) here.

engine/minskyTensorOps.cc
```diff
@@ -186,8 +194,10 @@ namespace minsky
   TensorOpFactory::TensorOpFactory()
   {
     registerType<TimeOp>(OperationType::time);
-    registerOps<MultiWireBinOp, OperationType::add, OperationType::log>(*this);
+    registerOps<MultiWireBinOp, OperationType::add, OperationType::log>(*this); 
     registerOps<TensorBinOp, OperationType::log, OperationType::copy>(*this);
+    registerType<ConstOp>(OperationType::euler);
+    registerType<ConstOp>(OperationType::pi);
```
highperformancecoder, line 200: This will cause the factory to generate the same object (a ConstOp), regardless of the constant. This would be an error, if it weren't for the fact that these operations are scalar, and hence the tensor factory will not be called anyway.

highperformancecoder:
Should add an example file using both of these constants, this will act as a regression test too.

BTW - have you adapted the LaTeX and Matlab export code?

engine/minskyTensorOps.cc
```diff
@@ -186,8 +186,9 @@ namespace minsky
   TensorOpFactory::TensorOpFactory()
   {
     registerType<TimeOp>(OperationType::time);
-    registerOps<MultiWireBinOp, OperationType::add, OperationType::log>(*this);
-    registerOps<TensorBinOp, OperationType::log, OperationType::copy>(*this);
+    registerOps<MinskyTensorOp, OperationType::euler, OperationType::pi>(*this);
```
highperformancecoder, line 189: The second argument is exclusive, not inclusive, like a container range (begin(), end()). So the second argument should be whatever comes after pi.

engine/node_latex.cc
```diff
@@ -374,6 +374,18 @@ namespace MathDAG
...
+  template <>
+  ostream& OperationDAG<OperationType::euler>::latex(ostream& o) const
+  {
+    return o<<" e ";
+  }
+ 
+  template <>
+  ostream& OperationDAG<OperationType::pi>::latex(ostream& o) const
+  {
+    return o<<" π ";
```
highperformancecoder, line 387: LaTeX for pi is \pi.

engine/node_matlab.cc
```diff
@@ -321,6 +321,18 @@ namespace MathDAG
...
+  template <>
+  ostream& OperationDAG<OperationType::euler>::matlab(ostream& o) const
+  {
+    return o<<"e";
+  }
+
+  template <>
+  ostream& OperationDAG<OperationType::pi>::matlab(ostream& o) const
+  {
+    return o<<"π";
```
highperformancecoder, line 334: Matlab for pi is "pi"

schema/operationType.cc
```diff
@@ -31,8 +31,9 @@ namespace minsky
 
   OperationType::Group OperationType::classify(Type t)
   {
-      if (t<add) return general;
-      if (t<copy) return binop;
+	  if (t<add) return constop; 
+      if (t<euler) return general;
```
highperformancecoder, line 35: Line 35 needs to appear before line 34.

engine/minskyTensorOps.cc
wdednam: Thanks. Fixing it now.

engine/node_latex.cc
wdednam: Thanks for catching this. My goodness, all the articles I've published were written in Latex. It does not cease to amaze me how being in a different context can render acquired knowledge/skills almost useless...

engine/node_matlab.cc
wdednam: Fixed this too.

schema/operationType.cc
wdednam: Swapped them, thanks.

test/testEqWrappedPrototype.tex
```diff
@@ -42,7 +48,8 @@
 k=\Theta\left(\Theta\left( t \right)-0.5\right)\Theta\left(\Theta\left( t \right)+\delta\left( t \right)-0.5\right)+0+0
 \end{dmath*}
 \begin{dmath*}
-l=\left(1-\Theta(0.5-max\left(\Theta\left(\Theta\left(\mathrm{Repay}- t \right)+\delta\left( t -\mathrm{Repay}\right)-0.5\right),\Theta\left(\Theta\left( t \right)\right)\right)\right)ight)+\Theta\left(\Theta\left( t \right)-0.5\right)+\Theta\left(\Theta\left( t \right)-0.5\right)
```
highperformancecoder, line 45: Whoopsies! Thanks for catching and fixing that.

wdednam: Was not aware of them. TexMaker sorted them out. :)

PR #172, opened 2020-04-12
Add certify as submodule. Compiles only locally but not remotely. 

wdednam:
This change is [Reviewable](https://reviewable.io/reviews/highperformancecoder/minsky/172)


wdednam:
I followed the instructions at [https://github.blog/2016-02-01-working-with-submodules/](https://github.blog/2016-02-01-working-with-submodules/) and don't understand why Minsky can't pick up headers within headers of certfiy/ subdirectory when it compiles online. I can't modify them because then it wants to commit changes to the original certify repository and not my own. I don't have this problem locally.

model/CSVDialog.cc
```diff
@@ -32,8 +32,8 @@
 #include <boost/regex.hpp>
 #include <boost/filesystem.hpp>
    
-#include "certify/extensions.hpp"         
-#include "certify/https_verification.hpp" 
+#include "certify/include/boost/certify/extensions.hpp"         
+#include "certify/include/boost/certify/https_verification.hpp"
```
highperformancecoder, line 36: I would have left this header include as is, and added  certify/boost to the include path, but this will work as well (CERTIFY_HOME stuff above is doing nothing in that case).

PR #173, opened 2020-04-15
Feature 94

wdednam:
For some reason the setters for width() and height() I added in the RenderVariable class do not change the values of the private members w and h of that class. I made them mutable but this still didn't make a difference. moveTo in VariableBase::resize() seems to work though...

---
This change is [Reviewable](https://reviewable.io/reviews/highperformancecoder/minsky/173)

### DIFF
--- a/model/cairoItems.h
+++ b/model/cairoItems.h
@@ -38,6 +38,7 @@ namespace minsky
   public:
     // render a variable to a given cairo context
     RenderOperation(const OperationBase& var, cairo_t* cairo=NULL);
+    
     /// render the cairo image
     void draw();
     /// half width of unrotated image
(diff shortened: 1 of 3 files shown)


highperformancecoder:
I think this is the wrong move. RenderVariable/ RenderOperation are about rendering and measuring the space onscreen icons take. The width() and height() attributes are not editable - they are calculated values depending on the variable or operation they are constructed from. In some ways, these classes ought to be folded back into the regular Item classes, which has the BoundingBox element performing the same function, but that would be a further refactoring exercise, that is probably not warranted at the present time.

What you need to do is merge something like the sheet implementation into the general Item class implementation. Things which are currently resizable are GodleyIcons, PlotWidgets, Ravels and Sheets. They should all be refactored to use the new method where possible - Sheet being the latest, and probably simplest implementation.

wdednam:
I understand, thanks.I guess I should take the most general case and then use if statements or a switch to specialize to different canvas objects.

highperformancecoder:
On Thu, Apr 16, 2020 at 02:17:25AM -0700, Wynand wrote:
> I understand, thanks.I guess I should take the most general case and then use
> if statements or a switch to specialize to different canvas objects.

Or virtual function override. Switches are considered a bit of a
smell, but still very common.


-- 

                      http://www.hpcoders.com.au

PR #174, opened 2020-04-18
going with virtual override of resize method in item class. Only moveTo working within VariableBase::resize(), don't understand why

wdednam:
This change is [Reviewable](https://reviewable.io/reviews/highperformancecoder/minsky/174)


model/item.cc
```diff
@@ -182,6 +182,16 @@ namespace minsky
     }
   }
   
+  // Refactor resize() code for all canvas items here. For feature 25
+  void Item::resize(const LassoBox& b)
+  {
+    float invZ=1/zoomFactor();
+    bb.right=abs(b.x1-b.x0)*invZ+bb.left;
+    bb.bottom=abs(b.y1-b.y0)*invZ+bb.top;
```
highperformancecoder, line 190: I don't think it is a good idea updating these here.

model/item.h
```diff
@@ -65,8 +65,8 @@ namespace minsky
   /// bounding box information (at zoom=1 scale)
   class BoundingBox
   {
-    float left=0, right=0, top, bottom;
   public:
+    float left=0, right=0, top, bottom;
```
highperformancecoder, line 69: These fields really should remain private, and only updated via the update() call, which uses cairo_ink_extents.

My thought is Item should sport two new member fields (m_width, m_height - say), and these are used to control how large the icon is on campus.

model/variable.cc
```diff
@@ -453,7 +459,7 @@ void VariableBase::draw(cairo_t *cairo) const
   rv.angle=angle+(notflipped? 0: M_PI);
 
   // parameters of icon in userspace (unscaled) coordinates
-  float w, h, hoffs;
+  float hoffs;
   w=rv.width()*z; 
   h=rv.height()*z;
```
highperformancecoder, line 464: Where are w and h declared now?

model/variable.h
```diff
@@ -89,6 +89,8 @@ namespace minsky
     void addPorts();
     
   public:
+    mutable float w, h;
```
highperformancecoder, line 92: I don't think this is a good idea.

model/operation.cc
```diff
@@ -135,6 +135,24 @@ namespace minsky
...
+    Rotate r(rotation()+(notflipped? 0: 180),0,0); // rotate into variable's frame of reference
+    double z=zoomFactor();
+    try
+      {
+        double dx=xx-x(), dy=yy-y();
+        if (fabs(fabs(dx)-iWidth()) < portRadius*z &&
+            fabs(fabs(dy)-iHeight()) < portRadius*z &&
+            fabs(hypot(dx,dy)-hypot(iWidth(),iHeight())) < portRadius*z)
+          return ClickType::onResize;
+      }
```
highperformancecoder, line 152: Why the try/catch block? What exception might be thrown here?

model/operation.cc
```diff
@@ -221,6 +240,7 @@ namespace minsky
               // we need to add some translation if the variable is bound
               cairo_rotate(cairo,rotation()*M_PI/180.0);
               coupledIntTranslation=-0.5*(i->intVarOffset+2*rv.width()+2+r)*z;
+              //if (rv.width()<iWidth()) coupledIntTranslation=-0.5*(i->intVarOffset+2*iWidth()+2+r)*z;
               //            cairo_translate(cairo, coupledIntTranslation, 0);
```
highperformancecoder, line 244: remove dead code

model/operation.cc
```diff
@@ -297,6 +321,8 @@ namespace minsky
           cairo_line_to(cairo,r,0);
           cairo_line_to(cairo,r+ivo,0);
           float rvw=rv.width()*z, rvh=rv.height()*z;
+          //if (rv.width()<iWidth()) rvw=iWidth()*z;
+          //if (rv.height()<iHeight()) rvh=iHeight()*z;
```
highperformancecoder, line 325: Dead code?

model/operation.cc
```diff
@@ -360,13 +386,24 @@ namespace minsky
...
     if (selected) drawSelected(cairo);
   }
+  
+  void OperationBase::resize(const LassoBox& b)
+  {
+    float w=iWidth(), h=iHeight(), invZ=1/zoomFactor();
+    iWidth(std::abs(b.x1-b.x0)*invZ);
+    iHeight(std::abs(b.y1-b.y0)*invZ);
+    moveTo(0.5*(b.x0+b.x1), 0.5*(b.y0+b.y1));
+    bb.update(*this);
```
highperformancecoder, line 404: Given that iWidth and iHeight call bb.update, this is a redundant third call. May need to perform moveTo before the calls to iWidth and iHeight.

model/operation.cc
```diff
@@ -360,13 +386,24 @@ namespace minsky
...
 
     cairo_new_path(cairo);
     clipPath.appendToCurrent(cairo);
     cairo_clip(cairo);
     if (selected) drawSelected(cairo);
   }
+  
+  void OperationBase::resize(const LassoBox& b)
+  {
+    float w=iWidth(), h=iHeight(), invZ=1/zoomFactor();
```
highperformancecoder, line 400: w and h are not used.

model/switchIcon.cc
```diff
@@ -114,13 +129,22 @@ namespace minsky
...
     cairo_clip(cairo);
      if (selected) drawSelected(cairo);
   }
+  
+  void SwitchIcon::resize(const LassoBox& b)
+  {
+    float invZ=1/zoomFactor();
+    iWidth(abs(b.x1-b.x0)*invZ);
+    moveTo(0.5*(b.x0+b.x1), 0.5*(b.x0+b.x1));
+    bb.update(*this);
```
highperformancecoder, line 146: Ditto with previous comment about iWidth call bb.update

model/variable.cc
```diff
@@ -82,8 +82,17 @@ ClickType::Type VariableBase::clickType(float xx, float yy)
     {
       double hpx=z*rv.handlePos();
       double hpy=-z*rv.height();
-      if (type()!=constant && hypot(xx-x() - r.x(hpx,hpy), yy-y()-r.y(hpx,hpy)) < 5)
+      if (rv.height()!=iHeight()) hpy=-z*iHeight();
```
highperformancecoder, line 85: Why not set hpy=-z*iHeight() already on line 84?

model/variable.cc
```diff
@@ -453,10 +462,16 @@ void VariableBase::draw(cairo_t *cairo) const
   rv.angle=angle+(notflipped? 0: M_PI);
 
   // parameters of icon in userspace (unscaled) coordinates
-  float w, h, hoffs;
+  float w, h, hoffs, fontFactor;
   w=rv.width()*z; 
   h=rv.height()*z;
+  fontFactor=min(iWidth()/rv.width(),iHeight()/rv.height());
```
highperformancecoder, line 468: fontFactor=min(1.0, min(iWidth()/rv.width(),iHeight()/rv.height()));

would eliminate the conditional on line 472. Just need to call setFontSize()

model/variable.cc
```diff
@@ -552,18 +567,27 @@ void VariableBase::draw(cairo_t *cairo) const
...
 
-void VariablePtr::makeConsistentWithValue()
+void VariableBase::resize(const LassoBox& b)
 {
-  retype(minsky::cminsky().variableValues[get()->valueId()].type());
+  float w=iWidth(), h=iHeight(), invZ=1/zoomFactor();
+  iWidth(abs(b.x1-b.x0)*invZ);
+  iHeight(abs(b.y1-b.y0)*invZ);
+  moveTo(0.5*(b.x0+b.x1), 0.5*(b.y0+b.y1));
+  bb.update(*this);
```
highperformancecoder, line 585: Ditto

highperformancecoder:
Looking pretty good now. Just some minor flaws to fix up, and ensure the regression suite passes.

test/testModel.cc
```diff
@@ -430,11 +430,12 @@ SUITE(Canvas)
...
-      canvas.mouseDown(c->x(),c->y());
-      canvas.mouseUp(400,500);
-      CHECK_EQUAL(400,c->x());
-      CHECK_EQUAL(500,c->y());
+      RenderVariable rv(*dynamic_cast<VariableBase*>(c.get()));
+      CHECK(c->clickType(c->x()+0.5*rv.width(),c->y()+0.5*rv.height()) == ClickType::onItem); // offset added because resize handles grabbed otherwise, for feature 94
+      canvas.mouseDown(c->x()+0.5*rv.width(),c->y()+0.5*rv.height());
+      canvas.mouseUp(500,500);
+      CHECK_EQUAL(500,c->x()+0.5*rv.width());
+      CHECK_EQUAL(500,c->y()+0.5*rv.height());
```
highperformancecoder, line 438: I think you've changed the nature of this test. The former test tests that the item is moved to (400,500). The latter test tests that the item was resized such that the top right corner is at (500,500).

Both are valid tests to have in the regression suite,

test/testModel.cc
```diff
@@ -445,7 +446,7 @@ SUITE(Canvas)
       cv->sliderMax=2000;
       // work out where slider is located
       RenderVariable rv(*cv);
-      float xc=cv->x()+rv.handlePos(), yc=cv->y()-rv.height();
+      float xc=cv->x()+rv.handlePos(), yc=cv->y();
```
highperformancecoder, line 449: I don't get this How can the mouse be on the slider when it is in the middle of the variable?

test/testModel.cc
highperformancecoder: OTOH, clickType is returning onItem. Should be onResizeHandle...

test/testModel.cc
wdednam: I've managed to fix this, so that's good.

test/testModel.cc
wdednam: This is proving to be somewhat more difficult than I had expected. The resize handles only appear in the correct place when I first create the default canvas item (a note). After grabbing the handles and pulling them, the resize handles then only reappear depending on the extent of zoom-in/out. The more zoomed in, the more easily I can "find" them, but they also appear slightly offset from their original position (towards the center of the note's bounding box). Reminds me of the problem Kresimir experiences with the disappearing canvas items...

wdednam:
Still struggling with this one. Apart from resizing the default item, the coupled integral is also difficult to wrap my head around. I've tried cairo_scale() in variable.cc, using an if condition (type()==integral), in cairoItems.cc in RenderVariable::draw(), and in both spots where the operation is dynamic_cast to an IntOp object within OperationBase::draw(). Nothing happens to the red variable attached to the triangle with an integral in it. I'll clean it all up a bit and commit the code as comments, so you can see what I've tried. Sorry...

test/testModel.cc
```diff
@@ -430,11 +430,11 @@ SUITE(Canvas)
     {
       cairo::Surface surf(cairo_recording_surface_create(CAIRO_CONTENT_COLOR,nullptr));
       c->draw(surf.cairo());// reposition ports
-      CHECK(c->clickType(c->x(),c->y()) == ClickType::onItem);
-      canvas.mouseDown(c->x(),c->y());
+      CHECK(c->clickType(c->x()+5,c->y()+5) == ClickType::onItem); // small offset added because resize handles grabbed otherwise, for feature 94
```
highperformancecoder, line 433: OK, but why does clickType(c->x(), c->y()) produce onResizeHandles? That is not right.

highperformancecoder:
I understand this is still WIP. There's quite a bit of dead code, so make sure that's removed once you're done. 

Also be a bit mindful of float <=> double conversions. These are not cost-free, and can add up. Float is adequate for coordinate purposes, and much of Minsky's API is based on that predicate. However, cairo takes doubles, so there will at least need to be float => double conversion at some point. But we can minimise (by declaring intermediate variables appropriately) the number of such conversions.

test/testModel.cc
wdednam: Yes, I agree, but I still don't understand why either. I guess it has something to do with differences between width() and height() in item.h, on one hand, and iWeight() and iHeight(), on the other. That's the only explanation I can find. I've tried setting width() and height() to iWidth() and iHeight(), but that hasn't worked either. I will keep trying.

wdednam:
> Also be a bit mindful of float <=> double conversions.

Thank you, I will keep this in mind and make sure I minimise those conversions once I have figured out why the default item, a note, does not behave consistently when it comes to setting iWidth() and iHeight().

model/canvas.cc
```diff
@@ -168,6 +168,10 @@ namespace minsky
         if (item)
           {
             item->resize(lasso);
+            // resize both operator and variable in coupled integral variable. for feature 94
+            if (auto i=dynamic_cast<IntOp*>(item.get()))
+               if (i->coupled())
+                 i->intVar->resize(lasso);
```
highperformancecoder, line 174: Would this work correctly? ISTM, the lasso should refer to resizing the integral op, and then the intVar needs to be scaled accordingly.

model/item.cc
```diff
@@ -129,14 +129,21 @@ namespace minsky
...
         if (hypot(x-p->x(), y-p->y()) < portRadius*zoomFactor())
           return ClickType::onPort;
-      }
+      }          
+     
+    // then, check whether a resize handle has been selected  
+	double w=0.5*width()*zoomFactor(), h=0.5*height()*zoomFactor(); 
+    if (abs(abs(x-this->x())-w) < portRadiusMult*zoomFactor() &&	  
+            abs(abs(y-this->y())-h) < portRadiusMult*zoomFactor() &&	  
+            abs(hypot((x-this->x()),(y-this->y()))-hypot(w,h)) < portRadiusMult*zoomFactor())
```
highperformancecoder, line 138: Why is the hypotenuse calculation required?

highperformancecoder:
I took the code for a quick spin, but it still needs some work to get everything scaling appropriately. Also, note that the ∫dt test is not scaling, nor are e and π. But in other operations the text does scale.

model/canvas.cc
wdednam: I have found no other way to scale the variable otherwise. Issuing cairo_scale() in RenderVariable::draw() was a bust, adding cairo_scale() in operation.cc where there are if (i->coupled()) has also been fruitless. I also added intop casts in VariableBase::draw() to scale the attached variable which is accessed via i->intVar, and this also had no effect. This latest attempt seems to work, but there seems to be something wrong with the outport and the size of the bounding box, which is what you may be referring to.

model/item.cc
wdednam: I copied it from clickType() in sheet.cc. I assumed that it is somehow needed to test for clicktype on resize inside all four corners of the bounding box.

wdednam:
Thanks, this has helped me realize that I have to add 0.5*scaleFactor() in cairo_move_to(), cairo_line_to() and cario_arc() in cases where no font size is specified and the operator symbols are drawn by hand. I'm busying adding in those now.

model/item.cc
highperformancecoder: Fair enough, but it looks redundant to me. Probably my mistake, for being overzealous.

model/canvas.cc
highperformancecoder: Why wouldn't you do something like:
```
float oldIWidth=iWidth();
iWidth(std::abs(b.x1-b.x0)*invZ);
intVar->iWidth(intVar->iWidth() * iWidth/oldOWidth);
```
ditto for iHeight

wdednam: OK. Not sure if you meant inside canvas.cc? I did this:

      case LassoMode::itemResize:
        if (item)
          {
            item->resize(lasso);
            // resize both operator and variable in coupled integral variable. for feature 94
            if (auto i=dynamic_cast<IntOp*>(item.get()))
               if (i->coupled()) {
				  float invZ=1.0/i->zoomFactor(); 
                  float oldIWidth=i->iWidth();
                  i->iWidth(std::abs(lasso.x1-lasso.x0)*invZ);
                  i->intVar->iWidth(i->intVar->iWidth() * i->iWidth()/oldIWidth);
                  
                  float oldIHeight=i->iHeight();
                  i->iHeight(std::abs(lasso.y1-lasso.y0)*invZ);
                  i->intVar->iHeight(i->intVar->iHeight() * i->iHeight()/oldIHeight);
			  }
               
            //     i->intVar->resize(lasso);
            requestRedraw();
          }
        break;

But this again only resizes the integral... Perhaps you meant inside VariableBase::resize()?

model/item.cc
wdednam: I thought it dealt with off diagonal cases.

model/canvas.cc
wdednam: Great, thanks. This seems to work:

              float oldIWidth=i->iWidth();
              i->intVar->iWidth(std::abs(lasso.x1-lasso.x0)*invZ);
              i->intVar->iWidth(i->intVar->iWidth() * i->iWidth()/oldIWidth);

and likewise for the height.

Just fixing the scaling of operator symbols within boxes now.

highperformancecoder:
minsky.xsd is a built file. It shouldn't be under version control.

model/canvas.cc
```diff
@@ -167,7 +167,27 @@ namespace minsky
       case LassoMode::itemResize:
         if (item)
           {
-            item->resize(lasso);
+            // resize both operator and variable in coupled integral variable. for feature 94
+            if (auto i=dynamic_cast<IntOp*>(item.get())) {
+               if (i->coupled()) {
+				  i->resize(lasso);
```
highperformancecoder, line 173: This line is common to all branches of the conditional, so should be moved before the first if statement.

model/operation.cc
```diff
@@ -555,6 +598,25 @@ namespace minsky
...
         g->removeItem(*intVar);
   }
+  
+  float IntOp::x() const 
+  {
+    if (auto g=group.lock())
+      return zoomFactor()*m_x+g->x();
+    else {
+	  // adjust x() for coupled integral resizing operation. for feature 94.	
+	  if (coupled()) return intVar->m_x;
```
highperformancecoder, line 608: I don't see that this will do anything. Almost everything will be handled on line 605.

model/operation.cc
```diff
@@ -555,6 +598,25 @@ namespace minsky
...
+     }
+  }
+
+  float IntOp::y() const 
+  {
+    if (auto g=group.lock())
+      return zoomFactor()*m_y+g->y();
+    else
+      return m_y;
+  }
```
highperformancecoder, line 619: Is this different than Item::y()?

model/operation.cc
```diff
@@ -921,37 +1002,49 @@ namespace minsky
 
   template <> void Operation<OperationType::le>::iconDraw(cairo_t* cairo) const
   {
+	float sf = scaleFactor();
```
highperformancecoder, line 1005: In these cases, declaring sf double would be better. cairo_scale takes double arguments.

test/testModel.cc
```diff
@@ -430,12 +430,24 @@ SUITE(Canvas)
...
+    TEST_FIXTURE(TestFixture,resizeItem)
+    {
+      cairo::Surface surf(cairo_recording_surface_create(CAIRO_CONTENT_COLOR,nullptr));
+      c->draw(surf.cairo());// reposition ports
+      float xc=c->x()+0.5*c->width(), yc=c->y()+0.5*c->height();      
+      CHECK(c->clickType(xc,yc) == ClickType::onResize); 
+      canvas.mouseDown(xc,yc);
+      canvas.mouseUp(800,1000);
+      CHECK_EQUAL(1062,c->x()+0.5*c->width());
+      CHECK_EQUAL(1460,c->y()+0.5*c->height());
```
highperformancecoder, line 449: Left hand constants are brittle here. Better to calculate them - eg 800-xc == 0.5*c->width(), etc.

test/testModel.cc
```diff
@@ -600,9 +612,9 @@ SUITE(Canvas)
         CHECK_EQUAL(2,model->numItems());
         CHECK_EQUAL(0,g->inVariables.size());
 
-        // move b into group
-        mouseDown(b->x(),b->y());
-        mouseUp(g->x(),g->y());
+        // move b into group. 
+        mouseDown(b->x()+5,b->y()+5);   
+        mouseUp(g->x()+5,g->y()+5);  // small offset added because resize handles grabbed otherwise, for feature 94. don't understand why?
```
highperformancecoder, line 617: I still don't understand the need for the offsets here...

test/testModel.cc
```diff
@@ -611,7 +623,7 @@ SUITE(Canvas)
         // move b out of group
         item=g;
         zoomToDisplay();
-        mouseDown(b->x(),b->y());
+        mouseDown(b->x()+5,b->y()+5);
```
highperformancecoder, line 626: ditto.

model/canvas.cc
wdednam: I fixed this by providing IntOp with its own draw() and drawResizeHandles() methods. So, now only item->resize(lasso).

model/operation.cc
wdednam: I got rid of this method and use intVar->x() instead at appropriate places.

model/operation.cc
wdednam: No, so I've removed it.

model/operation.cc
wdednam: Done :)

test/testModel.cc
wdednam: Sort of fixed this... But the default item does not resize properly. Is it perhaps a Pango related issue?

test/testModel.cc
wdednam: Honestly, I don't understand it either.

test/testModel.cc
```diff
@@ -444,9 +444,9 @@ SUITE(Canvas)
       float xc=c->x()+0.5*c->width(), yc=c->y()+0.5*c->height();      
       CHECK(c->clickType(xc,yc) == ClickType::onResize); 
       canvas.mouseDown(xc,yc);
-      canvas.mouseUp(800,1000);
-      CHECK_EQUAL(1062,c->x()+0.5*c->width());
-      CHECK_EQUAL(1460,c->y()+0.5*c->height());
+      canvas.mouseUp(600,800);
+      CHECK_CLOSE(600-xc,0.5*c->width(),4*portRadiusMult); // I know, volkswagening. I don't understand why, but default item does not resize predictably. Pango problem?
+      CHECK_CLOSE(800-yc,0.5*c->height(),4*portRadiusMult);
```
highperformancecoder, line 449: This isn't right. Given that the top right corner should now be at (600,800), the thing to test is that 600-c->x() is equal to 0.5*c->width(). Ditto in the y direction.

highperformancecoder:
I like the changes you made to push more functionality in virtual function overrides in IntOp. Definitely reduces the code smell.

model/operation.cc
```diff
@@ -135,7 +135,38 @@ namespace minsky
...
+      return ClickType::onResize;  
+    return Item::clickType(xx,yy);
+  }
+  
+  float OperationBase::scaleFactor() const
+  {
+	float z=zoomFactor();
+    float l=OperationBase::l*z, r=OperationBase::r*z, 
+      h=OperationBase::h*z;
+    return std::max(1.0,std::min(static_cast<double>(iWidth())*z/(std::max(l,r)),static_cast<double>(iHeight())*z/h));
```
highperformancecoder, line 167: You could maintain everything in float precision by stating the constant as 1.0f

model/operation.cc
```diff
@@ -161,15 +192,16 @@ namespace minsky
           auto& c=dynamic_cast<const DataOp&>(*this);
           
           Pango pango(cairo);
-          pango.setFontSize(10*z);
+          pango.setFontSize(10*scaleFactor()*z);
           pango.setMarkup(latexToPango(c.description()));
           pango.angle=angle + (textFlipped? M_PI: 0);
+          //Rotate r(rotation()+ (textFlipped? 180: 0),0,0);
```
highperformancecoder, line 198: dead code

model/item.cc
```diff
@@ -204,19 +233,23 @@ namespace minsky
   void Item::draw(cairo_t* cairo) const
   {
     Rotate r(rotation(),0,0);
+    //cairo_scale(cairo,scaleFactor(),scaleFactor());   // would like to see this work like operator icon contents, but width() and height() point to nothing at this stage.
```
highperformancecoder, line 236: dead code?

model/operation.cc
```diff
@@ -536,6 +509,194 @@ namespace minsky
...
+      h=OperationBase::h*z;
+      
+    if (fabs(l)<iWidth()*z) l=-iWidth()*z;        
+    if (r<iWidth()*z) r=iWidth()*z;    
+    if (h<iHeight()*z) h=iHeight()*z;   
+
+    if (coupled())
+      {
+        auto& iv=*intVar;
+        //            iv.zoomFactor=zoomFactor;
```
highperformancecoder, line 569: dead code

model/operation.cc
```diff
@@ -536,6 +509,194 @@ namespace minsky
...
+        RenderVariable rv(iv,cairo);
+        // we need to add some translation if the variable is bound
+        cairo_rotate(cairo,rotation()*M_PI/180.0);
+        coupledIntTranslation=-0.5*(intVarOffset+2*rv.width()+2+r)*z;
+        if (rv.width()<iv.iWidth()) coupledIntTranslation=-0.5*(intVarOffset+2*iv.iWidth()+2+r)*z;
+        //            cairo_translate(cairo, coupledIntTranslation, 0);
```
highperformancecoder, line 575: dead code

model/operation.cc
```diff
@@ -536,6 +509,194 @@ namespace minsky
...
+    if (textFlipped) swap(y1,y2);
+	
+    // adjust for integration variable
+    if (coupled())
+        x0+=intVarOffset+2*intVarWidth+2;
+	
+    cairo_save(cairo);
+    cairo_identity_matrix(cairo);
+    cairo_translate(cairo, x(), y());
+    //cairo_scale(cairo,zoomFactor,zoomFactor);
```
highperformancecoder, line 661: dead code

model/operation.cc
```diff
@@ -817,104 +977,121 @@ namespace minsky
...
     cairo_show_text(cairo,"\xE2\x88\x9a");
     cairo_set_line_width(cairo,0.5);
     cairo_rel_move_to(cairo,0,-9);
     cairo_rel_line_to(cairo,5,0);
     cairo_set_source_rgb(cairo,0,0,0);
-    cairo_stroke(cairo);
-    cairo_restore(cairo);
+    cairo_stroke(cairo);    
+    cairo_restore(cairo);  
     //    cairo_show_text(cairo,"sqrt");
```
highperformancecoder, line 1088: dead code

model/operation.cc
```diff
@@ -1080,48 +1290,62 @@ namespace minsky
...
   {
+	double sf = scaleFactor(); 	     
+    cairo_scale(cairo,sf,sf);	  
     cairo_set_font_size(cairo,9);
     cairo_move_to(cairo,-6,3);
     cairo_show_text(cairo,"|x|");
   }
   template <> void Operation<OperationType::floor>::iconDraw(cairo_t* cairo) const
   {
 //    cairo_set_font_size(cairo,8);
```
highperformancecoder, line 1328: dead code

highperformancecoder:
Hi Wynand, what do you want to do with this PR? It still bothers me that a couple of tests are not working in a way that we understand. Plus some minor nits. It will need syncing with the latest head too. I can fix these up in a manual merge of course, but maybe you still want to keep working on this.

wdednam:
It is a tough call. I put quite a bit of effort into it, but at the same time I understand that if it is not up to standard, it can't be merged. I believe the tests fail because the items are so small that clickType is "onResize" more often than it is "onItem". So, even if I got the coupled integral and default item, the textnote, to resize properly, it would still not be aesthetically pleasing.

I thought the alternative route Steve suggested, disabling zoom for individual objects on the canvas, would be easier, but I could not get the items to zoom separately from each other, even though I thought I was modifying their zoom factors individually. For it to work, I think I will have to use methods from the group class and a group zoom factor, which will probably also come in handy for the paginate and print feature. At present, though, I find the task daunting because of my repeated failures on this feature PR. If I can get the parameter and variable tabs to look good and work nicely, I will feel better and give it another go. So, I will understand completely if you want to close this PR.

highperformancecoder:
I wasn't suggesting that. There's some good stuff here, and it's almost ready for merging. I just wanted to know if you intended to keep working on it. In that case, I will manually merge it, and fix up those remaining things, but next week. It is beer o'clock now.

wdednam:
In terms of working on it, I can make the resize area on the vars and ops smaller to limit the problem I mentioned in my previous comment. I'll do so over the weekend. I've got an itch to work on the par and var tabs now, so I'm going to make the most of it. :)

wdednam:
BTW, enjoy beer time!

PR #175, opened 2020-04-20
title now appears near top of group, not sure margins are correct, however. several tests fail too, which I will fix, but would like your impression.

wdednam:
This change is [Reviewable](https://reviewable.io/reviews/highperformancecoder/minsky/175)

### DIFF
--- a/model/group.cc
+++ b/model/group.cc
@@ -400,16 +400,22 @@ namespace minsky
   
   Group::IORegion::type Group::inIORegion(float x, float y) const
   {
-    float left, right, z=zoomFactor();
-    margins(left,right);
+    float left, right, top, bottom, z=zoomFactor();
+    margins(left,right,top,bottom);
     float dx=(x-this->x())*cos(rotation()*M_PI/180)-
       (y-this->y())*sin(rotation()*M_PI/180);
-    float w=0.5*iconWidth*z;
+    float dy=(x-this->x())*sin(rotation()*M_PI/180)+
+      (y-this->y())*cos(rotation()*M_PI/180);      
+    float w=0.5*iconWidth*z,h=0.5*iconHeight*z;
     if (w-right*edgeScale()<dx)
       return IORegion::output;
     else if (-w+left*edgeScale()>dx)
       return IORegion::input;
-    else
+    else if (h-bottom*edgeScale()<dy)
+      return IORegion::footer;
+    else if (-h+top*edgeScale()>dy)
+      return IORegion::header;
+    else     
       return IORegion::none;
   }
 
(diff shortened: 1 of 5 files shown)


model/group.cc
```diff
@@ -478,12 +484,12 @@ namespace minsky
...
     // account for margins
-    float l, r;
-    margins(l,r);
+    float l, r, t, bm;
+    margins(l,r,t,bm);
     // rescale contents to fit
     double x0, x1, y0, y1;
     contentBounds(x0,y0,x1,y1);
-    double sx=(fabs(b.x0-b.x1)-z*(l+r))/(x1-x0), sy=fabs(b.y0-b.y1)/(y1-y0);
+    double sx=(fabs(b.x0-b.x1)-z*(l+r))/(x1-x0), sy=(fabs(b.y0-b.y1)-z*(t+bm))/(y1-y0);
```
highperformancecoder, line 492: I don't think either the horizontal or vertical computations correctly handles the case where b is smaller than the minimum size z*(l+r) and z*(t+bm) respectively.

model/group.h
```diff
@@ -219,7 +219,7 @@ namespace minsky
     void moveContents(Group& source);     
 
     /// returns which I/O region (x,y) is in if any
-    struct IORegion {enum type {none,input,output};};
+    struct IORegion {enum type {none,input,output,header,footer};};
```
highperformancecoder, line 222: Why are you adding this here? It's not wrong, I just don't understand the motivation.

model/group.cc
```diff
@@ -966,9 +973,9 @@ namespace minsky
   }
 
 
-  void Group::margins(float& left, float& right) const
+  void Group::margins(float& left, float& right, float& top, float& bottom) const
   {
-    left=right=10;
+    left=right=top=bottom=10;
```
highperformancecoder, line 978: Top and bottom margins are constant, and don't need to be computed. Provided this doesn't require adding additional calls to margins(), which doesn't seem to have happened, this is OK.

highperformancecoder:
For some reason, your pull requests are  not being run through Travis automatically. I'll have to do a check myself before merging. Ho hum.

model/group.cc
```diff
@@ -961,14 +969,32 @@ namespace minsky
...
+    cairo_fill(cairo);    
+    
+    // draw bottom margin. for feature 88
+    cairo_move_to(cairo,-w,h);
+    cairo_line_to(cairo,-w,h+bottom);
+    cairo_line_to(cairo,w,h+bottom);
+    cairo_line_to(cairo,w,h);
+    cairo_line_to(cairo,-w,h);
+    cairo_close_path(cairo);
+    cairo_fill(cairo);
```
highperformancecoder, line 989: For such a simple shape, cairo_rectangle works well...

highperformancecoder:
1. Not passing regression suite
2. Width of left and right margins not adjusting to the I/O variables. I'll add a png to the ticket.
3. Title should appear on the top margin (like a title bar), not in the interior part

wdednam:
The feature now works and I also fixed the tests to take into account the top and bottom margins. The problems affecting groups:

1. Left and Righ margins need to adjuist to the size of the I/O variables
2. The interior is not correctky filling the space in the margins.

I will work on them as part of the fixes required for ticket 1082.

PR #177, opened 2020-04-28
Bugfix 1177

wdednam:
first pass at ticket 1177. variable attached to Godley table does not get created and some tests fail as a result. Still a WIP

---
This change is [Reviewable](https://reviewable.io/reviews/highperformancecoder/minsky/177)


engine/flowCoef.cc
```diff
@@ -21,16 +21,40 @@
...
   {
       const char* f=formula.c_str();
       char* tail;
+      double fnum;      
+      
       // attempt to read leading numerical value
-      coef=strtod(f,&tail);
+      if (parseLine(f, fnum)) {  	
+	       const char* ff=(std::to_string(fnum)).c_str();  
+           coef=strtod(ff,&tail);
```
highperformancecoder, line 54: Why wouldn't you use fnum directly?

highperformancecoder:
I took the code for a spin. I don't think it is working:

x => x correct
1x => 1 
Infy => inf

engine/flowCoef.cc
wdednam: Thanks. that saves a line of code and resources too :)

wdednam:
I'm looking more carefully at the boost::spirit:qi library because it looks to be very promising in terms of parsing. If I can build enough expertise we might be able to use it for all our parsing needs:

[https://www.boost.org/doc/libs/1_72_0/libs/spirit/doc/html/spirit/qi/tutorials/quick_start.html](https://www.boost.org/doc/libs/1_72_0/libs/spirit/doc/html/spirit/qi/tutorials/quick_start.html)

engine/flowCoef.cc
```diff
@@ -21,16 +21,46 @@
...
+{
+  bool parseLine(std::string const& line, double& num) {
+      using It = std::string::const_iterator;
+      
+      // Construct grammar to skip in input formula
+      qi::rule<It, std::string()> nan = -qi::lit("1.0#") >> qi::no_case["nan"] >> -('(' >> *(qi::char_ - ')') >> ')');
+      qi::rule<It, std::string()> inf = qi::no_case[qi::lit("inf") >> -qi::lit("inity")];                                          
+      qi::rule<It, std::string()> sign =   qi::lit('+') | '-';
+      qi::rule<It, std::string()> skip = qi::char_( '/' ) | qi::char_( '\\' ) | qi::char_( '_' );
+
```
highperformancecoder, line 40: Yikes!!!

highperformancecoder: I think maybe your idea of just testing for nan or inf at start, maybe the way to go. Alternatively, test for digit or decimal point as the first character and run strtod, otherwise just treat as string.

wdednam:
I think this is finally working now. The only tests that fail on my machine are:

allItemsRenderCheck.sh
godleyTableWindow.sh
oldSchemas.sh

They invariably fail on my machine, however, even when a Travis build passes online. The problem with this PR is that I converted it to  a draft. I don't get Travis build reports in my inbox for draft PRs, so I won't know whether it passes or not.

highperformancecoder:
Re regression tests failing on your machine, this is probably due to minor changes to rendered images, which happen with different versions of librsvg, libpng and so on. You can disable the rendered image checks by setting the TRAVIS environment variable to 1, which is (surprise surprise) what Travis does. I will always do a visible check of reference images where the check fails on my NUC, and update the reference images if necessary prior to doing a full release.  Sometimes the process picks up genuine bugs, more often than not, it is due to a version change in those image libs.

PR #178, opened 2020-04-29
towards downloading and opening zipped csv files. just have to figure…

wdednam:
… out 'unknown escape character' error and we should be good to go. I have been testing files at https://www.stats.govt.nz/large-datasets/csv-files-for-download/

---
This change is [Reviewable](https://reviewable.io/reviews/highperformancecoder/minsky/178)


model/CSVDialog.cc
```diff
@@ -138,7 +140,25 @@ std::string CSVDialog::loadWebFile(const std::string& url)
   const std::string tempStr    = temp.string();
           
   std::ofstream outFile(tempStr, std::ofstream::out);  
-  outFile << res.get().body();                                            
+ 
+  // Deal with zipped csv files 
+  std::string targetStr = target.str();
+  std::string fileExt = ".zip";
+  std::size_t found = targetStr.find(fileExt);
```
highperformancecoder, line 147: Can we write the code in such a way as to not assume a specific extension? Someone may want to use a .gz for example, or (less likely) a .z or .Z, which are obsolete zip formats that zlib handles.

model/CSVDialog.cc
```diff
@@ -138,7 +140,25 @@ std::string CSVDialog::loadWebFile(const std::string& url)
...
+  if (found!=std::string::npos) {
+      vector<unsigned char> zbuf(res.get().body().size());
+      DeflateZStream zs(res.get().body(), zbuf);
+      zs.deflate();  
+      
+      vector<char> cbuf(a85::size_for_a85(zs.total_out,false));
+      a85::to_a85(&zbuf[0],zs.total_out, &cbuf[0], false);
+      // this ensures that the escape sequence ']]>' never appears in the data
+      replace(cbuf.begin(),cbuf.end(),']','~');      
+
```
highperformancecoder, line 158: What is this code doing in CSVDialog? a85 is about ASCII representation of binary data, which is not required for handling zipped CSV files.

model/CSVDialog.cc
```diff
@@ -149,14 +148,13 @@ std::string CSVDialog::loadWebFile(const std::string& url)
...
-      vector<char> cbuf(a85::size_for_a85(zs.total_out,false));
-      a85::to_a85(&zbuf[0],zs.total_out, &cbuf[0], false);
-      // this ensures that the escape sequence ']]>' never appears in the data
-      replace(cbuf.begin(),cbuf.end(),']','~');      
+      //vector<char> cbuf(zs.total_out);
+      //memcpy(&cbuf,&zbuf,zs.total_out);   
+      //memcpy(&cbuf[0],&zbuf[0],zbuf.size()*sizeof(zbuf[0]));      
     
-      outFile << cbuf.data();
+      outFile << zbuf.data();
```
highperformancecoder, line 157: This will not work. Passing a char* to an output stream assumes that nul ('\0') terminates the data.
You will need to use the binary API of std::ostream, which I can't tell you off hand, you'll need to look it up.

model/CSVDialog.cc
```diff
@@ -149,14 +148,13 @@ std::string CSVDialog::loadWebFile(const std::string& url)
   if (found!=std::string::npos) {
       vector<unsigned char> zbuf(res.get().body().size());
```
highperformancecoder, line 149: This zbuf will be too small. The point of compressing data is to make it smaller, and you are sizing the receive buffer to be the compressed size. You should use the zs.total_out value computed after zs.deflate.

model/CSVDialog.cc
wdednam: Oops, that was an obvious mistake, sorry. I first tried a memcpy(), but of course a container of size z.total_out will not be compatible with one of res.get().body(), so I got memory leaks. I thought I may have done something silly which could be fixed in a single memcpy() statement, say. But I will work through https://www.zlib.net/zlib_how.html and http://www.cplusplus.com/reference/ostream/ostream/ carefully. Thank you for the hint.

model/CSVDialog.cc
wdednam: Yes, I tried to do that in the lines following it, but I may have to assign a CHUNK of memory exceeding res.get.body().size() as discussed at https://www.zlib.net/zlib_how.html and then correct for it using zs.total_out, if I understand correctly:


            have = CHUNK - strm.avail_out;
            if (fwrite(out, 1, have, dest) != have || ferror(dest)) {
                (void)deflateEnd(&strm);
                return Z_ERRNO;
            }

highperformancecoder: Yes - you have to do that crap when dealing with the C API. The point of my ZStream class is to handle that detail for you. It automatically resizes the buffer as it reads in the compressed data.

wdednam:
I have unfortunately wasted another half-day on this, and made little progress except for making the original CSV import code a little shorter and more robust (I now use the more general "dynamic_body" as opposed to "string_body" response anad request parsers, but neither work for unzipping). I've also included some dead code of an alternative zlib inflate stream implementation, adapted from [https://pastebin.com/DiUME7Z1](https://pastebin.com/DiUME7Z1). It doesn't work either. I also tried boost::iostreams again, but that didn't work... I don't see how it can be so difficult to download and extract a zipped CSV file?? I'm completely baffled. The CSV import routine works for all the ordinary CVS files I've tested at [https://www.stats.govt.nz/large-datasets/csv-files-for-download/](https://www.stats.govt.nz/large-datasets/csv-files-for-download/).

model/CSVDialog.cc
```diff
@@ -123,22 +173,35 @@ std::string CSVDialog::loadWebFile(const std::string& url)
...
           
-  std::ofstream outFile(tempStr, std::ofstream::out);  
-  outFile << res.get().body();                                            
+  std::ofstream outFile(tempStr, std::ofstream::binary);  
+  
+  // Handle zipped CSV files. Not working... 
+  if (res[http::field::content_type]=="application/zip") {
+	   
+	  
+	   std::string responseBody = boost::beast::buffers_to_string(res.body().data());
```
highperformancecoder, line 194: Check what this does. Usually, strings in C++ are nul-terminated, and this conversion may be corrupting the buffer. If res.body().data() returns a simple const char*, then that is what you should use.

wdednam: res.body().data() returns a boost::beast::basic_multi_buffer which is not convertible to string or char by standard casting. [https://stackoverflow.com/questions/45761865/how-to-convert-boost-beast-multi-buffer-to-string](https://stackoverflow.com/questions/45761865/how-to-convert-boost-beast-multi-buffer-to-string)

Another half-day down the drain. Most of my attempts to add value to this Ravel feature have come to naught. :( 

I've searched high and low and tried several suggestions:

[https://html.developreference.com/article/13439501/Trying+to+understand+the+boost%3A%3Abeast+multibuffer](https://html.developreference.com/article/13439501/Trying+to+understand+the+boost%3A%3Abeast+multibuffer)

and

[https://github.com/boostorg/beast/issues/936](
https://github.com/boostorg/beast/issues/936)

I will have to abandon this endeavour and just implement the fix for ticket 1173....

wdednam: I tried to import the simplest possible zipped csv file at [https://www.stats.govt.nz/large-datasets/csv-files-for-download/](https://www.stats.govt.nz/large-datasets/csv-files-for-download/) :

[https://www.stats.govt.nz/assets/Uploads/Employment-indicators-October-2019-infoshare-tables/monthly-employment-indicators-october-2019.zip](
https://www.stats.govt.nz/assets/Uploads/Employment-indicators-October-2019-infoshare-tables/monthly-employment-indicators-october-2019.zip)

No luck even in this case...

highperformancecoder:
libz handles gzip files - in fact gunzip is just a program wrapped around libz. Can you correctly uncompress your sample file using gunzip?

Also, until now, these routines have only been used to uncompress data that's been compressed by the deflate routine. If .zip files have a slightly different structure, that could explain the problem.

wdednam:
Thanks for your feedback. I'm not sure the inflateZStream routine in zStream.h is the problem, because I tried at least two other zlib implementations I found online. I made a serious attempt to make the following one work, but got a completely inexplicable error ("<< operator not defined for Pc" for which google searches yielded nothing):

[https://pastebin.com/DiUME7Z1](
https://pastebin.com/DiUME7Z1)

As for the zip format,

(inflateInit2(this,-MAX_WBITS)!=Z_OK) throwError();  //none of MAX_WBITS + 16, MAX_WBITS or -MAX_WBITS works. see https://stackoverflow.com/questions/1838699/how-can-i-decompress-a-gzip-stream-with-zlib

is supposed to take care of that. I "know" that -MAX_WBITS is the correct one in this case because the others give me an "Incorrect Header" error, which is where the zip, gzip and other formats differ as far as I have read and understand.

There's some detail I'm missing which will probably only be resolved by asking on the Beast forum. I will do that, but using the implementation at [https://pastebin.com/DiUME7Z1](
https://pastebin.com/DiUME7Z1) instead of InflateZStream because Vinnie Falco knows about the latter implementation and also that it works (the issue affecting the sample code was marked as resolved by Vinnie Falco in a forum post).

Today, I will work on fixing ticket 1173 and finalising feature 94 by adding a separate draw function for the IntOp. I'm busy implementing your comments on ticket 1080-1183 now.

gui-tk/csvImport.tcl
```diff
@@ -138,7 +138,7 @@ proc csvImportDialogOK {url} {
         cancelWin .wiring.csvImport
         # Delete temporary file created by web import of CSV data
         if [string length $url] {
-	        file delete $filename      
+	        csvDialog.deleteFile $filename
```
highperformancecoder, line 141: This was not what I had in mind. There is no benefit to providing a C++ implementation of a file deleter, when TCL already provides one.

What I had in mind was that you provide a variant of MinskyTCL::loadVariableFromCSV that takes a URL, and manages any temporary file it might create within it.

You can probably even do away with storing anything to disk, as boost:beast probably provides a streaming interface.

wdednam: Sorry, I misunderstood. I thought that you didn't want a file delete method in TCL because it was more unreliable than C++, an impression that was reinforced by my own experience that files would sometimes not be deleted during web import. I will implement it the way you suggest, thanks.

PR #179, opened 2020-05-07
Following the example of GroupItems, I try to extract all the members related to the variables attached to a GodleyIcon for the copying which is no longer the default approach (see line 36 on group.h).. 

wdednam:
This change is [Reviewable](https://reviewable.io/reviews/highperformancecoder/minsky/179)


model/godleyIcon.h
```diff
@@ -33,15 +33,70 @@
...
+  };	  
 
-  class GodleyIcon: public ItemT<GodleyIcon>
+  class GodleyIcon: public GodleyVars
   {
-    /// for placement of bank icon within complex
-    float flowMargin=0, stockMargin=0, iconWidth=100, iconHeight=100;
-    /// icon scale is adjusted when Godley icon is resized
-    float m_iconScale=1;
+    friend class GodleyVars;
```
highperformancecoder, line 98: Why is this needed? Any state that needs to be shared should be declared in a protected section of GodleyVars.

model/godleyIcon.h
```diff
@@ -33,15 +33,70 @@
...
+    void setCurrency(const std::string& currency) 
+    {for (auto& i: m_stockVars) i->setUnits(currency);}
+    
+    /// performs dimensional analysis on stock var column \a stockName
+    /// @param check indicates whether a consistency check is applied
+    Units stockVarUnits(const std::string& stockName, bool check) const;
+      
+    void insertControlled(Selection& selection) override;    
+    
+  private:
```
highperformancecoder, line 85: Probably better as a protected section.

model/godleyIcon.h
```diff
@@ -33,15 +33,70 @@
...
+    
+    bool variableDisplay=true;
+    void toggleVariableDisplay() {variableDisplay=!variableDisplay;}   
+    
+    void removeControlledItems() const override;    
+    
+    /// flows, along with multipliers, appearing in \a col
+    std::map<string,double> flowSignature(int col) const;    
+    
+    GodleyTable table;
```
highperformancecoder, line 65: This should be in GodleyIcon

model/godleyIcon.h
```diff
@@ -33,15 +33,70 @@
...
+    GodleyVars(const GodleyVars& x) {};
+    GodleyVars& operator=(const GodleyVars&) {return *this;}
+    classdesc::Exclude<std::weak_ptr<GodleyIcon>> self; ///< weak ref to this    
+    
+    void clear() {
+      m_flowVars.clear();
+      m_stockVars.clear();
+    }  
+    
+    bool variableDisplay=true;
```
highperformancecoder, line 57: This should be in GodleyIcon

model/godleyIcon.h
```diff
@@ -33,15 +33,70 @@
 namespace minsky
 {
   class GodleyTableEditor;
+  
+  // Godley variables broken out in a separate structure, as copying is non-default
+  class GodleyVars: public ItemT<GodleyVars>
```
highperformancecoder, line 38: GodleyVars doesn't need to inherit from ItemT. Rather, use multiple inheritance in GodleyIcon.

model/godleyIcon.h
```diff
@@ -33,15 +33,70 @@
...
+                    const vector<string>& varNames, 
+                    VariableBase::Type varType);
+    /// move contained variables to correct locations within icon
+    void positionVariables() const;
+    Variables m_flowVars, m_stockVars;  
+    CLASSDESC_ACCESS(GodleyVars);
+  };	  
 
-  class GodleyIcon: public ItemT<GodleyIcon>
+  class GodleyIcon: public GodleyVars
```
highperformancecoder, line 96: Should also inherit from ItemT<GodleyIcon>

model/godleyIcon.h
```diff
@@ -33,15 +33,70 @@
...
+  public:	    
+    typedef std::vector<VariablePtr> Variables;
+    const Variables& flowVars() const {return m_flowVars;}
+    const Variables& stockVars() const {return m_stockVars;}    	  
+	  
+    GodleyVars() {}
+    virtual ~GodleyVars() {}
+    // copy operations not deleted to allow ItemT<Group> to compile
+    GodleyVars(const GodleyVars& x) {};
+    GodleyVars& operator=(const GodleyVars&) {return *this;}
```
highperformancecoder, line 49: If the only members are the two variable vectors, then the copy operations should be just a simple call to update()

model/godleyIcon.h
```diff
@@ -33,15 +33,70 @@
...
+    
+    void removeControlledItems() const override;    
+    
+    /// flows, along with multipliers, appearing in \a col
+    std::map<string,double> flowSignature(int col) const;    
+    
+    GodleyTable table;
+    
+    /// updates the variable lists with the Godley table
+    void update();
```
highperformancecoder, line 68: Make this a virtual function, as it needs to be implemented in GodleyIcon, but callable in this class to implement the copy operations.

model/godleyIcon.h
```diff
@@ -33,15 +33,70 @@
...
+    GodleyVars& operator=(const GodleyVars&) {return *this;}
+    classdesc::Exclude<std::weak_ptr<GodleyIcon>> self; ///< weak ref to this    
+    
+    void clear() {
+      m_flowVars.clear();
+      m_stockVars.clear();
+    }  
+    
+    bool variableDisplay=true;
+    void toggleVariableDisplay() {variableDisplay=!variableDisplay;}
```
highperformancecoder, line 58: Ditto

model/godleyIcon.h
```diff
@@ -33,15 +33,70 @@
...
+      m_stockVars.clear();
+    }  
+    
+    bool variableDisplay=true;
+    void toggleVariableDisplay() {variableDisplay=!variableDisplay;}   
+    
+    void removeControlledItems() const override;    
+    
+    /// flows, along with multipliers, appearing in \a col
+    std::map<string,double> flowSignature(int col) const;
```
highperformancecoder, line 63: I'd keep it simple, and leave all of these methods in GodleyIcon where they existed originally.

model/godleyIcon.cc
```diff
@@ -98,14 +98,14 @@ namespace minsky
     set<string> alreadyAdded;
 
     vars.clear();
-    shared_ptr<GodleyIcon> self;
-    if (auto g=group.lock())
-      self=dynamic_pointer_cast<GodleyIcon>(g->findItem(*this));
+    shared_ptr<GodleyIcon> godleySelf;
+    if (auto g=self->group.lock())
+      godleySelf=dynamic_pointer_cast<GodleyIcon>(g->findItem(*this->self));
```
highperformancecoder, line 103: It would be simpler to leave these methods in GodleyIcon, which has a full view of everything.

highperformancecoder:
>   compiles now after feedback, thank you. I have a memory leak though because of classdesc shared pointer in GodleyVars, hence Minsky crashes with segfault on creation of a GodleyIcon. How can I use classdesc weak_ptr instead of shared_ptr to refer to GodleyIcon members from GodleyVars members in godleyIcon.cc?

This is why you should leave those methods in GodleyIcon. All you need to do is make update virtual, so that the copy operations can call GodleyIcon::update().

model/godleyIcon.h
```diff
@@ -33,16 +34,43 @@
...
-    friend class SchemaHelper;
+  public:	    
+    typedef std::vector<VariablePtr> Variables;
+    const Variables& flowVars() const {return m_flowVars;}
+    const Variables& stockVars() const {return m_stockVars;}    	  
+	  
+    GodleyVars() {}
+    virtual ~GodleyVars() {}
+    // copy operations not deleted to allow ItemT<GodleyIcon> to compile
+    GodleyVars(const GodleyVars& x) {const_cast<GodleyVars&>(x).update();}
```
highperformancecoder, line 49: No - you shouldn't be updating x, but updating this.

model/godleyIcon.h
```diff
@@ -33,16 +34,43 @@
...
+    //classdesc::Exclude<std::weak_ptr<GodleyIcon>> self; ///< weak ref to this    
+    classdesc::Exclude<std::shared_ptr<GodleyIcon>> self;
+    
+    void clear() {
+      m_flowVars.clear();
+      m_stockVars.clear();
+    }    
+
+    /// updates the variable lists with the Godley table
+    virtual void update();
```
highperformancecoder, line 60: Make this method pure virtual (ie =0), and provide the implementation as GodleyIcon::update

model/godleyIcon.h
```diff
@@ -33,8 +33,24 @@
...
+  
+  // Godley variables broken out in a separate structure, as copying is non-default
+  struct GodleyVars
+  {
+  public:	    
+	  
+    GodleyVars() {}
+    virtual ~GodleyVars() {}
+    // copy operations not deleted to allow ItemT<GodleyIcon> to compile
+    GodleyVars(const GodleyVars& x) {} //{const_cast<GodleyVars&>(x).update();}
```
highperformancecoder, line 45: This should all update() as well.

highperformancecoder:
This change is looking good. Just need to fix up the copy constructor. Does it now solve the original problem?

model/godleyIcon.h
wdednam: I stiil don't understand why I need to call *this->update() or just update() in the copy constructor and not x.update()?

When I write     GodleyVars(const GodleyVars& x) {update();}

I get a compile warning:

model/godleyIcon.h: In copy constructor ‘minsky::GodleyVars::GodleyVars(const minsky::GodleyVars&)’:
model/godleyIcon.h:45:45: warning: pure virtual ‘virtual void minsky::GodleyVars::update()’ called from constructor
     GodleyVars(const GodleyVars& x) {update();}

and finally the build fails at the linking stage:

godleyIcon.o: In function `minsky::GodleyVars::GodleyVars(minsky::GodleyVars const&)':
/home/wd/minsky/model/godleyIcon.h:45: undefined reference to `minsky::GodleyVars::update()'
collect2: error: ld returned 1 exit status
Makefile:162: recipe for target 'gui-tk/minsky' failed
make: *** [gui-tk/minsky] Error 1

If you like, we can discuss it in our meeting at between 7 and 7.30 pm your time?

highperformancecoder:
On Mon, May 11, 2020 at 01:05:20AM -0700, Wynand wrote:
> 
> When I write GodleyVars(const GodleyVars& x) {update();}
> 
> I get a compile warning:
> 
> model/godleyIcon.h: In copy constructor ‘minsky::GodleyVars::GodleyVars(const
> minsky::GodleyVars&)’:
> model/godleyIcon.h:45:45: warning: pure virtual ‘virtual void
> minsky::GodleyVars::update()’ called from constructor
> GodleyVars(const GodleyVars& x) {update();}

Ah - yes, you cannot call update() at this point, because at this
point GodleyIcon has not been constructed.

Actually, even though it's not a compile error, calling update() in
the assignment operator will similarly be wrong, as it will be called
before the data vector in table is copied. Dang!

We will need to leave both copy operations blank (null ops), and add
the call to update in the overridden clone() method. Once the full
GodleyIcon object has been copy constructed, one can call update and
get what we want.

> 
> If you like, we can discuss it in our meeting at between 7 and 7.30 pm your
> time?

OK.

> 
> —
> You are receiving this because you commented.
> 

-- 

                      http://www.hpcoders.com.au

wdednam:
> We will need to leave both copy operations blank (null ops), and add
the call to update in the overridden clone() method. Once the full
GodleyIcon object has been copy constructed, one can call update and
get what we want.

I've added an overridden GodleyIcon::clone() method which calls update() after group.reset() and tested it, but I've probably set it up incorrectly because it still reproduces the twice-copied vars attached to the GodleyIcon after copying:

    GodleyIcon* clone() const override {
      auto r=new GodleyIcon(*this);
      r->group.reset();
      r->update();
      return r;
    }   

Chat to you in a bit.

wdednam:
So, after stepping through the code in the debugger, I've traced the bug to group.cc on lines 242-243:

     if (auto g=dynamic_cast<GodleyIcon*>(it.get()))
       g->update(); // handle scoping of contained variables

Commenting out these lines fixes the bug, but of course they exist to handle the scoping of the variables contained in group. Also, the bug only affects the stock variables. So, this brings us back to how the groups are handled in updateVars. I still haven't managed to find a null group when stepping through updateVars, but I will continue to look.

wdednam:
Hmmm. Those two lines may now be unnecessary because I use the new godleyIcon::clone() method in canvas::copyItem. I have tested this by putting the GodleyIcon from 1Free.mky inside a group, zoomed in until it became transparent, reselected and then copied the group back onto the Canvas. The scoping looks correct to me (see the attached figure), but I'd rather you take a look, so I'll commit the code after running the test suite.
scoping_of_Godley_vars_inside_groups

model/canvas.cc
```diff
@@ -640,7 +640,10 @@ namespace minsky
           newItem=group->copy();
         else
           {
-            newItem.reset(item->clone());
+			//  Use GodleyIcon's native clone() method. for tickets 1180/1183  
+            if (auto godley=dynamic_cast<GodleyIcon*>(item.get()))
+              newItem.reset(godley->clone());
```
highperformancecoder, line 645: Because clone() is a virtual method, these lines are strictly not needed (and indeed, add a small, probably impercetible, delay due to the unnecessary dynamic_cast.

model/godleyIcon.h
```diff
@@ -33,8 +33,22 @@
...
+  struct GodleyVars
+  {
+  public:	      
+    GodleyVars() {}
+    virtual ~GodleyVars() {}
+    // copy operations not deleted to allow ItemT<GodleyIcon> to compile
+    GodleyVars(const GodleyVars& x) {}
+    GodleyVars& operator=(const GodleyVars&) {return *this;}
+
+    virtual void update() = 0;
```
highperformancecoder, line 47: I guess this doesn't need to be there now.

model/godleyIcon.h
```diff
@@ -111,7 +125,14 @@ namespace minsky
     const Variables& stockVars() const {return m_stockVars;}
     GodleyTable table;
     /// updates the variable lists with the Godley table
-    void update();
+    void update() override;
```
highperformancecoder, line 128: Obviously, when you remove the virtual base class method, you'll need to remove the override.

model/group.cc
```diff
@@ -411,7 +409,7 @@ namespace minsky
       return IORegion::output;
     else if (-w+left*edgeScale()>dx)
       return IORegion::input;
-    else if (-h+10*z>dy || h-10*z<dy)     
+    else if ((fabs(-h-10*z)>fabs(dy) && dy<0) || (h+10*z>dy && dy>0))
```
highperformancecoder, line 412: I'm suspicious of this line. The two clauses aren't symmetric.

model/canvas.cc
wdednam: I tried to remove them, but then the wires would not be copied along with the Godley icons in the example 1Free.mky.

model/godleyIcon.h
wdednam: Still needed, otherwise the wheels come off worse than before during a copy.

model/godleyIcon.h
wdednam: Kept it for the above reasons.

model/group.cc
wdednam: The only thing I could fix. Have been working hard on feature 94 and made some progress. I hope to finish it by tomorrow but am not sure I'll get to ticket 1173 before Thursday, but I'll try :)

model/canvas.cc
highperformancecoder: Something is wrong, because this entire conditional is strictly equivalent to newItem.reset(item->clone());

model/godleyIcon.h
highperformancecoder: There is no longer any reason for update() to be called in a GodleyVars context.

model/canvas.cc
wdednam: I removed group.reset() from godleyIcon::clone(), which allowed me to remove the GodleyVars struct, but as soon as I remove newItem.reset(godley->clone()), the wires are no longer copied along with the Godley icon in 1Free.mky. I don't what else to do. I worked 9.5 hours on feature 94 yesterday so that may have sapped all my mental faculties when I looked at this again last night. I will try to look at this again after I get feature 94 working today.

model/godleyIcon.h
wdednam: I have managed to fix this.

model/godleyIcon.h
```diff
@@ -112,6 +112,12 @@ namespace minsky
     GodleyTable table;
     /// updates the variable lists with the Godley table
     void update();
+    
+    GodleyIcon* clone() const override {
+      auto r=new GodleyIcon(*this);
+      r->update();
+      return r;
```
highperformancecoder, line 119: The problem with this approach is that it leaves the new cloned GodleyIcon purportedly a member of the original group, but the controller items of the contained variables will be null, because the cloned GodleyIcon is not actually a member of the group.

Because of this, it may prove better, or more consistent to create a dummy group for the cloned Godley icon to live in while the update takes place.

wdednam: Thank you, that was a very useful suggestion. What I've done seems to work, but also looks a little hacky to me, i.e., I feel there may be a better way to write the bit of code I've added...

model/godleyIcon.h
```diff
@@ -115,7 +116,23 @@ namespace minsky
...
     GodleyIcon* clone() const override {
       auto r=new GodleyIcon(*this);
-      r->update();
+      // create a dummy group for the cloned Godley icon to live in while the update takes place. for tickets 1180-1183. 
+      if (auto g=r->group.lock())
+      {
+	   GroupPtr dummy(new Group);
+       for (auto& i: g->items)
+       {		
+          dummy->addItem(i);
```
highperformancecoder, line 125: This removes i from g, and puts it in dummy.

model/godleyIcon.h
```diff
@@ -115,7 +116,23 @@ namespace minsky
...
+       {		
+          dummy->addItem(i);			  
+          assert(!i->ioVar());
+       }
+	   for (auto& it: dummy->items) 
+	     if (auto godley=dynamic_cast<GodleyIcon*>(it.get())) {
+	        godley->update();
+	        r=godley;
+	     }
+	   dummy->clear();
```
highperformancecoder, line 133: This deletes all of the items in dummy (and hence g).

model/godleyIcon.h
```diff
@@ -115,7 +116,23 @@ namespace minsky
...
+	   GroupPtr dummy(new Group);
+       for (auto& i: g->items)
+       {		
+          dummy->addItem(i);			  
+          assert(!i->ioVar());
+       }
+	   for (auto& it: dummy->items) 
+	     if (auto godley=dynamic_cast<GodleyIcon*>(it.get())) {
+	        godley->update();
+	        r=godley;
```
highperformancecoder, line 131: This leaks the original object allocated on line 118. Plus, the reference obtained on line 129 is most likely invalidated by the clear() call on line 133.

model/godleyIcon.h
```diff
@@ -115,7 +116,23 @@ namespace minsky
...
+	     }
+	   dummy->clear(); 
+	  }
+	  else r->update();
```
highperformancecoder, line 135: With the above problems, you'd better hope the else branch is always taken.

model/godleyIcon.h
wdednam: Yikes, I thought this would be the only dubious line. I guess a shared pointer would be the way to go? Then I wouldn't have to create a new dummy group, but just share g with it, without deleting g?

wdednam:
There is still a problem when a Godley icon with attached variables and wires is copied by selection. If the selection is cleared by clicking in the canvas before pasting, then the wires are sometimes not copied and the following error appears on the command line:

clipboard get -type UTF8_STRING->CLIPBOARD selection doesn't exist or form "UTF8_STRING" not defined

PR #180, opened 2020-05-08
Added 0, 1 and infinity as constant operators and fixed up tests

wdednam:
The testLatexEq and testMatlabEq tests fail and just by looking at the diffs below I can't see the obvious reasons why. Sorry about that.

---
This change is [Reviewable](https://reviewable.io/reviews/highperformancecoder/minsky/180)


engine/evalOp.cc
```diff
@@ -159,6 +159,36 @@ namespace minsky
...
+  template <> 
+  double EvalOp<OperationType::one>::d1(double x1, double x2) const
+  {return 0;}
+  template <>
+  double EvalOp<OperationType::one>::d2(double x1, double x2) const
+  {return 0;}     
+  
+  template <>
+  double EvalOp<OperationType::inf>::evaluate(double in1, double in2) const
+  {return numeric_limits<double>::max();}
```
highperformancecoder, line 184: Is this the best value to use? Or should we use Inf?

highperformancecoder:
Just fix up the latexTestEq and matlabTestEq tests, and we're good to go.

PR #182, opened 2020-05-21
added percentage and factorial single argument operators. not sure ho…

wdednam:
…w to deal with vector derivative in factorial case

---
This change is [Reviewable](https://reviewable.io/reviews/highperformancecoder/minsky/182)


engine/derivative.cc
```diff
@@ -639,6 +640,31 @@ namespace MathDAG
...
   }
   
+  template <>
+  NodePtr SystemOfEquations::derivative<>
+  (const OperationDAG<OperationType::percent>& expr)
+  {
+    if (expr.arguments[0].empty())
+      return zero;
+    else
+       throw error("percent is not differentiable");
```
highperformancecoder, line 650: % is just divide by 100. So the derivative will be 0.01 x.

engine/derivative.cc
```diff
@@ -639,6 +640,31 @@ namespace MathDAG
...
+//  template <>
+//  NodePtr SystemOfEquations::derivative
+//  (const OperationDAG<OperationType::fact>& expr)
+//  {
+//    if (expr.arguments[0].empty())
+//      return zero;
+//    else
+//      {
+//        //Expr x(expressionCache, expr.arguments[0][0]);
+//        //Expr gamma_p = boost::math::digamma(one+x)*boost::math::tgamma(one+x);
```
highperformancecoder, line 662: gamma function is available in libmath - apparently, the one to use is tgamma(x) or tgammal(x) for the logarithm version.

engine/evalOp.cc
```diff
@@ -488,6 +491,26 @@ namespace minsky
...
+  template <>
+  double EvalOp<OperationType::percent>::d2(double x1, double x2) const
+  {return 0;}
+  
+  template <>
+  double EvalOp<OperationType::fact>::evaluate(double in1, double in2) const
+  {return boost::math::tgamma(in1+1);}
+  template <>
+  double EvalOp<OperationType::fact>::d1(double x1, double x2) const
+  {return boost::math::digamma(x1+1)*boost::math::tgamma(x1+1);}
```
highperformancecoder, line 510: I was going to say use the libm gamma function, but I don't think libm contains a digamma function.

engine/derivative.cc
highperformancecoder, line 662: This is a symbolic differentiator. The only way we can do this is to also add in all the polygamma functions ψ⁽ⁿ⁾(x) = dⁿ/dxⁿψ(x) as well.

Fortunately, boost::math provides implementations of all the polygamma functions.

engine/expr.h
```diff
@@ -156,6 +156,12 @@ namespace MathDAG
     r->arguments[0].push_back(x);
     return Expr(x.cache,r);
   }
+  
+  //inline Expr fact(const Expr& x) {
+  //  shared_ptr<OperationDAGBase> r(x.newNode(OperationType::fact));
+  //  r->arguments[0].push_back(x);
+  //  return Expr(x.cache,r);
+  //}
```
highperformancecoder, line 164: This code is fine. We need it, in case someone attaches the derivative operator to an expression containing a factorial.

engine/node_matlab.cc
```diff
@@ -493,6 +493,20 @@ namespace MathDAG
     checkArg(0,0);
     return o<<"frac("<<arguments[0][0]->matlab()<<")";
   }
+  
+  template <>
+  ostream& OperationDAG<OperationType::percent>::matlab(ostream& o) const
+  {
+    checkArg(0,0);
+    return o<<"("<<arguments[0][0]->matlab()<<")%";
```
highperformancecoder, line 501: I think in matlab, you'd need to write this as 0.01*(arguments[0][0]->matlab())

engine/node_matlab.cc
```diff
@@ -493,6 +493,20 @@ namespace MathDAG
...
+  }  
+  
+  template <>
+  ostream& OperationDAG<OperationType::fact>::matlab(ostream& o) const
+  {
+    checkArg(0,0);
+    return o<<"factorial("<<arguments[0][0]->matlab()<<")";
```
highperformancecoder, line 508: factorial is only defined for integers. I think you need to replace this with a call to gamma(1+arguments[0][0]->matlab())

highperformancecoder:
I think we'll need to add a gamma operation (symbol Γ), digamma (ψ) and polygamma (ψ⁽ⁿ⁾) as well.

engine/derivative.cc
highperformancecoder: Actually, mea culpa. % will be multiple by 100, so the derivative should be 100x.

engine/equationDisplay.cc
```diff
@@ -556,6 +556,54 @@ namespace MathDAG
...
+      Pango pango(s.cairo());
+      double oldFs=pango.getFontSize();
+      pango.setFontSize(r.height());
+      cairo_rel_move_to(s.cairo(),0,-(r.height()-oldFs));
+      pango.show();
+      cairo_rel_move_to(s.cairo(),0,(r.height()-oldFs));
+      cairo_rel_move_to(s.cairo(),pango.width(),0);
+      arguments[0][0]->render(s);
+      xx+=pango.width()+r.width();
+      pango.setMarkup("!");
```
highperformancecoder, line 602: Wouldn't you need to scale the ! symbol by the size of the text to the left?

highperformancecoder:
The other thing the % operator needs to is the multiply the Units field by %. This should prevent someone from adding a % to something that is not a percent. Longer term it should be possible, but with units converted via the "Conversions" concept, ie 1+10% = 1.1, and 10%+1 = 110%. But that would be a different ticket.

engine/derivative.cc
wdednam: I have been confused too. I guess when the % sign appears next to the number, the number must have been multiplied by 100. That helps me to know when to multiply instead divide by 100.

engine/evalOp.cc
wdednam: I didn't see one either, which I was I chose boost. In general, I try to use the std library first, but when a functionality/implementation is unavailable, I go shopping elsewhere.

engine/derivative.cc
wdednam: I'm not sure, but I would only have to go up to the trigamma function at most, wouldn't I? Otherwise we would need something like the following to deal with derivatives of order 4 or higher, or rather, to calculate nth order derivatives from n-1th ones:

[http://erikerlandson.github.io/blog/2016/06/15/computing-derivatives-of-the-gamma-function/](
http://erikerlandson.github.io/blog/2016/06/15/computing-derivatives-of-the-gamma-function/)

engine/expr.h
wdednam: Okay, I have put it back in.

engine/node_matlab.cc
wdednam: Need to fix this to be 100* times now. Will do so just now.

engine/node_matlab.cc
wdednam: Did this.

engine/equationDisplay.cc
wdednam: That's true, thanks. I will fix that now.

wdednam: Do you mean I should use recording surface s instead of r for the size of the ! and % symbols?

engine/derivative.cc
```diff
@@ -639,6 +640,34 @@ namespace MathDAG
...
+  template <>
+  NodePtr SystemOfEquations::derivative<>
+  (const OperationDAG<OperationType::percent>& expr)
+  {
+    if (expr.arguments[0].empty())
+      return zero;
+    else
+      {
+        Expr x(expressionCache, expr.arguments[0][0]);
+        return chainRule(x, 100*x/x);
```
highperformancecoder, line 652: The derivative of % is just 100x'

No need to invoke the chain rule, you can write this as 100*x->derivative(*this);

engine/derivative.cc
```diff
@@ -647,6 +676,8 @@ namespace MathDAG
     throw error("vector derivatives not implemented");  \
   }                                                     
 
+  VECTOR_DERIVATIVE_NOT_IMPLEMENTED(gamma)
+  VECTOR_DERIVATIVE_NOT_IMPLEMENTED(digamma)
   VECTOR_DERIVATIVE_NOT_IMPLEMENTED(sum)
```
highperformancecoder, line 681: These two shouldn't be here. But you will need to implement the recursive polygamma function.

engine/node_latex.cc
```diff
@@ -529,6 +529,34 @@ namespace MathDAG
...
+  {
+    checkArg(0,0);
+    return o<<"\\left"<<arguments[0][0]->latex()<<"\\right\\%";
+  }
+  
+  template <>
+  ostream& OperationDAG<OperationType::gamma>::latex(ostream& o) const
+  {
+    checkArg(0,0);
+    return o<<"\\gamma\\left("<<arguments[0][0]->latex()<<"\\right)";
```
highperformancecoder, line 544: Should be uppercase gamma (ie \Gamma).

model/operation.cc
```diff
@@ -435,9 +435,32 @@ namespace minsky
...
       case function: case reduction: case scan: case tensor:
-        if (check && !ports[1]->units(check).empty())
-          throw_error("function input not dimensionless");
-        return {};
+        switch (type())
+          {
+		  case percent:
+		    {
+             if (check && !ports[1]->units(check).empty())
+               throw_error("function input not dimensionless");
```
highperformancecoder, line 443: For percent, we can have arbitrary units, ISTM. 10% of a metre is a decimetre, for instance. Although the most obvious implementation would be to return units %m in that case, which would look a bit ugly. Similarly %² would refer to multiplying by 10,000.

engine/derivative.cc
wdednam: Thanks, I fixed this.

engine/node_latex.cc
wdednam: Fixed it, thanks :)

model/operation.cc
wdednam: Thanks. If I undestand correctly, this means I should get rid of this if conditional. The function input does not have to be unitless.

engine/derivative.cc
highperformancecoder: I think you only need to provide the derivative for the polygamma function:
d/dxψⁿ(x) = ψⁿ⁺¹(x)
polygamma will need to be implemented as a derived class from operator<polygamma>, with an order parameter attribute.
Where you have digamma now, use polygamma(0), unless there is any use case for making it a separate operation.

engine/derivative.cc
```diff
@@ -639,6 +640,53 @@ namespace MathDAG
...
+      {
+        Expr x(expressionCache, expr.arguments[0][0]);
+        return chainRule(x,polygamma(one,x));
+      }
+    else
+      {
+        Expr x(expressionCache, expr.arguments[0][0]);
+        Expr y(expressionCache,  expr.arguments[1][0]);
+        return chainRule(x,polygamma(y+one,x));;
+      }
```
highperformancecoder, line 673: This entails you're making polygamma a binary operation, with the second argument the order. But the second argument can change as a function of time, which would cause a lot of surprises. I'd prefer that the order were a separate attribute of the operation, that is fixed for all simulation time. You could potentially use the arg attribute of Operation for this purpose. It is a double, not an int, but you could check it for integralness at reset time.

engine/node_matlab.cc
```diff
@@ -493,6 +493,41 @@ namespace MathDAG
...
+  {
+    checkArg(0,0);
+    return o<<"gamma("<<arguments[0][0]->matlab()<<")";
+  }
+  
+  template <>
+  ostream& OperationDAG<OperationType::digamma>::matlab(ostream& o) const
+  {
+    checkArg(0,0);
+    return o<<"psi(2,("<<arguments[0][0]->matlab()<<"))";
```
highperformancecoder, line 515: The digamma function is psi(0,x)

highperformancecoder: In matlab, you can omit the order if it is the digamma function, ie psi(x) == psi(0,x)

engine/node_matlab.cc
```diff
@@ -493,6 +493,41 @@ namespace MathDAG
...
+  }      
+  
+  template <>
+  ostream& OperationDAG<OperationType::polygamma>::matlab(ostream& o) const
+  {
+    checkArg(0,0); checkArg(1,0);
+    return o<<"psi(("<<arguments[1][0]->matlab()<<"),("<<arguments[0][0]->matlab()<<"))";
```
highperformancecoder, line 522: You would need the floor function here to ensure the order is an integer.

model/operation.cc
```diff
@@ -435,16 +435,42 @@ namespace minsky
...
+				   auto r=ports[1]->units(check);	 
+				   if (!r.empty()) {
+			           vV->setUnits("%"+r.str());
+			           vV->units.normalise();
+                       return vV->units; 
+		           } else {
+					  vV->setUnits("%");
+					  vV->units.normalise();
+                      return vV->units;
+				   }
```
highperformancecoder, line 456: Please try to maintain consistent indenting - this is quite hard to read!

model/operation.cc
```diff
@@ -435,16 +435,42 @@ namespace minsky
...
+          {
+		  case percent:
+		    { 
+		     // Add % sign to units from input to % operator		
+             if (!ports[0]->wires().empty() && !ports[1]->wires().empty())  
+               {
+				 if (auto vV=dynamic_cast<VariableValue*>(&ports[0]->wires()[0]->to()->item())) 
+	             {    
+				   auto r=ports[1]->units(check);	 
+				   if (!r.empty()) {
```
highperformancecoder, line 448: I don't think this test is required. If r.empty(), then I'm pretty sure r.str() will return the empty string, in which case the true and false parts of this conditional are identical.

model/operation.cc
```diff
@@ -1120,6 +1146,35 @@ namespace minsky
...
+    cairo_move_to(cairo,-6,3);
+    cairo_show_text(cairo,"ψ");
+  }
+  template <> void Operation<OperationType::polygamma>::iconDraw(cairo_t* cairo) const
+  {
+    cairo_move_to(cairo,-7,3);
+    cairo_show_text(cairo,"ψ");
+    cairo_rel_move_to(cairo,0,-3);
+    cairo_set_font_size(cairo,7);
+    cairo_show_text(cairo,"(n)");
```
highperformancecoder, line 1170: If polygamma used the arg attribute for it's order, rather than being a binary op, you could display the actual order here.

engine/derivative.cc
wdednam: Thanks, I've done this. :)

engine/derivative.cc
wdednam: Can I still leave it as a binop as in the most recent commit? I've added it as a ConstantDAG in expr.h:

  inline Expr polygamma(const Expr& x, double y) {
    shared_ptr<OperationDAGBase> r(x.newNode(OperationType::polygamma));
    r->arguments[0].push_back(x);
    r->arguments[1].push_back(**NodePtr(new ConstantDAG(y))**);    
    return Expr(x.cache,r);
  }     

and also call it as polygamma(x,1+expr.arguments[1][0]) in derivative.cc now.

engine/node_matlab.cc
wdednam: I've removed digamma because it is no longer needed if I already have polygamma(0,x).

engine/node_matlab.cc
wdednam: Done, thanks :)

model/operation.cc
wdednam: Sorry, it appears correctly indented in my editor but does not on github because of tabs. I've removed the tabs and it should appear better now.

model/operation.cc
wdednam: I've removed the if conditional in question.

model/operation.cc
wdednam: I sort of do that in equationDisplay.cc, have not considered it here. I guess 
I have not seen an example in which the arg attribute is used for another operation. Is it similar to how it's done in equationDisplay.cc?

wdednam:
Several tests in unittests.sh are failing. The ones involving numerical inaccuracies in testDerivative.cc in particular concern me:

gamma
testDerivative.cc:298:1: error: Failure in singleArgFuncs: Expected 1 +/- 0.003 but was 0.000670325
fact
testDerivative.cc:298:1: error: Failure in singleArgFuncs: Expected 1 +/- 0.003 but was 0.459969
testDerivative.cc:299:1: error: Failure in singleArgFuncs: abs(f->value()-f0)>0.1*f0

 I have not yet discovered an obvious way to fix them.

engine/derivative.cc
highperformancecoder: OK - let's suck it and see.

engine/derivative.cc
highperformancecoder: I would implement the derivative for gamma (which is very similar to the factorial one, of course), and then get rid of digamma as a  concept entirely (since polygamma with 0 as the second argument is equivalent).

engine/node_matlab.cc
highperformancecoder: Sounds good.

model/operation.cc
highperformancecoder: I don't want to get into the whole tabs vs spaces war, but in my experience, you're better off using spaces everywhere.

highperformancecoder: emacs has an "untabify" command that can help here, if you're still using emacs.

model/operation.cc
highperformancecoder: Arg was originally introduced to specify the delta in the difference operation, and the window size in the running sum.
Given  you've decided to stick with the binary function, it might still be nice to dynamically display the order in the symbol here...

highperformancecoder:
This is looking very close now. I agree, you need to sort out those failing tests - they test that your derivative implementation is done correctly by integrating it.

engine/derivative.cc
```diff
@@ -639,6 +640,64 @@ namespace MathDAG
...
+  (const OperationDAG<OperationType::polygamma>& expr)
+  {
+      assert(expr.arguments.size()==2);
+      if (expr.arguments[0].empty())
+        return zero;
+      else 
+        {
+          Expr x(expressionCache, expressionCache.reverseLookup(*expr.arguments[0][0]));
+          if (expr.arguments[1].empty())
+            return (polygamma(x,1))->derivative(*this);
```
highperformancecoder, line 680: Are you sure this works? I would have though chainRule(x, polygamma(x,1)) would be what is needed here. The current expression looks like it would blow up the stack due to recursion.

engine/derivative.cc
```diff
@@ -639,6 +640,64 @@ namespace MathDAG
...
+          else
+            {
+              return (polygamma(x,1+expr.arguments[1][0]))->derivative(*this);
```
highperformancecoder, line 683: ditto

model/operation.cc
```diff
@@ -435,16 +435,36 @@ namespace minsky
...
-        if (check && !ports[1]->units(check).empty())
-          throw_error("function input not dimensionless");
-        return {};
+        switch (type())
+          {
+          case percent:
+            { 
+              // Add % sign to units from input to % operator. Need the first conditional otherwise Minsky crashes		
+                auto r=ports[1]->units(check);	 	 
+                if (!r.empty()) {
```
highperformancecoder, line 444: If r.empty(), you would want to return just "%". So I don't think this conditional should be here.

test/testDerivative.cc
```diff
@@ -290,13 +290,31 @@ SUITE(Derivative)
...
               reset(); 
             }
-          nSteps=1;step(); // ensure f is evaluated
-          // set the constant of integration to the value of f at t=0
-          double f0=f->value();
-          integ.intVar->value(f0);
-          nSteps=800; step();
-          CHECK_CLOSE(1, f->value()/integ.intVar->value(), 0.003);
-          CHECK(abs(f->value()-f0)>0.1*f0); // checks that evolution of function value occurs
+          if (OperationType::Type(op)==OperationType::gamma) {
```
highperformancecoder, line 293: Cleaner to use a switch statement for this type of construct.

test/testDerivative.cc
```diff
@@ -290,13 +290,31 @@ SUITE(Derivative)
...
+             integ.intVar->value(f0);			  
+             nSteps=1883; step();       
+             CHECK_CLOSE(1, f->value()/integ.intVar->value(), 0.003);
+             CHECK(abs(f->value()-f0)>0.1*f0); // checks that evolution of function value occurs             
+          } else if (OperationType::Type(op)==OperationType::fact) {
+             nSteps=1;step(); // ensure f is evaluated
+             // set the constant of integration to the value of f at t=0
+             double f0=f->value();
+             integ.intVar->value(f0);			  
+             nSteps=1448; step();
```
highperformancecoder, line 306: The only difference I can spot between these code branches is the number of steps. Originally it was 800, but it sounds like more is required for these high order transcendentals. Could we just set this to say 2000 for everything, and remove the conditional branches altogether?

test/testUnits.cc
```diff
@@ -184,14 +190,21 @@ SUITE(Units)
     from2->setUnits(from1->unitsStr());
     CHECK_THROW(opp->checkUnits(),std::exception);
   }
+
+  template <> void TestOp::impl<OperationType::polygamma>()
+  {
+    init(OperationType::polygamma);
+    // log's arguments must be dimensionless
```
highperformancecoder, line 197: Please update comment.

engine/derivative.cc
wdednam: I looks like it works, but I have reverted it to chainRule() which seems to work equally well.

engine/derivative.cc
wdednam: Reverted this to chainRule() too.

model/operation.cc
wdednam: When I remove this conditional I get a segfault. I have replaced it with:
if (!ports[1]->wires().empty()), which seems to work.

test/testDerivative.cc
wdednam: I have increased it to 1000, but also had to decrease stepmin and stepmax by a lot and reduce CHECK(abs(f->value()-f0)>0.1*f0);  to CHECK(abs(f->value()-f0)>0.00000000001*f0); in SingleArgFuncs. This is still Volkswagening, but not as specialized as I had before, because it works across the board for all the SingleArgFuncs.

test/testUnits.cc
wdednam: Thanks for catching this, I've updated it.

highperformancecoder:
About the only thing left to do in this PR is to change the display of the polygamma function to display, rather than (n) the actual order at that point in time.

highperformancecoder:
Looks good to me, ready to merge. If you lift the draft classification, I'll merge it.

PR #184, opened 2020-06-05
tried to come at this from various angles, but the wires that connect…

wdednam:
… operators to flow vars on the Godley icon in 1Free.mky don't always get copied after lasso selection. Maybe the selection misses wires when they are attached to flow vars on goldey icons and we need a special condition in Selection::intersects()?

---
This change is [Reviewable](https://reviewable.io/reviews/highperformancecoder/minsky/184)


highperformancecoder:
Don't forget that wires aren't copied unless both endpoints of the wire are copied.

model/canvas.cc
```diff
@@ -390,8 +390,13 @@ namespace minsky
       if (i->visible() && lasso.contains(*i))
         selection.wires.push_back(i);
 
-    if (focusFollowsMouse)
+// always copy on lasso selection on X11 systems. for ticket 1189
+#if defined(__linux__)
```
highperformancecoder, line 394: focusFollowsMouse should be enough here. Now I have focusFollowsMouse to be true, and still the selection is not copied by default. It's a weird one.

model/minsky.cc
```diff
@@ -267,7 +267,10 @@ namespace minsky
     canvas.model->addGroup(g);
     m.populateGroup(*g);
     // Default pasting no longer occurs as grouped items or as a group within a group. Fix for tickets 1080/1098    
-    canvas.selection.clear();    
+    canvas.selection.clear();        
+
+    // Needed to ensure all wires are copied in a complex selection as in, for example, 1Free.mky. For ticket 1190
+    g->splitBoundaryCrossingWires();
```
highperformancecoder, line 273: SplitBoundaryCrossingWires is only for groups. Only needed when items are moved between groups.

highperformancecoder: With copy/paste, wires should not be in the selection unless both endpoints are in the selection. In which case, the wires will not be crossing group boundaries.

model/minsky.cc
```diff
@@ -296,22 +299,30 @@ namespace minsky
                          }
                      return false;
                    });
+    
+    auto copyOfGroups=g->groups;
+    for (auto& i: copyOfGroups)
```
highperformancecoder, line 304: What's the reasoning for pasting the groups first?

model/canvas.cc
wdednam: I my case, it worked on and off when focusFollowsMouse was activated. The same thing seems to happen with the wires that are sometimes copied or not.

model/minsky.cc
wdednam: I understand now. The paste operation sometimes fails with wires that are very curved, at least in the 1Free.mky example that has been my experience. It has something to do with groups.

wdednam:
Curved wires seemed to be selected twice by contains and intersects in selection.h because canvas.selection->numWires() returns 9 when I lasso the Godley icon in 1Free.mky, which only has 6 wires. Furthermore, m.populateGroup(*g) in minsky::paste() misses the curved wires because g->wires.size() and g-numWires() both return 3 when I lasso the same GodleyIcon.

wdednam:
The problem is definitely in minsky::paste(), because "save selection as.." works correctly and the whole group of items, curved wires and all, appears in the new mky file that is created after the selection has been saved. I also noticed that a defining variable has a different definition in minsky::dimensionalAnalysis:

         if (auto v=(*i)->variableCast())
           {
             // check only the defining variables
             **if (v->isStock() && (v->inputWired() || v->controller.lock().get()))**
               v->checkUnits();
           }

and in variable::defined():

    /// @return true if variable is defined (inputWired() || isStock() && controlled)
    bool defined() const **{return inputWired() || (isStock() && controller.lock());}**

Is the latter the most recent version of a defining variable? It doesn't really affect this ticket, but I noticed it while looking for the origin of the bug.

model/minsky.cc
```diff
@@ -266,14 +266,12 @@ namespace minsky
...
-    // Default pasting no longer occurs as grouped items or as a group within a group. Fix for tickets 1080/1098    
-    canvas.selection.clear();    
-
+    
     // convert stock variables that aren't defined to flow variables, and other fix up multiply defined vars
     g->recursiveDo(&GroupItems::items,
                    [&](Items&, Items::iterator i) {
                      if (auto v=(*i)->variableCast())
-                       if (v->defined() || v->isStock())
+                       if ((v->defined() || v->isStock()) && !v->group.lock())
```
highperformancecoder, line 274: How can an item in g's items vector not be owned by g? This condition should never be true.

model/minsky.cc
```diff
@@ -295,23 +293,57 @@ namespace minsky
...
                      return false;
-                   });
+                   });               
+    
+    // curved wires not always included in the new group, add them if they aren't. For ticket 1190.
+    if (canvas.selection.numWires()!=g->numWires())
+      {
+        canvas.selection.recursiveDo(&GroupItems::wires,
+                       [&](Wires&, Wires::iterator w) {
+                         auto alreadyAdded = g->findAny
```
highperformancecoder, line 303: This is a quadratic algorithm, which is very inefficient. Better to do something with a set, which is O(n log n).

model/minsky.cc
```diff
@@ -295,23 +293,57 @@ namespace minsky
...
+                           (&GroupItems::wires,
+                            [&w](const WirePtr& u)
+                            {return u.get()==w->get();});
+                                      
+                         if (!alreadyAdded) {  
+                           auto f=w->get()->from(), t=w->get()->to();
+                           g->addWire(new Wire(f,t,w->get()->coords()));
```
highperformancecoder, line 310: I have already given a reason why wires may not be pasted (potential double definition being the main one), so I don't think this a good idea to do this blindly here. We need to establish why wires are not being added, and fix any issue there.

wdednam:
I'm closing this since you fixed the problem elsewhere...

PR #185, opened 2020-06-11
first real attempt at par and var tabs. selecting a par or var by las…

wdednam:
…so populates the corresponding tab with a sheet containing values, as long as the object is of rank 2. still a WIP however. could probably merge the two classes into a single one.

---
This change is [Reviewable](https://reviewable.io/reviews/highperformancecoder/minsky/185)


model/parameterSheet.cc
```diff
@@ -45,6 +47,137 @@ namespace
...
+                  pango.show();
+                }
+              else
+                {
+                  float x0=0.0, y0=0.0;//+pango.height();
+                  float x=x0, y=y0;
+                  double colWidth=0;
+                  pango.setMarkup("9999");
+                  //float rowHeight=pango.height();
+                  if (value->hypercube().rank()==0)
```
highperformancecoder, line 92: This looks like copy/paste from the Sheet widget. Please consider refactoring this out into some common code to keep things DRY.

model/parameterSheet.cc
```diff
@@ -45,6 +47,137 @@ namespace
       Pango::show();
     }
   };
+}
+
+void ParameterSheet::populateItemvector() {
+  itemvector.clear();	
+  minsky().canvas.selection.recursiveDo(&GroupItems::items,
```
highperformancecoder, line 54: Shouldn't we doing this recursiveDo on minsky().model (or even minsky().canvas.model)?

model/parameterSheet.cc
wdednam: I will, thanks. I just want to get a rudimentary version working for the parameters. My idea is to crop the borders of the rectangles which contain the 2-D hypercube slices to the ink extents, and then to print successive parameter sheets down the parameter tab so that one can scroll to view them. I just need to get ink extents right. For some reason, neither        

               cairo_recording_surface_ink_extents(surface->surface(),0,0,&width,&height);    
nor
               cairo_recording_surface_ink_extents(cairo_get_target (cairo),0,0,&width,&height); 

nor 
                      width = cairo_image_surface_get_width(cairo_get_target(cairo));                                                                 
                      height = cairo_image_surface_get_height(cairo_get_target(cairo));
   
work. they just return 0 for the width and height. Obviously, accessing the surface that has been drawn on is not as straightforward as I thought it would be.

So far, I have only made progress at getting the width as in the attached image. The latest commit reproduces this when I import [this CSV file](https://www.stats.govt.nz/assets/Uploads/Research-and-development-survey/Research-and-development-survey-2019/Download-data/research-and-development-survey-2019-csv.csv
) 

I used cairo_stroke_extents() and cairo_path_extents() for this. Now, if I could only get the correct vertical extents...

parameter_tab_partial

model/parameterSheet.cc
wdednam: Yes. Once I can draw several parameters as cropped sheets down the parameter tab, I will apply the recursive do to all parameters on the canvas.

wdednam:
Finally made some serious progress! (See the attached image.) Still need to work out some kinks, like overlapping rectangles when two or more parameters have different ranks, because then the vertical offsets aren't quite right. When that works, the hard part will come: making the individual hypercube elements editable. Also need to consider parameters of rank 3 or higher. Maybe I can arrange them horizontally  across the tab...

parameter_tab_progress
parameter_tab_progress_canvas

wdednam:
Won't forget to dry out the code either, just keeping everything separate for now in the early stages.

model/parameterSheet.cc
highperformancecoder: Ink extents should be the go. surface_get_width et al will only return the width you specified when creating the surface, which is likely to be infinite. You may have some luck using the ecolab::cairo::Surface wrapper, which include accessors working out some of this stuff.

highperformancecoder:
Looking nice! I like the idea of the tensor parameters being editable.

model/parameterSheet.cc
```diff
@@ -49,31 +50,28 @@ namespace
...
 
-void ParameterSheet::populateItemvector() {
-  itemvector.clear();	
-  minsky().canvas.selection.recursiveDo(&GroupItems::items,
+void ParameterSheet::populateItemVector() {
+  itemVector.clear();	
+  minsky().canvas.model->recursiveDo(&GroupItems::items,
                                         [&](Items&, Items::iterator i) {                                 
-                                          itemvector.emplace_back(*i);
+                                          itemVector.emplace_back(*i);
```
highperformancecoder, line 57: You could test for item being a parameter before placing this in the itemVector.

model/parameterSheet.cc
```diff
@@ -49,31 +50,28 @@ namespace
...
-      //auto& value=ports[0]->getVariableValue();
-      if (!itemvector.empty())
-      //if (item)
-        if (auto v=itemvector[0]->variableCast())
-        //if (auto v=item->variableCast())
+      if (!itemVector.empty())
+      {
+	   float x0=0.0, y0=0.0;//+pango.height();	
+	   double w,h,h_prev,lh; 
+       for (size_t i=0; i<itemVector.size(); ++i)
```
highperformancecoder, line 71: for (auto& i: itemVector) is neater, unless you need to index into the vector for some reason later.

model/variableSheet.cc
```diff
@@ -53,8 +53,9 @@ namespace
 void VariableSheet::populateItemVector() {
   itemVector.clear();	
   minsky().canvas.model->recursiveDo(&GroupItems::items,
-                                        [&](Items&, Items::iterator i) {                                 
-                                          itemVector.emplace_back(*i);
+                                        [&](Items&, Items::iterator i) {
+									      if ((*i)->variableCast()->type()!=VariableType::parameter)
```
highperformancecoder, line 57: Don't you want to reverse the logic here, ie (*i)->variableCast()->type()==VariableType::parameter

model/parameterSheet.cc
wdednam: Thanks. I've fixed this.

model/variableSheet.cc
wdednam: I thought we wanted separate parameter and variable tabs, because we want to make the former editable, while flow vars, in particular, shouldn't be?

model/variableSheet.cc
```diff
@@ -32,184 +31,48 @@ using ecolab::cairo::CairoSave;
...
-            }
-        }
-    }
-  catch (...) {throw;/* exception most likely invalid variable value */}
-}        
+  void VariableSheet::populateItemVector() {
+    itemVector.clear();	
+    minsky().canvas.model->recursiveDo(&GroupItems::items,
+                                       [&](Items&, Items::iterator i) {
+                                         if ((*i)->variableCast()->type()!=VariableType::parameter)
```
highperformancecoder, line 56: Two approaches you can use to share between variablesheet and  parametersheet
1. Add virtual selector function to the common base class that returns returns v && v->type()==parameter / v && v->type()!=parameter respectively in parameterSheet and VariableSheet. Then call that with the argument (*i)->variableCast() in this lambda (which can now be moved to the common base class. Note the presence of the guard v - as it currently stands, this code will do a null dereference whenever the item is not a variable.

wdednam: Thank you, that is a very useful suggestion.

wdednam:
Progress can be seen in the attached screencap.

parameter_tab_progress_rank_3_and_higher

model/minsky.h
```diff
@@ -151,6 +151,8 @@ namespace minsky
 
     /// write current state of all variables to the log file
     void logVariables() const;
+    
+    bool m_multipleEquities=false;
```
highperformancecoder, line 155: What's going on here?

model/minsky.h
```diff
@@ -159,7 +161,8 @@ namespace minsky
     Panopticon panopticon{canvas};
     FontDisplay fontSampler;
     ParameterSheet parameterSheet;
-    VariableSheet variableSheet;    
+    VariableSheet variableSheet;
+    ParVarSheet parVarSheet{parameterSheet};
```
highperformancecoder, line 165: What's this for?

model/variableSheet.cc
```diff
@@ -30,39 +31,44 @@ using ecolab::cairo::CairoSave;
...
+      void setxy(double xx, double yy) {x=xx; y=yy;}
+      void show() {
+        CairoSave cs(cairo);
+        cairo_rectangle(cairo,x,y,w,height());
+        cairo_clip(cairo);
+        cairo_move_to(cairo,x,y);
+        Pango::show();
+      }
+    };
+  }
```
highperformancecoder, line 51: This class is a candidate for being dried out.

model/variableSheet.cc
```diff
@@ -30,39 +31,44 @@ using ecolab::cairo::CairoSave;
...
+          cairo_translate(cairo,offsx,offsy);           
+          draw(cairo);         
+          ecolab::cairo::Surface surf
+            (cairo_recording_surface_create(CAIRO_CONTENT_COLOR_ALPHA,NULL));
+          draw(surf.cairo());
+          m_width=surf.width();
+          m_height=surf.height();              
+        }     
+      }
+    }
```
highperformancecoder, line 71: This method is a candidate for being dried out by moving it to the common base class

highperformancecoder:
Could you please address all the review comments prior to the merge? I don't think there's anything to difficult about them.

model/variableSheet.h
```diff
@@ -21,21 +21,15 @@
...
 #define VARIABLESHEET_H
 #include <cairoSurfaceImage.h>
-#include <item.h>
+#include <parVarSheet.h>
 
 namespace minsky
 {
 	 
-  class VariableSheet: public ecolab::CairoSurface, public ItemT<VariableSheet>
+  class VariableSheet: public ecolab::CairoSurface, public ParVarSheet
```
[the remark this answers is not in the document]
wdednam: I'm not sure I understand these comments.

ParVarSheet is the base class for derived classes ParameterSheet and VariableSheet.

Since I have introduced rank 3 or higher parameters into ParVarSheet::draw(), I don't think that I can use the code in Sheet here, if that is what you meant?

model/parVarSheet.h
```diff
@@ -0,0 +1,44 @@
...
+  class ParVarSheet: public ItemT<ParVarSheet>
+  {     
+  public:
+    double xoffs=80;  
+    double rowHeight=0;
+    double colWidth=50;    
+    float m_width=2e09, m_height=2e09;
+    Items itemVector;
+    void populateItemVector();
+    virtual bool variableSelector(ItemPtr i) {return false;}
```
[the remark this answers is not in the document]
wdednam: Done, will soon commit the new dried out code.

model/variableSheet.h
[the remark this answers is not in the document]
wdednam: Oh, sorry, you've already resolved this conversation. Still getting used to Github...

model/parVarSheet.h
```diff
@@ -22,25 +22,147 @@
...
 #include <item.h>
 #include <variable.h>
+#include "classdesc_access.h"
 
 namespace minsky
 {
+  
+  //class ParameterSheet;	
 	 
   class ParVarSheet: public ItemT<ParVarSheet>
```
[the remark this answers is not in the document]
wdednam: Not anymore. Since draw() is no longer derived from the Item class, I have removed this dependency.

model/parVarSheet.h
```diff
@@ -22,25 +22,147 @@
...
+    int hoverRow=-1, hoverCol=-1;
+    /// computed positions of the table columns
+    std::vector<double> colLeftMargin;          
+    
+    unsigned insertIdx=0, selectIdx=0;                                       // useful for editable par tab!!                                
+
     void populateItemVector();
     virtual bool variableSelector(ItemPtr i) {return false;}
-    void draw(cairo_t* cairo) const override;      
+    void draw(cairo_t* cairo);
```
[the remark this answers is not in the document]
wdednam: Yes, I realized that it needed its own independent draw() method when I started working on making the cells editable and looked at the GodleyTableWindow class for inspiration.

model/minsky.h
wdednam: It sneaked in from my develop branch. I've removed it, thanks.

model/minsky.h
wdednam: I've removed it, thanks.

model/variableSheet.cc
wdednam: Done.

model/variableSheet.cc
wdednam: Done.

wdednam:
Unit tests failing on and off again, on my machine too, but not usually here. 
Screenshot from 2020-07-06 11-12-01

PR #187, opened 2020-06-27
Bugfix 1202

wdednam:
This change is [Reviewable](https://reviewable.io/reviews/highperformancecoder/minsky/187)

### DIFF
--- a/model/godleyIcon.cc
+++ b/model/godleyIcon.cc
@@ -305,6 +305,8 @@ namespace minsky
     // position of margin in absolute canvas coordinate
     float zoomFactor=iconScale()*this->zoomFactor();
     float vdf=variableDisplay? 1: -1; // variable display factor
+    // fix variable spacing on Godleys. Take into account model and item zoom crossover. for ticket 1202.
+    float globalZoomFactor=minsky().canvas.model->zoomFactor() < this->zoomFactor() ? minsky().canvas.model->zoomFactor() : this->zoomFactor();
     float x= this->x() - 0.5*gWidth()+leftMargin();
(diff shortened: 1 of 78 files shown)


model/godleyIcon.cc
```diff
@@ -313,7 +315,7 @@ namespace minsky
         //RenderVariable rv(*v);
         v->rotation(0);
         v->moveTo(x+v->x() - (variableDisplay? v->right(): v->left()), y);
-        y+=v->height();//2*rv.height()*zoomFactor;
+        y+=v->height()*v->zoomFactor()*globalZoomFactor;//2*rv.height()*zoomFactor;
```
highperformancecoder, line 318: I code reviewed this last week, and Github has lost my comments! Grrr!

I really don't understand what you're doing here - in the global group, the godleyTable's zoomFactor will be the same as the global zoomFactor, so you are adjusting these variables by the square of the zoom factor.

Could you please explain a bit your thinking here?

wdednam: Multiplying only by the zoomFactor introduced at the top of this method, which what was done previously with RenderVariable rv(*v), works fine when you are close to he model, but there's a cross-over to another zooming regime when moving far enough out from the model. Here the spacing between the variables on the Godleys begin to increase again. It is less extreme when multiplying by zoomFactor, but still there. It seems that in the approach I took, two of the zoom factors change value in opposite directions and thus annul one another in the "far field" regime.

highperformancecoder:
I have made comments on #1202, and committed what I think is the current best fix for this issue. As a result, I'm closing this without merging.

PR #188, opened 2020-07-03
Feature 94

wdednam:
This change is [Reviewable](https://reviewable.io/reviews/highperformancecoder/minsky/188)


model/item.cc
```diff
@@ -144,7 +144,8 @@ namespace minsky
      
     // then, check whether a resize handle has been selected  
     float z=zoomFactor();
-    if ((abs(x-left()) < portRadius*z || abs(x-right()) < portRadius*z) &&
+    if (!variableCast() && !operationCast() &&
```
highperformancecoder, line 147: Could we not provide an overridden clickType in VariableBase and OperationBase?

model/operation.cc
```diff
@@ -167,6 +169,42 @@ namespace minsky
...
+      cairo_line_to(cairo,-1,-1);
+      cairo_line_to(cairo,1,1);
+      cairo_line_to(cairo,1,0.2);
+      cairo_move_to(cairo,-1,-1);
+      cairo_line_to(cairo,-.2,-1);
+      cairo_move_to(cairo,.2,1);
+      cairo_line_to(cairo,1,1);
+    }
+}  
+
```
highperformancecoder, line 191: Is there a way of drying out this code so we can reuse the Item::drawResizeHandles code as much as possible?

We will probably be asked to render the resize handles differently, perhaps as small boxes.

wdednam:
Made more improvements, removed some dead code in SwitchIcon class. On my machine unittests.sh fail up to 5 times before passing. When it fails, the message I get (in boldface below) is:

wd@wd-X550VXK:~/minsky$ ./unittests.sh 
invalid matrix (not invertible)illegal exception caught in draw(): invalid matrix (not invertible)
invalid matrix (not invertible)illegal exception caught in draw(): invalid matrix (not invertible)
invalid matrix (not invertible)illegal exception caught in draw(): invalid matrix (not invertible)
invalid matrix (not invertible)illegal exception caught in draw(): invalid matrix (not invertible)
attempt to move a group into itselfinvalid matrix (not invertible)illegal exception caught in draw(): invalid matrix (not invertible)
invalid matrix (not invertible)illegal exception caught in draw(): invalid matrix (not invertible)
invalid matrix (not invertible)illegal exception caught in draw(): invalid matrix (not invertible)
invalid matrix (not invertible)illegal exception caught in draw(): invalid matrix (not invertible)
invalid matrix (not invertible)illegal exception caught in draw(): invalid matrix (not invertible)
invalid matrix (not invertible)illegal exception caught in draw(): invalid matrix (not invertible)
**unittests: model/canvas.cc:609: void minsky::Canvas::renameAllInstances(std::__cxx11::string): Assertion `model->numItems()==numItems' failed.**
./unittests.sh: line 16: 26340 Aborted                 (core dumped) $here/test/unittests

model/item.cc
wdednam: Sorry, I missed this comment. I saw that the differences between clickType in Operationbase and the Item class were minimal, which is why I delegated to the method in the base class in that case.

VariableBase has its own clickType but still calls Item::clickType() at the end in "  return Item::clickType(xx,yy);". So, if a case is not fully covered, e.g., for the coupled it op when the resize handles are activated as one moves the mouse over the top of the bounding box midway between the int op and attached var, then Item::clickType() gets called. So, to avoid that, I issue a !variableCast() here. The alternative is to remove the final "  return Item::clickType(xx,yy);" in the overridden members in VariableBase, OperationBase, PlotWidget, etc. I don't know which is better but assumed that, since clickType is expensive, using as few of them as possible would be the way to go?

model/item.cc
```diff
@@ -213,11 +213,15 @@ namespace minsky
...
   void Item::drawResizeHandles(cairo_t* cairo) const
   {
-    double sf=portRadiusMult*zoomFactor();  
-    drawResizeHandle(cairo,right()-x(),top()-y(),sf,0.5*M_PI);
-    drawResizeHandle(cairo,left()-x(),top()-y(),sf,M_PI);
-    drawResizeHandle(cairo,left()-x(),bottom()-y(),sf,1.5*M_PI);
-    drawResizeHandle(cairo,right()-x(),bottom()-y(),sf,0);
+    double sf=portRadiusMult*zoomFactor();
+    // Ops, vars and switch icon only resize from bottom right corner. for ticket 1203
+    if (!switchIconCast() && !variableCast() && (!operationCast() || operationCast()->type()==OperationType::ravel))
```
highperformancecoder, line 218: This is a case where we can refactor in terms of virtual functions, but let's leave it as is for now, and add another ticket to refactor this in terms of virtual functions.

model/operation.cc
```diff
@@ -142,24 +142,15 @@ namespace minsky
...
-               (fabs(hypot(dl,dy)-hypot(wl,h)) < 0.5*portRadius*z) || (fabs(hypot(dr,dy)-hypot(wr,h)) < 0.5*portRadius*z)) &&
-              fabs(fabs(dy)-h) < 0.5*portRadius*z 
-              )
-            return ClickType::onResize;		  
-        } 
-    if (fabs(fabs(dx)-w) < portRadius*z &&
-        fabs(fabs(dy)-h) < portRadius*z &&
-        fabs(hypot(dx,dy)-hypot(w,h)) < portRadius*z)
+    float dx=xx-x(), dy=yy-y();
+    // Ops, vars and switch icon only resize from bottom right corner. for ticket 1203
```
highperformancecoder, line 146: Reading ticket #1203, I don't think there is a requirement to preserve resize handles on all corners for Ravels, plots and sheets. Simplest might be to have a single resize handle on all canvas items.

Also, I think we have a problem with rotated items. Best if the resize behaviour was the same for all to fix that.

model/operation.cc
```diff
@@ -618,48 +573,48 @@ namespace
...
-       // save the render width for later use in setting the clip
-       intVarWidth=rv.width()*z;
-       if (rv.width()<intVar->iWidth()) intVarWidth=intVar->iWidth()*z;
-       // set the port location...
-       intVar->moveTo(x()+r+ivo+intVarWidth, y());
+        // display an integration variable next to it
+        RenderVariable rv(*intVar, cairo);
+        // save the render width for later use in setting the clip
+        intVarWidth=rv.width()*z;
+        if (rv.width()<intVar->iWidth()) intVarWidth=0.5*intVar->iWidth()*z;
```
highperformancecoder, line 588: Shouldn't the condition be <0.5*intVar->iWidth()*z? Also, this seems to be a ratcheting mechanism, ie scales one way, but not the other...

highperformancecoder: I'm going to merge this anyway, but bear in mind I suspect this is not quite right.

wdednam: Sorry, I was working on this, and had changed those conditions to be scalable in both directions. But to make the correction on bugfix-1204, I discarded all the changes on this branch (a few tests were failing and I hadn't quite got the bottom of the problem yet).

wdednam: I'll try again tomorrow morning. Sorry, I don't know where the problem with the test slipped in and was still diagnosing it when I switched to bugfix-1204...

wdednam: Misunderstood the question about the 0.5 factor because I didn't see it... fixing that now.

PR #189, opened 2020-07-07
Bugfix 1204

wdednam:
This change is [Reviewable](https://reviewable.io/reviews/highperformancecoder/minsky/189)


model/godleyTable.h
```diff
@@ -154,8 +154,8 @@ namespace minsky
         throw std::out_of_range("Godley table index error");
       return data[row][col];
     }
-    bool cellInTable(int row, int col) const
-    {return row>=0 && size_t(row)<rows() && col>=0 && size_t(col)<cols();}
+    bool cellInTable(unsigned row, unsigned col) const
```
highperformancecoder, line 157: I had chosen int deliberately here. If unsigned, the row>=0, col>=0 make no sense.
May be OK - -ve numbers are mapped to >2e9, so row<rows() and col<cols() still works...

model/group.cc
```diff
@@ -270,6 +270,14 @@ namespace minsky
           else
             intOp->intVar->controller.reset();
         }
+        
+    // need to deal with godley to set initial width and height. for ticket 1205
+    if (auto godley=dynamic_cast<GodleyIcon*>(it.get()))
```
highperformancecoder, line 275: Why is this set here in addItem? This will mean godleyIcons are resized when being transferred from one group to another.

model/item.cc
```diff
@@ -145,7 +145,7 @@ namespace minsky
     // then, check whether a resize handle has been selected  
     float z=zoomFactor();
     if ((abs(x-left()) < portRadius*z || abs(x-right()) < portRadius*z) &&
-      (abs(y-top()) < portRadius*z || abs(y-bottom()*z) < portRadius*z))
```
highperformancecoder, line 148: Nice catch!

test/testModel.cc
```diff
@@ -603,7 +603,8 @@ SUITE(Canvas)
         auto& group=dynamic_cast<Group&>(*itemFocus);
         group.updateBoundingBox();
         group.relZoom=0.5; // ensure displayContents is false
-        double w=group.iconWidth, h=group.iconHeight;
+        //double w=group.iconWidth, h=group.iconHeight;
```
highperformancecoder, line 606: remove dead code.

model/godleyTable.h
wdednam: Ah okay, I understand now. I changed this because I was getting a test build failure with AEGIS=1, but not when I use only TRAVIS=1. So, I can change it back to int? I also now see that you had fixed the problem in the mean time and I had missed it on my last commit after I had pulled the latest remote. I mean you changed 

 {return row>=0 && row<rows() && col>=0 && col<cols();}

to

 {return row>=0 && size_t(row)<rows() && col>=0 && size_t(col)<cols();}

So, I'm changing it back to int with the above fix.

model/group.cc
wdednam: For some reason, the unit tests were failing before I did this. But it seems the those test failures had more to do with the changes I made in the Group class. So, I've removed it.

test/testModel.cc
wdednam: Done. 

The failure of some tests on my laptop may have something to do with the fact that it freezes up completely sometimes and then I have to restart it. It's doing it more often lately (it's really hot here at the moment, but I have a fan platform underneath the laptop). The Psensor temperature has been spiking to almost a 100 degrees C lately. I can't do anything else at the same time or else in the worst case it freezes up.

So, that may be why I'm seeing more failures, because there aren't enough resources available for some of the tests to run properly. I might run some diagnostics over the weekend.

model/group.cc
wdednam: I didn't know where to initialize the size of the Godley icon now that iconWidth and IconHeight are gone. They were initialized to be 100x100. So, I'm trying the approach you used in Item::resize():

	 // Set initial iWidth() and iHeight() to initial Pango determined values. This resize method is not very reliable. Probably a Pango issue. 

    float w=iWidth(width()), h=iHeight(height()), invZ=1/zoomFactor();   

in GodleyIcon:resize():

	float z=zoomFactor(), **iw=this->iWidth(this->width())**, **ih=this->iHeight(this->height())**, is=iconScale();

It seems to work. Committing now.

model/sheet.cc
```diff
@@ -35,11 +35,13 @@ ClickType::Type Sheet::clickType(float x, float y)
...
-  double w=0.5*m_width*z, h=0.5*m_height*z;
-  // check if (x,y) is within portradius of the 4 corners
-  if (fabs(fabs(dx)-w) < portRadiusMult*z &&
-      fabs(fabs(dy)-h) < portRadiusMult*z &&
-      fabs(hypot(dx,dy)-hypot(w,h)) < portRadiusMult*z)
+  //double w=0.5*m_width*z, h=0.5*m_height*z;
+  //// check if (x,y) is within portradius of the 4 corners
+  //if (fabs(fabs(dx)-w) < portRadiusMult*z &&
+  //    fabs(fabs(dy)-h) < portRadiusMult*z &&
+  //    fabs(hypot(dx,dy)-hypot(w,h)) < portRadiusMult*z)
```
highperformancecoder, line 42: Remove dead code

wdednam: Sorry, I just missed the chance to make this change.. I was working on your suggestions for ticket 1203 and had to commit all my changes before checking out this branch and removing the dead code.

highperformancecoder: No worries - I fixed this up when merging feature #94

PR #193, opened 2020-07-15
partially there with Godley icons, but the same is not true of flow v…

wdednam:
…ars with lhs()

---
This change is [Reviewable](https://reviewable.io/reviews/highperformancecoder/minsky/193)


engine/equations.cc
```diff
@@ -641,9 +641,16 @@ namespace MathDAG
...
+    if (auto v=minsky.definingVar(valueId)) {
       if (v->type()!=VariableType::integral && v->numPorts()>1 && !v->ports[1]->wires().empty())
         r->rhs=getNodeFromWire(*v->ports[1]->wires()[0]);
+	  // ensure that flows used as intial conditions, which inherit their values from other vars, correctly initialise corresponding stock var. for ticket 1137.
+	  auto vi=minsky.variableValues[VariableValue::valueId(v->group.lock(),vv->init)];
+	  if (vi->lhs()) {
+		  vv->init=str(vv->value());        
+		  r->init=vv->init;
+	  }
+	}
```
highperformancecoder, line 653: Indentation looks wrong. Is this code correct?

engine/equations.cc
```diff
@@ -641,9 +641,16 @@ namespace MathDAG
     shared_ptr<VariableDAG> r(new VariableDAG(valueId, nm, type));
     expressionCache.insert(valueId, r);
     r->init=vv->init;
-    if (auto v=minsky.definingVar(valueId))
+    if (auto v=minsky.definingVar(valueId)) {
       if (v->type()!=VariableType::integral && v->numPorts()>1 && !v->ports[1]->wires().empty())
         r->rhs=getNodeFromWire(*v->ports[1]->wires()[0]);
+	  // ensure that flows used as intial conditions, which inherit their values from other vars, correctly initialise corresponding stock var. for ticket 1137.
+	  auto vi=minsky.variableValues[VariableValue::valueId(v->group.lock(),vv->init)];
```
highperformancecoder, line 648: vv->init may be an expression, like 2 foobar, where foobar is the name of another variable. In this case, vi is of type undefined.

engine/equations.cc
```diff
@@ -641,9 +641,16 @@ namespace MathDAG
...
+	  if (vi->lhs()) {
+		  vv->init=str(vv->value());        
+		  r->init=vv->init;
```
highperformancecoder, line 651: This then overwrites the initialisation string with the current value of the defining variable. This is definitely the wrong this to do.

model/godleyIcon.cc
```diff
@@ -277,6 +277,9 @@ namespace minsky
               {
                 FlowCoef fc(table.cell(r,c).substr(start));
                 v.init=fc.str();
+                // ensure that flows used as intial conditions, which inherit their values from other vars, correctly initialise corresponding stock var. for ticket 1137.
+                auto vv=minsky().variableValues[VariableValue::valueIdFromScope(group.lock(),v.init)];
+                if (vv->lhs()) vv->init=str(vv->value());
```
highperformancecoder, line 282: Again, setting the initialisation string the the current value of a variable is the wrong thing. Initialisation should happen a reset time, and computed from the values of the variables at that point in time.

highperformancecoder:
You need to rethink exactly how initialisation is performed. The init string is meant to be (in general) an expression, not just a number. It is used to initialise the value of a variable at reset time.

engine/equations.cc
wdednam: Thanks, I realized that when I managed to get the init value field of the int op and Godley stock var in Kresimir's examples to be populated by the correct name but not its value.

engine/equations.cc
wdednam: Okay, thanks

model/godleyIcon.cc
wdednam: I spent many hours this weekend trying initValue(minsky().variableValues) in several places (equation.cc and godleyIcon.cc). The problem is that it reinitialises ALL the variable values and not just the ones I'm interested in (a flow var with a lhs()).

Should I replicate what you do in ConstantDAG::addEvalOps with initValue():

          result=tmpResult;
        result->init=value;
        ***result=result->initValue(values);**

in VariableDAG::addEvalOps? ConstantDAG has a "value" string member. Would it make sense to add an "init" string member to VariableDAG for this?

highperformancecoder: I'm not entirely sure why that code is there, to be honest. ConstantDAGs may not be associated with variable values stashed in the variableValues map, but if they are, then they should be consistent.

model/variable.cc
```diff
@@ -209,9 +209,21 @@ void VariableBase::ensureValueExists(VariableValue* vv, const std::string& nm) c
...
+  if (value!=minsky().variableValues.end()) {
+    //if (auto i=dynamic_cast<IntOp*>(controller.lock().get()))
+    //  if (i->ports.size()>1 && !i->ports[1]->wires().empty())
+    //     i->intVar->init(i->ports[1]->wires()[0]->from()->item().variableCast()->init());
+    //else if (auto g=dynamic_cast<GodleyIcon*>(controller.lock().get()))
+    //  for (auto v: g->stockVars())
+    //     if (v->ports.size()>1 && !v->ports[1]->wires().empty())
+    //        v->init(v->ports[1]->wires()[0]->from()->item().variableCast()->init());        	  
+	if (inputWired())
+        value->second->init=ports[1]->wires()[0]->from()->item().variableCast()->init();
```
highperformancecoder, line 221: Not sure what you're trying to do here, but the item the wire is from needn't be a variable, so potentially you will be dereferencing a null pointer.

Truth is, only integral variables need their initialisation updated according to what they're attached to, and I think that might be handled elsewhere.

wdednam: > Not sure what you're trying to do here, but the item the wire is from needn't be a variable, so potentially you will be dereferencing a null pointer.

I took inspiration from how vars and pars inherit their units from input wires and set their initial values to the value of the par or var on the left hand side.

> Truth is, only integral variables need their initialisation updated according to what they're attached to, and I think that might be handled elsewhere.

Kresimir also pointed out that stock vars in Godleys are not initialised correctly if their initial value is a symbolic par or var that, in turn, inherits their intial value from a another par or var via a wire.

model/godleyIcon.cc
```diff
@@ -280,8 +280,12 @@ namespace minsky
             string::size_type start=table.cell(r,c).find_first_not_of(" ");
             if (start!=string::npos)
               {
-                FlowCoef fc(table.cell(r,c).substr(start));                      
+                FlowCoef fc(table.cell(r,c).substr(start));                                      
                 v.init=fc.str();              
+                // set initial value of stock var to init value of flow that is defined by a parameter or a constant. for ticket 1137
+                if (auto initVar=minsky().definingVar(VariableValue::valueId(group.lock(),table.cell(r,c).substr(start))))
+                  if (initVar->inputWired() && initVar->type()==VariableType::flow)
+                    v.init=initVar->ports[1]->wires()[0]->from()->item().variableCast()->vValue()->init;
```
highperformancecoder, line 288: This still has the problem that the variable is not connected to another variable, hence variableCast() will retun null, and you get a null dereference error.

model/variable.cc
```diff
@@ -205,13 +205,25 @@ void VariableBase::ensureValueExists(VariableValue* vv, const std::string& nm) c
...
   auto value=minsky().variableValues.find(valueId());
-  if (value!=minsky().variableValues.end())
+  if (value!=minsky().variableValues.end()) {   	
+    try
+      {
+        // set initial value of int var to init value of input to second port. for ticket 1137
+        if (!ports[0]->wires().empty())
+          if (auto i=dynamic_cast<IntOp*>(&ports[0]->wires()[0]->to()->item()))
+            if (!i->ports[2]->wires().empty()) {
+              auto vv=i->ports[2]->wires()[0]->from()->item().variableCast()->vValue();
```
highperformancecoder, line 218: ditto.

highperformancecoder: Probably should also check that ports.size()>2, although that should be the case with an IntOp, so maybe leave it as an assertion.

model/variable.cc
```diff
@@ -205,13 +205,25 @@ void VariableBase::ensureValueExists(VariableValue* vv, const std::string& nm) c
...
+              if (vv) value->second->init=vv->init;
+            }
+      }
+    catch (...)
+      {}
```
highperformancecoder, line 223: What exception are we absorbing here?

model/godleyIcon.cc
```diff
@@ -279,8 +279,13 @@ namespace minsky
...
             if (start!=string::npos)
               {
-                FlowCoef fc(table.cell(r,c).substr(start));                      
+                FlowCoef fc(table.cell(r,c).substr(start));                                      
                 v.init=fc.str();              
+                // set initial value of stock var to init value of flow that is defined by a parameter or a constant. for ticket 1137
+                if (auto initVar=minsky().definingVar(VariableValue::valueId(group.lock(),table.cell(r,c).substr(start))))
+                  if (initVar->inputWired() && initVar->type()==VariableType::flow)
+                    if (auto lhsVar=initVar->ports[1]->wires()[0]->from()->item().variableCast())
+                       v.init=lhsVar->vValue()->init;
```
highperformancecoder, line 288: Code is now good, but I still don't understand why it is necessary. Also, initialisations of the form 2foo will not be handled by this code.

wdednam: I fixed the 2foo case, thanks.

I guess I've given too much attention to this ticket which actually covers very special cases. Maybe we should have changed the status of the ticket to "won't fix"?

wdednam: 2foo

highperformancecoder: I think you've missed the point of 2foo. 2foo means twice foo, so if foo's init string is 2p, then the resultant init string should be 4p.
Init strings can be parsed by the flowcoef class, so you'd want something like:
{
FlowCoef fc1(lhsVar->vValue()->init)
fc1.coef*=fc.coef;
v.init=fc1.str();
}

wdednam: Ah, I see. I thought it was just another way people could name their vars. It makes much more sense now. Thanks :)
2foo_correct

PR #194, opened 2020-07-21
attempt replace GodleyIcon::iconScale() by Item::scaleFactor() toward…

wdednam:
…s fixing tickets 1209 and 1213. I don't understand why Item::scaleFactor behaves differently to GodleyIcon::iconScale if I set it the same way in GodleyIcon::scaleIconForHeight?

---
This change is [Reviewable](https://reviewable.io/reviews/highperformancecoder/minsky/194)

### DIFF
--- a/model/godleyIcon.cc
+++ b/model/godleyIcon.cc
@@ -73,8 +73,8 @@ namespace minsky
     {
       float h=0;
       for (auto& v: vars)
-        {
-          RenderVariable rv(*v);
+        { 
+		  RenderVariable rv(*v);	
           h+=2*rv.height();
           if (h>height) height=h;
           float w=2*rv.width();
@@ -161,24 +161,24 @@ namespace minsky
         else
           editor->enableButtons();
       }
-  }
+  }  
 
   double GodleyIcon::schema1ZoomFactor() const
   {
     if (auto g=group.lock())
-      return iconScale()*g->zoomFactor();
+      return scaleFactor()*g->zoomFactor();
     else
-      return iconScale();
+      return scaleFactor();
   }
 
   void GodleyIcon::resize(const LassoBox& b)
   {
-	float z=zoomFactor(), iw=this->iWidth(svgRenderer.width()), ih=this->iHeight(svgRenderer.height()), is=iconScale();  
-	float minusLeftMargin=iw*z*is, minusBottomMargin=ih*z*is;
+    float z=zoomFactor(), iw=this->iWidth(svgRenderer.width()), ih=this->iHeight(svgRenderer.height()), is=scaleFactor();  
+    float minusLeftMargin=iw*z*is, minusBottomMargin=ih*z*is;
     auto bw=abs(b.x0-b.x1), bh=abs(b.y0-b.y1);
     if (bw<=leftMargin() || bh<=bottomMargin()) return;
-    this->iWidth(iw*(bw-leftMargin())/(minusLeftMargin));
-    this->iHeight(ih*(bh-bottomMargin())/(minusBottomMargin));
+    this->iWidth((bw-leftMargin())/(minusLeftMargin));
+    this->iHeight((bh-bottomMargin())/(minusBottomMargin));
     scaleIconForHeight(bh);
     update();
     moveTo(0.5*(b.x0+b.x1), 0.5*(b.y0+b.y1));
@@ -267,25 +267,26 @@ namespace minsky
       if (table.initialConditionRow(r))
         for (size_t c=1; c<table.cols(); ++c)
           {
-            string name=trimWS(table.cell(0,c));
+            string name=trimWS(table.cell(0,c));           
             auto vi=minsky().variableValues.find(VariableValue::valueId(group.lock(),name));
             if (vi==minsky().variableValues.end()) continue;
-            VariableValue& v=*vi->second;
+            VariableValue& v=*vi->second;           
             v.godleyOverridden=false;
             string::size_type start=table.cell(r,c).find_first_not_of(" ");
             if (start!=string::npos)
               {
-                FlowCoef fc(table.cell(r,c).substr(start));
-                v.init=fc.str();
+                FlowCoef fc(table.cell(r,c).substr(start));                      
+                v.init=fc.str();              
                 v.godleyOverridden=true;
               }
             else
-              {
+              { 
                 // populate cell with current variable's initial value
-                FlowCoef fc(v.init);
+                FlowCoef fc(v.init);   
                 table.cell(r,c)=fc.str();
                 v.godleyOverridden=true;
               }
+
           }
 
 
@@ -309,7 +310,7 @@ namespace minsky
   void GodleyIcon::positionVariables() const
   {
     // position of margin in absolute canvas coordinate
-    float z=iconScale()*this->zoomFactor();
+    float z=scaleFactor()*this->zoomFactor();
     float vdf=variableDisplay? 1: -1; // variable display factor
     float x= this->x() - 0.5*iWidth()*z+0.5*leftMargin();
     float y= this->y() - 0.5*bottomMargin()-0.15*iHeight()*z;
@@ -346,7 +347,7 @@ namespace minsky
   void GodleyIcon::draw(cairo_t* cairo) const
   {
 	  
-    float z=zoomFactor()*iconScale();   
+    float z=zoomFactor()*scaleFactor();   
     positionVariables();
     double titley;
     
@@ -375,7 +376,7 @@ namespace minsky
         CairoSave cs(cairo);
         Pango pango(cairo);
         pango.setMarkup("<b>"+latexToPango(table.title)+"</b>");
-        pango.setFontSize(12*zoomFactor());
+        pango.setFontSize(12*this->zoomFactor());
         cairo_move_to(cairo,-0.5*(pango.width()-leftMargin()), titley);
         pango.show();
       }
@@ -452,8 +453,8 @@ namespace minsky
     auto z=zoomFactor();
     double w=iWidth()*z, h=iHeight()*z;
     // check if (x,y) is within portradius of the 4 corners
-    if ((abs(x-Item::left()) < portRadiusMult*z || abs(x-Item::right()) < portRadiusMult*z) &&
-      (abs(y-Item::top()) < portRadiusMult*z || abs(y-Item::bottom()) < portRadiusMult*z))    
+    if ((abs(x-left()) < portRadiusMult*z || abs(x-right()) < portRadiusMult*z) &&
+        (abs(y-top()) < portRadiusMult*z || abs(y-bottom()) < portRadiusMult*z))    
       return ClickType::onResize;
     // Make it possible to pull wires from variables attached to Godley icons. For ticket 940  
     if (auto item=select(x,y))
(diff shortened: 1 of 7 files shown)


model/godleyIcon.cc
```diff
@@ -267,25 +267,29 @@ namespace minsky
...
             v.godleyOverridden=false;
             string::size_type start=table.cell(r,c).find_first_not_of(" ");
             if (start!=string::npos)
               {
-                FlowCoef fc(table.cell(r,c).substr(start));
+                FlowCoef fc(table.cell(r,c).substr(start));                      
                 v.init=fc.str();
+                // ensure that flows used as intial conditions, which inherit their values from other vars, correctly initialise corresponding stock var. for ticket 1137.
+                //auto vv=minsky().variableValues[VariableValue::valueIdFromScope(group.lock(),v.init)];
+                //if (vv->lhs()) vv->init=str(vv->value());
```
highperformancecoder, line 282: dead code

model/godleyIcon.cc
```diff
@@ -267,25 +267,29 @@ namespace minsky
...
                 v.godleyOverridden=true;
               }
             else
-              {
+              { 
                 // populate cell with current variable's initial value
-                FlowCoef fc(v.init);
-                table.cell(r,c)=fc.str();
+                FlowCoef fc(v.init);   
+				table.cell(r,c)=fc.str();
```
highperformancecoder, line 289: funny indenting

model/godleyIcon.cc
```diff
@@ -450,10 +454,11 @@ namespace minsky
   {
     double dx=fabs(x-this->x()), dy=fabs(y-this->y());
     auto z=zoomFactor();
+    //double w=0.5*Item::width()*z, h=0.5*Item::height()*z;
```
highperformancecoder, line 457: dead code

model/godleyIcon.h
```diff
@@ -40,7 +40,7 @@ namespace minsky
     /// for placement of bank icon within complex
     float flowMargin=0, stockMargin=0;
     /// icon scale is adjusted when Godley icon is resized
-    float m_iconScale=1;
+    //float m_iconScale=1;
```
highperformancecoder, line 43: dead code

model/godleyIcon.h
```diff
@@ -79,15 +79,16 @@ namespace minsky
     void toggleVariableDisplay() {variableDisplay=!variableDisplay;}
     
     /// scale icon until it's height matches \a h 
-    void scaleIconForHeight(float h) {update(); m_iconScale*=h/(bottomMargin()+iHeight()*iconScale()*zoomFactor());}        
+    //void scaleIconForHeight(float h) {update(); m_iconScale*=h/(bottomMargin()+iHeight()*iconScale()*zoomFactor());}
```
highperformancecoder, line 82: dead code

model/godleyIcon.h
```diff
@@ -79,15 +79,16 @@ namespace minsky
...
     /// left margin of bank icon with Godley icon
-    float leftMargin() const {return variableDisplay? flowMargin*iconScale()*zoomFactor(): 0;}
+    float leftMargin() const {return variableDisplay? flowMargin*this->scaleFactor()*zoomFactor(): 0;}
     /// bottom margin of bank icon with Godley icon
-    float bottomMargin() const {return variableDisplay? stockMargin*iconScale()*zoomFactor(): 0;}
+    float bottomMargin() const {return variableDisplay? stockMargin*this->scaleFactor()*zoomFactor(): 0;}
 
     /// icon scale is adjusted when Godley icon is resized
-    float iconScale() const {return m_iconScale;}
+    //float iconScale() const {return m_iconScale;}
```
highperformancecoder, line 91: dead code

highperformancecoder:
All looks reasonable, minor nits aside. Just waiting to see that it passes Travis.

model/godleyIcon.cc
wdednam: I've removed it, thanks.

model/godleyIcon.cc
wdednam: Fixed this up in Emacs.

model/godleyIcon.cc
wdednam: removed.

model/godleyIcon.h
wdednam: ditto

model/godleyIcon.h
wdednam: ditto

model/godleyIcon.h
wdednam: ditto

wdednam:
Thanks. I've committed a better version now, but there are still some issues, mentioned by Steve in tickets 1209, 1213 and now 1214.

model/godleyIcon.h
```diff
@@ -78,17 +78,17 @@ namespace minsky
...
     /// left margin of bank icon with Godley icon
-    float leftMargin() const {return variableDisplay? flowMargin*this->scaleFactor()*zoomFactor(): 0;}
+    float leftMargin() const {return variableDisplay? flowMargin*Item::scaleFactor()*zoomFactor(): 0;}
     /// bottom margin of bank icon with Godley icon
-    float bottomMargin() const {return variableDisplay? stockMargin*this->scaleFactor()*zoomFactor(): 0;}
+    float bottomMargin() const {return variableDisplay? stockMargin*Item::scaleFactor()*zoomFactor(): 0;}
 
     /// icon scale is adjusted when Godley icon is resized
-    //float iconScale() const {return m_iconScale;}
+    float scaleFactor() const override;
```
highperformancecoder, line 90: This is identical to Item::scaleFactor(), so not needed.

model/godleyIcon.h
```diff
@@ -78,17 +78,17 @@ namespace minsky
...
+    float scaleFactor(const float& sf) override;
```
highperformancecoder, line 91: I'm assuming this override is to prevent updating the boundingBox, is that correct?

model/godleyIcon.h
```diff
@@ -40,7 +40,7 @@ namespace minsky
     /// for placement of bank icon within complex
     float flowMargin=0, stockMargin=0;
     /// icon scale is adjusted when Godley icon is resized
-    //float m_iconScale=1;
+    float m_sf=1;
```
highperformancecoder, line 43: Why not use Item::m_sf here?

model/godleyIcon.h
wdednam: Removed it.

model/godleyIcon.h
wdednam: Yes, because the bounding box was wrong after some resize operations.

model/godleyIcon.h
wdednam: I guess I can dynamic_cast to Godley in Item::scaleFactor(const float& sf) to prevent the bounding box from being updated in that case.

highperformancecoder:
On Thu, Jul 23, 2020 at 12:44:35AM -0700, Wynand wrote:
> 
> 
> In model/godleyIcon.h:
> 
> > @@ -40,7 +40,7 @@ namespace minsky
>      /// for placement of bank icon within complex
>      float flowMargin=0, stockMargin=0;
>      /// icon scale is adjusted when Godley icon is resized
> -    //float m_iconScale=1;
> +    float m_sf=1;
> 
> I guess I can dynamic_cast to Godley in Item::scaleFactor(const float& sf) to
> prevent the bounding box from being updated in that case.

No need. Just make m_sf protected, not private, then update in the overridden scaleFactor(const float& sf)

> 
> —
> You are receiving this because you commented.
> 

-- 

                      http://www.hpcoders.com.au

wdednam:
> No need. Just make m_sf protected, not private, then update in the overridden scaleFactor(const float& sf)

Mmmm, I've tried the above, but when I use Item::zoomFactor() in the GodleyIcon Class, I also have to use it in the VariableBase class where attached variables' zoomFactor are calculated and then the Godleyicon does not resize properly...

schema/schema3.cc
```diff
@@ -428,7 +428,10 @@ namespace schema3
         if (y.arg) x1->arg=*y.arg;
       }
    if (auto x1=dynamic_cast<minsky::GodleyIcon*>(&x))
-      {	  
+      {	
+        if (y.width) x1->iWidth(*y.width);
+        if (y.height) x1->iHeight(*y.height);
+        x1->bb.update(*x1);
```
highperformancecoder, line 434: As per comment on PR#195, this update call should be redundant.

model/godleyIcon.h
```diff
@@ -61,6 +59,8 @@ namespace minsky
       bb.update(*this); 
       selected=wasSelected;
     }
+  protected:
+     float m_sf=1;
```
highperformancecoder, line 63: I meant use Item::m_sf. I thought maybe you didn't have access, because Item::m_sf might have been private, hence the comment about making it protected,  but I see it is actually public right now.

highperformancecoder: "Overriding" a member variable like this is going to cause a lot of confusion going forward.

wdednam:
This is my best attempt so far. The test suite passes locally. i hope it also does remotely.

model/godleyIcon.cc
```diff
@@ -298,7 +296,7 @@ namespace minsky
         flowMargin=0;
         accumulateWidthHeight(m_stockVars, stockH, stockMargin);
         accumulateWidthHeight(m_flowVars, flowH, flowMargin);
-        float iw=this->iWidth(), ih=this->iHeight();
+        float iw=this->iWidth()*this->zoomFactor(), ih=this->iHeight()*this->zoomFactor();
```
wdednam, line 299: I'm comparing with the sum of all v->width() or all v->height() with g->iWidth() or g->iHeight(). I added the zoomFactor in item::width() and item::height() for ticket 1204, so I need to multiply iWdith() and iHeight() by the zoomFactor here

schema/schemaHelper.h
```diff
@@ -73,12 +73,11 @@ namespace minsky
     }
     static void setPrivates
     (minsky::GodleyIcon& g, const vector<vector<string> >& data, 
-     const vector<GodleyTable::AssetClass>& assetClass,float iconScale)
+     const vector<GodleyTable::AssetClass>& assetClass)
     {
       setPrivates(g.table, data, assetClass);
-      g.m_iconScale=iconScale;
     }
```
highperformancecoder, line 79: We can now get rid if this helper function too.

highperformancecoder:
I took this for a spin, and it is looking good. The resulting shape of the icon still seems a little counterintuitive when shrinking the icon down towards the flow and stock variables (the icon should really maintain its aspect ratio with the scaleFactor adjust the variables to fit, IMHO). I did make one change setting iWidth/iHeight had a mysterious factor of 1/2, which made no sense. Removing that improved things a little bit. I have integrated the changes now.

I also checked the status of old schema - nothing too hilarious there!

model/godleyIcon.cc
wdednam: This was a mistake, The fact that v->width() and height() now contain the zoomFactor causes problems, as shown in ticket 1221. I'm reverting it to RenderVariable width() and height() and removing the zoomFactors here.

schema/schemaHelper.h
wdednam: I just tried this and it caused the variables attached to GodleyIcons to disappear. So, I've left it in.

PR #196, opened 2020-07-28
fix for Steve's latest problem. Zoomfactor in v->height() and v->widt…

wdednam:
…h() causes growth in godleyicons on mouseover away from standard zoom

---
This change is [Reviewable](https://reviewable.io/reviews/highperformancecoder/minsky/196)


wdednam:
Unit tests have failed again... But not on my machine, probably docker issue again?
unittestspass

PR #198, opened 2020-07-28
possible fixes for tickets 1217 and 1220? build passes locally

wdednam:
This change is [Reviewable](https://reviewable.io/reviews/highperformancecoder/minsky/198)

### DIFF
--- a/model/godleyIcon.cc
+++ b/model/godleyIcon.cc
@@ -314,9 +314,9 @@ namespace minsky
     float y= this->y() - 0.5*bottomMargin()-0.15*iHeight()*z;
     for (auto& v: m_flowVars)
       {
-        // right justification if displayed, left otherwisw
+        // right justification if displayed, left otherwise. Remove this for ticket 1220?
         v->rotation(0);
-        v->moveTo(x+v->x() - (variableDisplay? v->right(): v->left()), y);
+        v->moveTo(x+v->x() - v->right(), y);
         y+=v->height();
       }
     x= this->x() + 0.55*leftMargin()-0.45*iWidth()*z;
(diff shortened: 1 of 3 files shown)


model/godleyIcon.cc
```diff
@@ -314,19 +314,19 @@ namespace minsky
...
     x= this->x() + 0.55*leftMargin()-0.45*iWidth()*z;
     y= this->y() + 0.5*iHeight()*z-0.5*bottomMargin();
 
     for (auto& v: m_stockVars)
       {
-        // top justification at bottom of icon if displayed, bottom justified otherwise
+        // top justification at bottom of icon if displayed, bottom justified otherwise. Remove this for ticket 1220?
         v->rotation(90);
-        v->moveTo(x, y + v->y() - (variableDisplay? v->top(): v->bottom()));
+        v->moveTo(x, y + v->y() - v->top());
```
highperformancecoder, line 329: No - because this will stuff up wires going to the Godley table. See the 1Free example. To solve #1217, we need to disable selecting on subsidiary variables when variableDisplay is false. To solve #1220, we need to disable pulling wires from subsidiary flow variables, which we should probably do regardless of the state of variableDisplay.

model/variable.cc
```diff
@@ -89,6 +89,10 @@ ClickType::Type VariableBase::clickType(float xx, float yy)
       // Ops, vars and switch icon only resize from bottom right corner. for ticket 1203  
       if (fabs(xx-right()) < portRadius*z && fabs(yy-bottom()) < portRadius*z)
         return ClickType::onResize;
+      //if (auto g=dynamic_cast<GodleyIcon*>(controller.lock().get()))
+      //  if (!g->variableDisplay && fabs(xx-g->x()) < g->iWidth()*g->zoomFactor() 
+      //      && fabs(yy-g->y()) < g->iHeight()*g->zoomFactor())
+      //      return ClickType::outside;
```
highperformancecoder, line 95: What was the reasoning for disabling this?

model/godleyIcon.cc
wdednam: I thought it was a long shot. Disabling selecting on subsidiary variables would involve more than just VariableBase::clickType, wouldn't it? I think the mouse events in the Canvas class will also be affected by this?

model/variable.cc
wdednam: Because it didn't seem to work, but maybe it is in the wrong place? I deliberately left it in to get feedback from you.

wdednam:
I'm going to close this pull request. It's back to the drawing board. :)

PR #201, opened 2020-07-31
fix for ticket 1225 and several other improvements (hopefully) to Gro…

wdednam:
…up class

---
This change is [Reviewable](https://reviewable.io/reviews/highperformancecoder/minsky/201)

### DIFF
--- a/gui-tk/wiring.tcl
+++ b/gui-tk/wiring.tcl
@@ -559,23 +559,10 @@ proc doubleButton {x y} {
 }
 # for ticket 1062, new hierarchy of context menu access on mouse right click: wires, items and background canvas.
 bind .wiring.canvas <<contextMenu>> {
-	set item minsky.canvas.item
     if [getWireAt %x %y] {
-		# if not display the contents of groups, disable context menu of wires. for ticket 1225.
-		if [getItemAt %x %y] {
-           switch [$item.classType] {
-		   	  Group { 
-		   	  	  if {![$item.displayContents]} {
-		   	  	  	rightMouseGroup %x %y %X %Y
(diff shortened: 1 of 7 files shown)


gui-tk/wiring.tcl
```diff
@@ -559,10 +559,23 @@ proc doubleButton {x y} {
 }
 # for ticket 1062, new hierarchy of context menu access on mouse right click: wires, items and background canvas.
 bind .wiring.canvas <<contextMenu>> {
+	set item minsky.canvas.item
     if [getWireAt %x %y] {
-        wireContextMenu %X %Y  	
+		# if not display the contents of groups, disable context menu of wires. for ticket 1225.
+		if [getItemAt %x %y] {
```
highperformancecoder, line 565: The idea is right, but wires have a "visible" attribute that more straightforwardly solves this problem.

model/group.cc
```diff
@@ -408,7 +408,7 @@ namespace minsky
       (y-this->y())*sin(rotation()*M_PI/180);
     float dy=(x-this->x())*sin(rotation()*M_PI/180)+
       (y-this->y())*cos(rotation()*M_PI/180);      
-    float w=0.5*iconWidth*z,h=0.5*iconHeight*z;
+    float w=0.5*iWidth()*z,h=0.5*iHeight()*z;
```
highperformancecoder, line 411: It's probably wise to do this change as a separate PR from the #1225 fix. That way, it can be accepted or rejected independently of the other changes.

highperformancecoder:
Changes to iconScale look good. Should use the visible attribute of wire to determine whether to display a context menu.

PR #202, opened 2020-08-01
proposed fix for ticket 1225 using Wire:visible() instead of Group::d…

wdednam:
…isplaycontents()

---
This change is [Reviewable](https://reviewable.io/reviews/highperformancecoder/minsky/202)

### DIFF
--- a/gui-tk/wiring.tcl
+++ b/gui-tk/wiring.tcl
@@ -560,7 +560,11 @@ proc doubleButton {x y} {
 # for ticket 1062, new hierarchy of context menu access on mouse right click: wires, items and background canvas.
 bind .wiring.canvas <<contextMenu>> {
     if [getWireAt %x %y] {
-        wireContextMenu %X %Y  	
+        if [minsky.canvas.wire.visible] {   # prevents wire context menu from being accessed when group contents are not transparent. for ticket 1225.
+			wireContextMenu %X %Y  	
+		} elseif [getItemAt %x %y] {  # Wire::visible() tests for group.lock(), so only this option used here.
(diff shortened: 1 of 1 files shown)


gui-tk/wiring.tcl
```diff
@@ -560,7 +560,11 @@ proc doubleButton {x y} {
 # for ticket 1062, new hierarchy of context menu access on mouse right click: wires, items and background canvas.
 bind .wiring.canvas <<contextMenu>> {
     if [getWireAt %x %y] {
-        wireContextMenu %X %Y  	
+        if [minsky.canvas.wire.visible] {   # prevents wire context menu from being accessed when group contents are not transparent. for ticket 1225.
+			wireContextMenu %X %Y  	
+		} elseif [getItemAt %x %y] {  # Wire::visible() tests for group.lock(), so only this option used here.
+			rightMouseGroup %x %y %X %Y 
+		} else {wireContextMenu %X %Y}
```
highperformancecoder, line 567: This case still posts a wire menu even if over an invisible wire.
Also, in case of clicking over an invisible wire, you're assuming that the item you're over is a group - I'm not sure that is always true.

What I would do is 
if {[getWireAt %x %y] && [minsky.canvas.wire.visible]} {
  wireContextMenu %X %Y 
} elseif [getItemAt %x %y] {
   ...

wdednam: I had assumed for some reason that this would lead to wires that are not in groups, and visible within them, to be non-selectable. But testing it now, I see it works perfectly. Thanks , submitting the code now after running the test suite locally.

PR #205, opened 2020-08-05
proposed fix for ticket 1232

wdednam:
This change is [Reviewable](https://reviewable.io/reviews/highperformancecoder/minsky/205)

### DIFF
--- a/model/item.cc
+++ b/model/item.cc
@@ -46,8 +46,8 @@ namespace minsky
     x.onResizeHandles=false;
     try
       {
-        //cairo_rotate(surf.cairo(),-x.rotation()*M_PI/180);
         x.draw(surf.cairo());
+		cairo_rotate(surf.cairo(),-x.rotation()*M_PI/180);  // perform transformation after drawing, otherwise ink extents not calculated correctly below. For ticket 1232      
       }
     catch (const std::exception& e) 
       {cerr<<"illegal exception caught in draw(): "<<e.what()<<endl;}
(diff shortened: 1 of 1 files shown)


highperformancecoder:
I think we need to be very clear what we mean by the bounding box. With angles that aren't right angles, the bounding box only makes sense if it refers to the unrotated icon. Which is why I made that change. Unfortunately, there is code in the codebase that assumes it refers to the rotated icon, and given that right angle rotations are the most common scenarios, we get away with it.

Then there is the related question of what we mean by Item::left(), right(), top(), bottom(). Probably, it makes sense for them to refer to the four points of the icon, which in general should be a pair of floats, to handle non-right-angle cases. But again, there is plenty of code which assumes they refer to the rotated icon, and then assumes that it is a right angle rotation.

For Malthus, we don't really need to support the non-right-angle rotation. So perhaps we can delay this by ticketting the problem out - but can you please check that undoing that change (as in this PR) causes at least the flipped and the 90 degree Godley stock variables to work?

model/item.cc
```diff
@@ -132,7 +132,45 @@ namespace minsky
...
+      x1=r1.x(right(),bottom());
+      y1=r1.y(right(),bottom());						  
+    }
+    else if (abs(fm)==180) {
+      angle=rotation();
+      x1=right();
+      y1=top();					
+    }
+    else angle=0;	
+    if (auto i=dynamic_cast<const IntOp*>(this))
```
highperformancecoder, line 150: Should do virtual function override for this

model/item.cc
```diff
@@ -132,7 +132,45 @@ namespace minsky
...
+      {
+        if (i->coupled()) {
+          angle=i->intVar->rotation();    // apparently this angle has to be positive???
```
highperformancecoder, line 153: No angle is not necessarily positive, or even in the range 0..360

model/item.cc
```diff
@@ -248,8 +286,46 @@ namespace minsky
...
+      x1=r1.x(right(),bottom());
+      y1=r1.y(right(),bottom());						  
+    }
+    else if (abs(fm)==180) {
+      angle=rotation();
+      x1=right();
+      y1=top();					
+    }
+    else angle=0;	
+    if (auto i=dynamic_cast<const IntOp*>(this))
```
highperformancecoder, line 305: Again, this should be done via virtual function overload

model/item.cc
```diff
@@ -248,8 +286,46 @@ namespace minsky
   }
 
   void BottomRightResizerItem::drawResizeHandles(cairo_t* cairo) const
-  {
-    drawResizeHandle(cairo,right()-x(),bottom()-y(),0.5*resizeHandleSize(),0);
+  { 
+    // ensure resize handle is always drawn on the same corner of variable/items for 90 and 180 degree rotations. for ticket 1232   
+    double fm=std::fmod(rotation(),360), angle;	
+    float x1=right(),y1=bottom();  
+    if (fm==-90 || fm==270) {
```
highperformancecoder, line 293: Clearly this won't work for non-right angle rotations.

model/item.cc
```diff
@@ -248,8 +286,46 @@ namespace minsky
...
+      angle=-rotation();
+      Rotate r1(angle,this->x(),this->y());
+      x1=r1.x(right(),bottom());
+      y1=r1.y(right(),bottom());						  
+    }
+    else if (abs(fm)==180) {
+      angle=rotation();
+      x1=right();
+      y1=top();
```
highperformancecoder, line 302: It strikes me there is a lot of shared code between onResizeHandle and drawResizeHandle that can be DRY'd.

highperformancecoder:
If this code solves the problem for right angle rotations, then I would be happy to merge provided we have a ticket raised to deal with non-right angle situations. DRYing the code and converting to virtual functions can take place as part of that, but later in Say.

wdednam:
Sorry, I know this is late in coming, but as I mentioned in my email to you and Steve, I have partially dried the code, so I'll open a new PR for that. But because we need to add virtual functions to prevent adding a plethora of different Cast() methods, such as the new intOpCast() method I've added, I will understand if you choose to reject the PR.

PR #206, opened 2020-08-05
isn't this supposed to work as a proposed solution to ticket 1231?

wdednam:
This change is [Reviewable](https://reviewable.io/reviews/highperformancecoder/minsky/206)

### DIFF
--- a/model/item.cc
+++ b/model/item.cc
@@ -132,7 +132,9 @@ namespace minsky
 
    bool BottomRightResizerItem::onResizeHandle(float x, float y) const
   {
-    return near(x,y,right(),bottom(),resizeHandleSize(),Rotate(rotation(),this->x(),this->y()));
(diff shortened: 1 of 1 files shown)


model/item.cc
```diff
@@ -132,7 +132,9 @@ namespace minsky
 
    bool BottomRightResizerItem::onResizeHandle(float x, float y) const
   {
-    return near(x,y,right(),bottom(),resizeHandleSize(),Rotate(rotation(),this->x(),this->y()));
+    double angle=rotation() * M_PI / 180.0;	  // Why doesn't this work???
+	double sa=sin(angle), ca=cos(angle);
```
highperformancecoder, line 136: See the Rotate class, which is a convenience class for doing exactly this.

highperformancecoder: Also see my comments on PR#205

wdednam:
I'm going to close this, as it has been dealt with in ticket 1232

PR #207, opened 2020-08-09
improvements to previous commit

wdednam:
This change is [Reviewable](https://reviewable.io/reviews/highperformancecoder/minsky/207)

### DIFF
--- a/model/item.cc
+++ b/model/item.cc
@@ -137,7 +137,7 @@ namespace minsky
        y1=top();					
      }
      else angle=0;	
-     if (auto i=intOpCast())
+     if (auto i=dynamic_cast<const IntOp*>(this))
        {
          if (i->coupled()) {
            angle=i->intVar->rotation();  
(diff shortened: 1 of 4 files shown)


model/item.cc
```diff
@@ -118,6 +118,48 @@ namespace minsky
     {
       return abs(x0-r.x(x1,y1))<d && abs(y0-r.y(x1,y1))<d;
     }
+}
+    
+   std::pair<double,std::pair<float,float>> Item::rotatedPoints() const
```
highperformancecoder, line 123: You can use the Point datatype (defined in geometry.h) for the second part of this pair.

model/item.cc
```diff
@@ -118,6 +118,48 @@ namespace minsky
...
+             angle=-i->intVar->rotation();
+             x1=i->intVar->right();  			  
+             y1=i->intVar->top();  				       		  
+           } else if (fm==0 || fm==360) {
+             angle=-i->intVar->rotation();
+             x1=i->intVar->right(); 			
+             y1=i->intVar->bottom(); 				  
+           } else angle=0;
+         }
+       }
```
highperformancecoder, line 161: This part should be implemented in an overridded virtual function.

highperformancecoder:
Looking pretty good. Couple of nits to address...

PR #208, opened 2020-08-10
Bugfix 1166

wdednam:
This change is [Reviewable](https://reviewable.io/reviews/highperformancecoder/minsky/208)

### DIFF
--- a/model/godleyTableWindow.cc
+++ b/model/godleyTableWindow.cc
@@ -23,11 +23,13 @@
 #include "minsky.h"
 #include <pango.h>
 #include "minsky_epilogue.h"
+#include <boost/locale.hpp>
 
 using namespace std;
 using namespace minsky;
 using ecolab::Pango;
 using namespace ecolab::cairo;
+using namespace boost::locale::conv;
 
 #include <cairo/cairo-ps.h>
 #include <cairo/cairo-pdf.h>
@@ -227,7 +229,8 @@ namespace minsky
             CairoSave cs(cairo);
             if (row!=0 || col!=0)               
               {
-                string text=godleyIcon->table.cell(row,col);
+				// Make sure non-utf8 chars converted to utf8 as far as possible. for ticket 1166.  
+                string text=utf_to_utf<char>(godleyIcon->table.cell(row,col));
                 if (!text.empty())
                   {
                     string value;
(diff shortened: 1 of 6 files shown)


model/godleyTableWindow.cc
```diff
@@ -995,7 +1001,7 @@ namespace {
                 {
                   auto savedItem=minsky().canvas.item;
                   minsky().canvas.item=sv;
-                  minsky().canvas.renameAllInstances(godleyIcon->table.cell(selectedRow,selectedCol));
+                  minsky().canvas.renameAllInstances(utf_to_utf<char>(godleyIcon->table.cell(selectedRow,selectedCol)));
```
highperformancecoder, line 1004: Could you enlighten me as to what utf_to_utf<char> does on an 8 bit string? Nominally, it should just be the identity transformation, but does it somehow clean up non UTF-8 characters also?

highperformancecoder: Ah - it took me a while to find this:

This function may use an explicit encoding name like "Latin1" or "ISO-8859-8", or use std::locale as a parameter to fetch this information from it. It also receives a policy parameter that tells it how to behave if the conversion can't be performed (i.e. an illegal or unsupported character is found). By default this function skips all illegal characters and tries to do the best it can, however, it is possible ask it to throw a conversion_error exception by passing the stop flag to it:

std::wstring s=to_utf<wchar_t>("\xFF\xFF","UTF-8",stop); 
// Throws because this string is illegal in UTF-8

highperformancecoder: So the answer is that it simply ignores invalid utf-8 characters. Neat!

highperformancecoder:
So I think this is the right approach to sanitise any textual input. It may take us a few goes around the roundabout to apply this fix to all areas receiving text input, but that's fine.

highperformancecoder:
Are you sure about the unit test failure? Seems to be passing in Travis.

highperformancecoder:
I think this is good to go as soon as you take it out of draft.

wdednam:
> Are you sure about the unit test failure? Seems to be passing in Travis.

That's strange, mayybe I have an old version of the unit test in my top Minsky directory, which is where I launch tests from if I just want to run one at a time.

model/godleyTableWindow.cc
wdednam: I'm glad this one has been sorted out :)

wdednam:
Do you now feel this one is perhaps not ready yet?

PR #209, opened 2020-08-19
may still be a WIP, but submitting draft PR to get feedback

wdednam:
This change is [Reviewable](https://reviewable.io/reviews/highperformancecoder/minsky/209)


wdednam:
Not done with this one yet, i think. Changing all pair<float,float> to Point type in geometry.h too.

wdednam:
> Not done with this one yet, i think. Changing all pair<float,float> to Point type in geometry.h too.

Mmmm, not as straight forward as I thought it would be... And based on Steve's comment at sourceforge, I'm going to leave it for now.

model/operation.cc
```diff
@@ -492,6 +494,103 @@ namespace minsky
     r.normalise();
     return r;
   }
+  
+  ConnectOpAccessor::ConnectOpAccessor(): ecolab::TCLAccessor<ConnectOp, std::string>
+    ("description",(Getter)&ConnectOp::description,(Setter)&ConnectOp::description) {}  
+    
+  template <> void Operation<OperationType::connector>::iconDraw(cairo_t* cairo) const
+  {/* moved to IntOp::draw() but needs to be here, and is actually called */}
```
highperformancecoder, line 502: Comment out of date

highperformancecoder:
How does connector differ from copy? I couldn't see any specific handling of connector in equations.cc.
Ideally connector should behave like a wire, and actually be invisible to the whole equations subsystem. If you ask what port is at the other end of a wire, the algorithm should iterate through all connector segments in the middle. Maybe it should be a subclass of wire rather than item.

I'm inclined to think this feature needs some more work, so let's not add this to Mathus.

wdednam:
> How does connector differ from copy? I couldn't see any specific handling of connector in equations.cc.
> Ideally connector should behave like a wire, and actually be invisible to the whole equations subsystem. If you ask what port is at the other end of a wire, the algorithm should iterate through all connector segments in the middle. Maybe it should be a subclass of wire rather than item.
> 
> I'm inclined to think this feature needs some more work, so let's not add this to Mathus.

That's correct. It doesn't differ from copy. That thought was nagging at the back of my mind when I was working on it, yesterday.

Knowing this one was tricky, I left it until too late and because Cairo's documentation is so terse, and online examples so rare, it took several rereads to figure out that there are functions which can return many points along a curved path. However, the functions are very memory intensive (my laptop slowed down several times yesterday) and the points are not always current on the drawn curved wire. In other words, we're back at square one and will have to roll our own method to draw a finer grid of handles along a wire.

I'll undo these changes and just keep the randomized searched for the the closest handle in Wire.cc?

model/wire.h
```diff
@@ -51,6 +51,8 @@ namespace minsky
 
     constexpr static float handleRadius=3;
     mutable int unitsCtr=0; ///< for detecting wiring loops in units()
+  protected:  
+    mutable std::vector<std::pair<float,float>> cairoCoords; ///< contains all the internal cairo coordinates used to draw a wire
```
highperformancecoder, line 55: IIUC, this member is only used within draw(), and draw() computes its value every time, so you could just have calcCairoCoords() return the vector of points, and do away with this mutable member.

Does anything inherit from Wire? Just wondering why you chose protected rather than private here.

highperformancecoder: Ah sorry, I just saw the access to cairoCoords in near().

model/wire.cc
```diff
@@ -490,12 +532,17 @@ namespace
     if (c.size()==4)
       return segNear(c[0],c[1],c[2],c[3],x,y);
     else {
-	  // fixes for tickets 991/1095
+      // fixes for tickets 991/1095
       vector<pair<float,float>> p=allHandleCoords(c);
+      if (!cairoCoords.empty()) {
+        p.clear();
```
highperformancecoder, line 538: Could be more simply expressed as p=cairoCoords

highperformancecoder:
This is looking promising. Once you're happy, put it out of draft, and I'll merge. I might wait for this PR to be merged prior to releasing the next beta, which I'll call rc.1

model/wire.h
wdednam: Thanks, making it private but leaving the mutable member in. I don't foresee anything inheriting from Wire using it.

model/wire.cc
wdednam: Thanks, done :) I'm so wary of containers not being populated correctly, since that is how I encounter most segfaults, that I've been overly zealous here.

PR #210, opened 2020-08-25
small fixes to GodleyIcon class to ensure bounding box is correct siz…

wdednam:
…e after toggling editor, variableDisplay and buttonDisplay options

---
This change is [Reviewable](https://reviewable.io/reviews/highperformancecoder/minsky/210)

### DIFF
--- a/model/godleyIcon.cc
+++ b/model/godleyIcon.cc
@@ -189,7 +189,7 @@ namespace minsky
     iWidth((bw-leftMargin())*invZ);
     iHeight((bh-bottomMargin())*invZ);  
     scaleIcon(bw,bh);
-    updateBB(); 
+    updateBoundingBox();
   }
 
   void GodleyIcon::removeControlledItems(Group& g) const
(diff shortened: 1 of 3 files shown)


model/godleyIcon.cc
```diff
@@ -308,7 +308,7 @@ namespace minsky
       }
     
     positionVariables();
-    updateBoundingBox();
+    updateBB();
```
highperformancecoder, line 311: I'm getting a little confused as to the difference between updateBB and updateBoundingBox. IIUC, updateBB is a more strongly worded hint to use the existing iWidth and iHeight attributes. Why did you change updateBoundingBox to updateBB here, and vice versa on line 192?

wdednam: updateBB() updates the "selection" bounding box, if I'm not mistaken. I don't understand why it is necessary and was considering removing it altogether. However, I know from experience that if you put something into the code, there is a very good reason for it. So, I thought update() was the right place to put this more sophisticated bounding box update method. I removed it from resize() because update(), and hence updateBB(), is also called from within scaleIcon().

PR #211, opened 2020-08-26
added very simple test (to be expanded somewhat) of CSVDialog::loadWe…

wdednam:
…bFile in testCSVParser.cc

---
This change is [Reviewable](https://reviewable.io/reviews/highperformancecoder/minsky/211)


test/testCSVParser.cc
```diff
@@ -129,6 +129,36 @@ SUITE(CSVParser)
       
       CHECK(osn.str().find("missing numerical data") != std::string::npos);
     }    
+
+  TEST_FIXTURE(CSVDialog,loadWebFile)
+    {
+	  spec=DataSpec();	
+      string url="https://www.dickimaw-books.com/latex/admin/html/examples/ordergroups.csv";
```
highperformancecoder, line 136: This test is too dependent on somebody else's (unrelated) website. It would probably be sufficient to download that csv file, and then convert the URL into a file:// URL. Also please check before doing so that we have a clear license to use the file that way - maybe just create a random .csv file to avoid any copyright concerns.

wdednam: I've found several online random CSV generators, but those I've tried don't give me a URL download link. For example, https://www.generatedata.com/

So, at first I decided to go the route you followed with Ravel, i.e., dowloading the daily COVID reports from github, but then it occurred to me we might have uploaded a CSV file at sourceforge under one of the raised tickets. Voila: https://sourceforge.net/p/minsky/ravel/20/

So, I'm using BIS_GDP.csv now.

wdednam:
The SpecData class is now covered in testing by over 80% (90% of lines only).
SPECDATAcoverage

I will open a new PR for subsequent additions to testing.

test/testCSVParser.cc
```diff
@@ -128,7 +128,45 @@ SUITE(CSVParser)
...
+      CHECK_EQUAL(1, v.rank()); 
+      spec.toggleDimension(2);
+      CHECK_EQUAL(2,spec.dimensionCols.size());  // equity column is empty
+         
+    }      
+
+  TEST_FIXTURE(CSVDialog,loadWebFile)
+    {
+      string url="https://sourceforge.net/p/minsky/ravel/20/attachment/BIS_GDP.csv";
+      CHECK(url.find("://")!=string::npos);
```
highperformancecoder, line 167: This is more of a test of std::string, which we don't need to bother with, bit oh well...

test/testCSVParser.cc
```diff
@@ -128,7 +128,45 @@ SUITE(CSVParser)
       reportFromCSVFile(isn,osn,*this);
       
       CHECK(osn.str().find("missing numerical data") != std::string::npos);
-    }    
+    }
+    
+   TEST_FIXTURE(CSVDialog,guessSpaceFile)
+    {
+      spec=DataSpec();
+      string url="../test/testEqGodley.csv";
```
highperformancecoder, line 136: This will break the test/00/unittest.sh regression test. The unittest executable is executed out of some random directory, not in the repo directory. Any supporting files, such as this one, is copied into the current directory before running, so you can just refer to the unqualified name.

test/testCSVParser.cc
```diff
@@ -128,7 +128,45 @@ SUITE(CSVParser)
...
+      // Parse path to input file. 
+      boost::regex ex(R"(^(.+)\/([^/]+)$)");      
+      boost::cmatch what;
+      if (regex_match(url.c_str(), what, ex)) {
+       // what[0] contains the whole string 	 
+       // what[1] is the directory path
+       // what[2] is the fragment   
+      } else throw runtime_error("Failure to match file: "+url);
```
highperformancecoder, line 144: I don't understand what you're doing here.

highperformancecoder: If you need to test loading a file url, you can use the getcwd() system call to obtain the full path to the current working directory, then construct a URL like string("file://")+getcwd(buffer)+"/testEqGodley.csv".

test/testCSVParser.cc
wdednam: Thanks, I've got rid of it.

test/testCSVParser.cc
wdednam: I've changed it to just "testGodley.csv" and it works as you've pointed out, thanks.

test/testCSVParser.cc
wdednam: > I don't understand what you're doing here.

Because I didn't know the unit tests are executed from a random directory, I parsed the route to the CSV file. But it is no longer necessary :)


> If you need to test loading a file url, you can use the getcwd() system call to obtain the full path to the current working directory, then construct a URL like string("file://")+getcwd(buffer)+"/testEqGodley.csv".

Thank you, that is a much simpler solution to what I had. However, because of what you explained in a previous post, the parsing of the file path is no longer necessary. I'm just running the coverage test again and will commit the code with the suggested changes as soon as it is done.

PR #212, opened 2020-08-28
Equation display coverage

wdednam:
I'm trying to touch the code in equationDisplay.cc. I manage to select the equations tabs in my new test equationDisplay.sh but I don't know if the equations are rendered on loading testEq.mky, which I think will touch most of the code in equationDisplay.cc?

---
This change is [Reviewable](https://reviewable.io/reviews/highperformancecoder/minsky/212)

### DIFF
--- a/test/00/equationDisplay.sh
+++ b/test/00/equationDisplay.sh
@@ -32,25 +32,20 @@ trap "fail" 1 2 3 15
 # use \$ in place of $ to refer to variable contents
 # exit 0 to indicate pass, and exit 1 to indicate failure
 cat >input.tcl <<EOF
-source assert.tcl
-proc afterMinskyStarted {} {
-minsky.load $here/test/testEq.mky
-.tabs select 1
-minsky.equationDisplay.requestRedraw
-assert {[lindex [.tabs tabs] [.tabs index current]]==".equations"}
-#event generate .equations.canvas  <B1-Motion> -x 100 -y 100 -warp 1
-minsky.equationDisplay.requestRedraw
-puts [minsky.equationDisplay.width]
-puts [minsky.equationDisplay.height]
-set [minsky.equationDisplay.offsx] 100
-set [minsky.equationDisplay.offsy] 100
-assert {[minsky.equationDisplay.width]>0}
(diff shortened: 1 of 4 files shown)


test/00/equationDisplay.sh
```diff
@@ -0,0 +1,57 @@
...
+# exit 0 to indicate pass, and exit 1 to indicate failure
+cat >input.tcl <<EOF
+source assert.tcl
+proc afterMinskyStarted {} {
+minsky.load $here/test/testEq.mky
+.tabs select 1
+minsky.equationDisplay.requestRedraw
+assert {[lindex [.tabs tabs] [.tabs index current]]==".equations"}
+#event generate .equations.canvas  <B1-Motion> -x 100 -y 100 -warp 1
+minsky.equationDisplay.requestRedraw
```
highperformancecoder, line 42: Is this enough to trigger the redraw? Or do you need to bring the equation tab to the fore.

test/testMinsky.cc
```diff
@@ -189,6 +189,18 @@ SUITE(Minsky)
       CHECK_CLOSE(0.3, var["h"]->value(), 1e-4);
 
     }
+    
+//  TEST_FIXTURE(TestFixture,displayEquations)
```
highperformancecoder, line 193: Not surprised this didn't work :)

test/00/equationDisplay.sh
wdednam: I trigger redraw explicitly now, so let's hope it improves the coverage score. If it does, I'll add similar tests for the parameter and variable sheets...

test/testMinsky.cc
wdednam: Hehe, it had me going around circles for a while this morning :)

test/00/equationDisplay.sh
wdednam: Hmmm, it seems like "if (surface.get())" in draw() is always false in parVarSheet.cc and minsky.cc, because the if conditional is touched but then skipped...

wdednam:
MinskyCoverage

PR #214, opened 2020-08-31
added tmax attribute in schema3 and rungeKutta, but the attribute is …

wdednam:
…not being persisted?

---
This change is [Reviewable](https://reviewable.io/reviews/highperformancecoder/minsky/214)


schema/schema3.cc
```diff
@@ -347,6 +347,7 @@ namespace schema3
     m.fileVersion=minskyVersion;
     
     static_cast<minsky::RungeKutta&>(m)=rungeKutta;
+    m.tmax=tmax;
```
highperformancecoder, line 350: This statement shouldn't be needed, as the previous statement will do it for you.

schema/rungeKutta.h
```diff
@@ -30,6 +30,7 @@ namespace minsky
     bool implicit{false};
     int simulationDelay{0};
     std::string timeUnit;
+    double tmax;
```
highperformancecoder, line 33: Also need to move Minsky::t0 into this struct.

schema/schema3.h
```diff
@@ -231,6 +232,7 @@ namespace schema3
       Minsky(*m.model)  {
       minskyVersion=m.minskyVersion;
       rungeKutta=m;
+      tmax=m.tmax;
```
highperformancecoder, line 235: Again, not needed.

highperformancecoder:
The reason this code change doesn't solve the problem is because the actual tmax variable is a TCL variable. You need to modify minsky.tcl to refer the this new minsky.tmax attribute, and then this should be good to go.

wdednam:
Okay, the only thing that needs to be done with this is to get checkXMLSchemas.sh to pass. I thought that resaving all the example mky files (so that they contain the tmax and t0 attributes)  would solve it, but I was wrong.

I am stumped, so I would really appreciate a hint for how to update the test so that it works. Thank you.

highperformancecoder:
rewriting all the example files needs to be done whenever the schema changes. There is a script for doing that (test/rewriteMky.tcl) that simplifies that task  - sorry I should have mentioned.

Sometimes checkXMLSchmeas fails because it can't phone home to w3c.

wdednam:
Note that checkXMLSchemas.sh now passes on my machine if run from Minsky's top folder:
checkXMLSchemas_pass

wdednam:
It's not working yet, sorry. The new simTMax and simTStart in minsky.tcl were stuffing things up, grrr. I'm trying to fix it again now after having removed them...

PR #215, opened 2020-09-01
Added percent constant which Steve requested

wdednam:
Also changed previous % operator such that it is Pango rendered now.

---
This change is [Reviewable](https://reviewable.io/reviews/highperformancecoder/minsky/215)


highperformancecoder:
I'm still a little perplexed by this feature. I thought the percent operator basically divides its input by 100, so 10% -> 0.1 etc. What you implemented here is just a named constant for 100. Maybe that's needed, but I would have chosen the name "hundred" rather than percentConst. But what is wrong with the constant "100" - it takes 3 keystrokes to type?

wdednam:
> I'm still a little perplexed by this feature. I thought the percent operator basically divides its input by 100, so 10% -> 0.1 etc. What you implemented here is just a named constant for 100. Maybe that's needed, but I would have chosen the name "hundred" rather than percentConst. But what is wrong with the constant "100" - it takes 3 keystrokes to type?

"Hundred" is definitely better than percentConst, yes :)

The point about just typing "100" when creating a constant is a very good one. I guess Steve wants an easy to find "percent" operator and that where I put it is a bit out of the way. Here is what he requested at sourceforge:

> One more point--I know I'm breaching protocol putting it here(!), but could you also add one more "fundamental constant": a % operator. The simplest version of this would be a block with that label that simply had the value 100 hard wired into it. We could get fancier in a later release with it being a block with an input and output which was 100 times the input. But simply the number 100 hard wired into a parameter block with the label % would be sufficient now.

[https://sourceforge.net/p/minsky/tickets/1079/#d8f7](https://sourceforge.net/p/minsky/tickets/1079/#d8f7)

highperformancecoder:
But we already have the % operator (multiply by 100) that he wanted. I think we decided to "classify" it as a fundamental constant, in order for it to appear on the same menu as pi et al.

wdednam:
> But we already have the % operator (multiply by 100) that he wanted. I think we decided to "classify" it as a fundamental constant, in order for it to appear on the same menu as pi et al.

I'm very confused now too. I think I added a % operator to the unary operators under sqrt. And because it is right near the bottom of that list, Steve may not have noticed it. It is in fact the fancier version he is talking about:

> We could get fancier in a later release with it being a block with an input and output which was 100 times the input.

So, the latest addition merely addresses the first part he requested:

> One more point--I know I'm breaching protocol putting it here(!), but could you also add one more "fundamental constant": a % operator. The simplest version of this would be a block with that label that simply had the value 100 hard wired into it. ... But simply the number 100 hard wired into a parameter block with the label % would be sufficient now.

I guess he wants a % label on it instead of using a constant with "100" showing on top of it. 

So, should I just move the existing % unary operator to the fundamental consts category?

highperformancecoder:
On Thu, Sep 03, 2020 at 01:56:12AM -0700, Wynand wrote:
> 
> So, should I just move the existing % unary operator to the fundamental consts
> category?


Yes - I think that resolves the task.


-- 

                      http://www.hpcoders.com.au

wdednam:
> On Thu, Sep 03, 2020 at 01:56:12AM -0700, Wynand wrote: So, should I just move the existing % unary operator to the fundamental consts category?
> Yes - I think that resolves the task.
> […](#)
> --
> ---------------------------------------------------------------------------- Dr Russell Standish Phone 0425 253119 (mobile) Principal, High Performance Coders hpcoder@hpcoders.com.au http://www.hpcoders.com.au ----------------------------------------------------------------------------

Okay, I will do that firs thing tomorrow morning. I will also try t fix the issue with the coverage test for CSVDialog, thanks for your tips on how to do that

PR #216, opened 2020-09-02
attempt to improve CSVDialog coverage almost but not quite successful...

wdednam:
Both .wiring.csvImport.buttonBar.ok invoke  and csvImportDialogOK fail with segmentation fault at the end of the test and I don't understand why...

### DIFF
--- a/test/00/importCSV.sh
+++ b/test/00/importCSV.sh
@@ -1,0 +1,54 @@
+#! /bin/sh
+
+here=`pwd`
+if test $? -ne 0; then exit 2; fi
+tmp=/tmp/$$
+mkdir $tmp
+if test $? -ne 0; then exit 2; fi
+cd $tmp
+if test $? -ne 0; then exit 2; fi
+
+fail()
+{
+    echo "FAILED" 1>&2
+    cd $here
+    chmod -R u+w $tmp
+    rm -rf $tmp
+    exit 1
+}
+
+pass()
+{
+    echo "PASSED" 1>&2
+    cd $here
+    chmod -R u+w $tmp
+    rm -rf $tmp
+    exit 0
+}
+
+trap "fail" 1 2 3 15
+
+cat >input.tcl <<EOF
+source $here/test/assert.tcl
+proc afterMinskyStarted {} {
+  minsky.addVariable par parameter
+  assert {[findObject Variable:parameter]}  
+  editItem
+  assert {[winfo ismapped .wiring.editVar]}     
+  .wiring.editVar.buttonBar.import invoke
+  CSVImportDialog
+  assert {[winfo ismapped .wiring.csvImport]}      
+  minsky.value.csvDialog.url https://sourceforge.net/p/minsky/ravel/20/attachment/BIS_GDP.csv
+  minsky.value.csvDialog.loadFile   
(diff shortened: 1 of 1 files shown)


test/00/importCSV.sh
```diff
@@ -0,0 +1,54 @@
...
+  minsky.addVariable par parameter
+  assert {[findObject Variable:parameter]}  
+  editItem
+  assert {[winfo ismapped .wiring.editVar]}     
+  .wiring.editVar.buttonBar.import invoke
+  CSVImportDialog
+  assert {[winfo ismapped .wiring.csvImport]}      
+  minsky.value.csvDialog.url https://sourceforge.net/p/minsky/ravel/20/attachment/BIS_GDP.csv
+  minsky.value.csvDialog.loadFile   
+  # Both this button invoke and csvImportDialogOK fail with segmentation fault here and I don't understand why...
```
highperformancecoder, line 43: Have you run this under a debugger?

wdednam: Ah, didn't think I could run a bash script in a debugger, but of course I can, just not GDB:

[https://tldp.org/LDP/Bash-Beginners-Guide/html/sect_02_03.html](https://tldp.org/LDP/Bash-Beginners-Guide/html/sect_02_03.html)

wdednam: Mmm, the debugging does not tell me much apart from the exit code "139", but a search online for the error code 139 tells me that there is a problem with file i/O:

[https://stackoverflow.com/questions/42882168/how-to-solve-exit-code-139-error-when-reading-from-file-on-unix/42882179](https://stackoverflow.com/questions/42882168/how-to-solve-exit-code-139-error-when-reading-from-file-on-unix/42882179)

I still don't understand why there's a problem though. I thought it may be because the file was temporary, having been downloaded from the internet. But I've tried $here/test/testGodley.csv which loads without generating error reports, e.g., "duplicate key action?", and it returns a segfault too.

Even when I click on "import" in the CSV dialog box during the test, allowed by using a sufficiently long "aftter 10000 {tcl_exit}", the script crashes.

highperformancecoder: Ah - let me explain my technique. Firstly, I define the shell variable 
  here = `pwd`
Then I copy paste the "here document" (starting with cat and ending in EOF) to generate a copy of the tcl input file (often input.tcl, though sometimes different names).
Then I run gdb on minsky. 
Then type "run input.tcl".

wdednam: Thank you, that's brilliant and finally got me somewhere :)

minsky: engine/variableValue.cc:45: const minsky::VariableValue& minsky::VariableValue::operator=(const civita::TensorVal&): Assertion `(isFlowVar() && x.size()+m_idx<=ValueVector::flowVars.size()) || (!isFlowVar() && x.size()+m_idx<=ValueVector::stockVars.size())' failed.

I will get to the bottom of it.

wdednam: So, it seems the data does not get loaded into the parameter I've created when I issue csvImportDialogOk. That means I'll have to add CSVParser commands...

wdednam:
Minsky_engine_coverg

PR #217, opened 2020-09-09
Does this achieve the goal in feature ticket 126?

wdednam:
This change is [Reviewable](https://reviewable.io/reviews/highperformancecoder/minsky/217)

### DIFF
--- a/gui-tk/minsky.tcl
+++ b/gui-tk/minsky.tcl
@@ -273,7 +273,7 @@ if {[tk windowingsystem] == "aqua"} {
     event add <<middleMouse-Motion>> <B3-Motion>
     event add <<middleMouse-ButtonRelease>> <B3-ButtonRelease>
 } else {
-    event add <<contextMenu>> <Button-3> <Control-Button-2>
+    event add <<contextMenu>> <Button-3> <Control-Button-1>
     event add <<middleMouse>> <Button-2>
     event add <<middleMouse-Motion>> <B2-Motion>
     event add <<middleMouse-ButtonRelease>> <B2-ButtonRelease>
--- a/gui-tk/minsky.tcl
+++ b/gui-tk/minsky.tcl
@@ -273,7 +273,7 @@ if {[tk windowingsystem] == "aqua"} {
     event add <<middleMouse-Motion>> <B3-Motion>
     event add <<middleMouse-ButtonRelease>> <B3-ButtonRelease>
 } else {
-    event add <<contextMenu>> <Button-3>
+    event add <<contextMenu>> <Button-3> <Control-Button-2>
     event add <<middleMouse>> <Button-2>
     event add <<middleMouse-Motion>> <B2-Motion>
     event add <<middleMouse-ButtonRelease>> <B2-ButtonRelease>
--- a/gui-tk/minsky.tcl
+++ b/gui-tk/minsky.tcl
@@ -268,12 +268,12 @@ source $minskyHome/library/obj-browser.tcl
 # Macs have a weird numbering of mouse buttons, so lets virtualise B2 & B3
 # see http://wiki.tcl.tk/14728
 if {[tk windowingsystem] == "aqua"} {
-    event add <<contextMenu>> <Button-2> <Control-Button-1>
+    event add <<contextMenu>> <Button-2> <Meta-Button-1>
     event add <<middleMouse>> <Button-3>
     event add <<middleMouse-Motion>> <B3-Motion>
     event add <<middleMouse-ButtonRelease>> <B3-ButtonRelease>
 } else {
-    event add <<contextMenu>> <Button-3> <Control-Button-1>
+    event add <<contextMenu>> <Button-3>
     event add <<middleMouse>> <Button-2>
     event add <<middleMouse-Motion>> <B2-Motion>
     event add <<middleMouse-ButtonRelease>> <B2-ButtonRelease>
--- a/gui-tk/wiring.tcl
+++ b/gui-tk/wiring.tcl
@@ -211,10 +211,12 @@ proc wrapHoverMouse {op x y} {
 }
     
 bind .wiring.canvas <ButtonPress-1> {wrapHoverMouse mouseDown %x %y}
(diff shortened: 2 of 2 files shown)


gui-tk/minsky.tcl
```diff
@@ -273,7 +273,7 @@ if {[tk windowingsystem] == "aqua"} {
     event add <<middleMouse-Motion>> <B3-Motion>
     event add <<middleMouse-ButtonRelease>> <B3-ButtonRelease>
 } else {
-    event add <<contextMenu>> <Button-3>
+    event add <<contextMenu>> <Button-3> <Control-Button-2>
```
highperformancecoder, line 276: Yes that should work, except it should be <Control-Button-1>, not <Control-Button-2>
Very neat - even I didn't think it would be quite this simple.

gui-tk/minsky.tcl
```diff
@@ -273,7 +273,7 @@ if {[tk windowingsystem] == "aqua"} {
     event add <<middleMouse-Motion>> <B3-Motion>
```
highperformancecoder, line 273: Interesting - it looks like Macs have already been dealt with on line 271...

gui-tk/minsky.tcl
wdednam: Thanks, I've fixed it :)

gui-tk/minsky.tcl
wdednam: I noticed that too, which is why it turned out so simple to implement. I had inspiration :)

gui-tk/minsky.tcl
wdednam: I'm not sure it works... I don't know if it is a limitation of Ubuntu, but it doesn't work on my system.

wdednam: I've been reading the documentation and it seems to me that the second binding <Control-Button-1> is being ignored (see under the heading "Multiple Matches":

[https://www.tcl.tk/man/tcl8.4/TkCmd/bind.htm](https://www.tcl.tk/man/tcl8.4/TkCmd/bind.htm)

As far as I understand second bindings get ignored when they have been attached to some other event or binding, but upon doing a "grep -r" on <Control-Button-1> (or <Meta-Button-1> or other variations thereof) from Minsky's top directory, I don't see it anywhere else but inside minsky.tcl where I've made the change.

Maybe it is my Spanish keyboard? I say this because I disabled the right-click button event and only left <Control-Button-1> active, and it still did nothing when I tried it on the Canvas. Would you mind trying it out on yours, please? Thanks.

highperformancecoder: Control double left mouse click works. But why doesn't control left mouse click?

highperformancecoder: I think I've worked out the issue. Control mouse is already bound to individually adding and removing items from a selection. Could you do a bit more research as to how to deal with context menus in that case? Control mouse click is a common idion for individual selections, we don't want to lose that.

wdednam: Ah, I didn't see the "Control-ButtonPRESS-1" bound to controlMouseDown. Thanks for pointing it out, I was only searching for "Control-Button-1". I've tried "Alt-Button-1", but I think it has already been bound to mouse coordinates. Then, I've managed to check  that "Control-Double-Button-1" works like you mentioned at sourceforge. I will see if I can't use the [bindtags](https://www.tcl.tk/man/tcl8.4/TkCmd/bindtags.htm) command to change the order in which "Control-Button-1" is executed. Since in once case it is used to select an item and the other to open the context menu on the background canvas, it may be possible to differentiate between those two events using bindtags.

gui-tk/wiring.tcl
```diff
@@ -235,7 +235,7 @@ bind .wiring.canvas <Button-5> {zoomAt  %x %y [expr 1.0/1.1]}
 bind .wiring.canvas <MouseWheel> { if {%D>=0} {zoomAt %x %y 1.1} {zoomAt  %x %y [expr 1.0/(1.1)]} }
 
 if {[tk windowingsystem]=="aqua"} {
-    bind .wiring.canvas <Command-Button-1> {
+    bind .wiring.canvas <Meta-Button-1> {
         tk_messageBox -message "Mouse coordinates %x %y"
```
highperformancecoder, line 239: Maybe we'll have to forgo the mouse coordinate reporting function on Macs. It's only for debugging purposes anyway.

FYI, the Meta modifier only works on certain historical keyboards  - eg the Sun keyboard had a meta key IIRC, with a diamond symbol on it. I only ever used it access emacs "meta" functions(eg meta-x). On PC keyboards, emacs interpreted Alt as the meta modifier, and on Mac keyboards, it interprets Command as such (or maybe "Apple", I forget - my last Apple keyboard would be now more than 20 years old, and all the labels worn off).

Maybe I should just take this ticket over, and test out the Mac pathways through this script.

wdednam: > Maybe we'll have to forgo the mouse coordinate reporting function on Macs. It's only for debugging purposes anyway.
> 
> FYI, the Meta modifier only works on certain historical keyboards - eg the Sun keyboard had a meta key IIRC, with a diamond symbol on it. I only ever used it access emacs "meta" functions(eg meta-x). On PC keyboards, emacs interpreted Alt as the meta modifier, and on Mac keyboards, it interprets Command as such (or maybe "Apple", I forget - my last Apple keyboard would be now more than 20 years old, and all the labels worn off).
> 


I see what I've done wrong now. The $meta option confused me, because I thought it was defined as "set meta Control" for the keyboard shortcuts. Hence, I didn't understand that it could be used on Macs as command or control. I saw the Meta modifier in Tcl's documentation and thought you meant that instead...

> Maybe I should just take this ticket over, and test out the Mac pathways through this script.

That might be a good idea, thanks. I have no experience with Macs keyboards and because I'm using keyboards with a Spanish layout, which sometimes caused me headaches on Windows machines, it makes things all the more confusing for me.

highperformancecoder:
I fixed this up in a different commit, and tested it on a real Mac.

PR #218, opened 2020-09-13
Bugfix 1258

wdednam:
proposed solution for ticket 1258: do not check for alreadyDefined vars in empty canvas or new system

---
This change is [Reviewable](https://reviewable.io/reviews/highperformancecoder/minsky/218)

### DIFF
--- a/model/minsky.cc
+++ b/model/minsky.cc
@@ -283,15 +283,13 @@ namespace minsky
   void Minsky::paste()
   {
     istringstream is(getClipboard());
-    xml_unpack_t unpacker(is); 
+    xml_unpack_t unpacker(is);
     schema3::Minsky m(unpacker);
     GroupPtr g(new Group);
     g->self=g;
     m.populateGroup(*g);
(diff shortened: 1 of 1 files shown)


model/minsky.cc
```diff
@@ -283,13 +283,15 @@ namespace minsky
...
+    xml_unpack_t unpacker(is); 
     schema3::Minsky m(unpacker);
     GroupPtr g(new Group);
     g->self=g;
     m.populateGroup(*g);
     // Default pasting no longer occurs as grouped items or as a group within a group. Fix for tickets 1080/1098    
-    canvas.selection.clear();    
+    canvas.selection.clear();
+    // The following is only necessary if one pastes into an existing model. For ticket 1258   
+	if (!history.empty() || !canvas.model.get()->empty()) {
```
highperformancecoder, line 294: No problem with this change, however I don't understand how it achieves anything. If canvas.model were empty, then the alreadyDefined flag below will be false. For parameters, v->isStock() will be false, so nothing happens for them. Stock variables without definitions will be retyped as flow variables, of course, but if they were left as stock variable, the  model will be inconsistent.

PR #219, opened 2020-09-15
Bugfix 1258

wdednam:
Oops. Because the branch already existed and was previously merged, it didn't ask me to open a new pull request. Sorry, I would have done so sooner, but I was too tired last night after committing the code.

---
This change is [Reviewable](https://reviewable.io/reviews/highperformancecoder/minsky/219)

### DIFF
--- a/model/minsky.cc
+++ b/model/minsky.cc
@@ -297,7 +297,7 @@ namespace minsky
       // convert stock variables that aren't defined to flow variables, and other fix up multiply defined vars
       g->recursiveDo(&GroupItems::items,
                      [&](Items&, Items::iterator i) {
-                       if (auto v=(*i)->variableCast())
+                       if (auto v=(*i)->variableCast()) {
                          if (v->defined() || v->isStock())
                            {
                              // if defined, check no other defining variable exists
@@ -329,7 +329,18 @@ namespace minsky
                                  assert(v->ports.size()>1 && !v->ports[1]->wires().empty());
                                  g->removeWire(*v->ports[1]->wires()[0]);
                                }
-                           } 
+                           }
+                         // make sure parameter values are preserved after copy/paste. for ticket 1258  
+                         if (v->type()==VariableType::parameter) 
+                         {
+                            auto existingParm = canvas.model->findAny
+                               (&GroupItems::items,
+                                [&v](const ItemPtr& j)
+                                {return j->variableCast() && j->variableCast()->valueId()==v->valueId();});
+                            if (existingParm) v->init(existingParm->variableCast()->init());
+                            else v->ensureValueExists(v->vValue().get(),v->name());   
+				         }  
+					   } 
                        return false;
                      });
     }                              
@@ -343,16 +354,16 @@ namespace minsky
     canvas.model->moveContents(*g); 
 
     // leave newly ungrouped items in selection
-    for (auto& i: copyOfItems) canvas.selection.ensureItemInserted(i);
+    for (auto& i: copyOfItems) canvas.selection.toggleItemMembership(i);
 	
     // Attach mouse focus only to first visible item in selection. For ticket 1098.      
-    for (auto& i: copyOfItems)
+    for (auto& i: canvas.selection.items)
       if (i->visible())
         {
           canvas.setItemFocus(i);
           break;
         }
-        
+                        
     if (!copyOfGroups.empty()) canvas.setItemFocus(copyOfGroups[0]);   
 
     canvas.model->removeGroup(*g);  
--- a/model/minsky.cc
+++ b/model/minsky.cc
@@ -338,8 +338,11 @@ namespace minsky
     auto copyOfItems=g->items;
     auto copyOfGroups=g->groups;
     
-    canvas.model->moveContents(*g);
+    g->resizeOnContents(); // used in insertGroupFromfile(), so why not here?
+    // ungroup g, putting all its contents on the canvas
+    canvas.model->moveContents(*g); 
 
+    // leave newly ungrouped items in selection
     for (auto& i: copyOfItems) canvas.selection.ensureItemInserted(i);
 	
     // Attach mouse focus only to first visible item in selection. For ticket 1098.      
--- a/model/minsky.cc
+++ b/model/minsky.cc
@@ -354,8 +354,10 @@ namespace minsky
        // ensure that initial values of pasted parameters are correct. for ticket 1258
        if (auto v=i->variableCast())
 		 if (v->type()==VariableType::parameter && !existingParms.empty()) 
-		   for (auto& it: existingParms)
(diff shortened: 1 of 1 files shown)


model/minsky.cc
```diff
@@ -288,78 +288,87 @@ namespace minsky
...
-    for (auto& i: copyOfItems)
-      {		
-         canvas.model->addItem(i);			  
-         canvas.selection.ensureItemInserted(i);
-         assert(!i->ioVar());
-      }
-
+    auto copyOfGroups=g->groups;
+    
+    g->resizeOnContents(); // used in insertGroupFromfile(), so why not here?
```
highperformancecoder, line 345: Because this just changes the physical layout of the group, but that aspect is not used later on

model/minsky.cc
```diff
@@ -288,78 +288,87 @@ namespace minsky
...
+    // ungroup g, putting all its contents on the canvas
+    canvas.model->moveContents(*g); 
+
+    // leave newly ungrouped items in selection
+    for (auto& i: copyOfItems) {
+       canvas.selection.ensureItemInserted(i);
+       // ensure that initial values of pasted parameters are correct. for ticket 1258
+       if (auto v=i->variableCast())
+		 if (v->type()==VariableType::parameter && !existingParms.empty()) 
+		   for (vector<pair<string,string>>::iterator it = existingParms.begin() ; it != existingParms.end(); ++it)
```
highperformancecoder, line 355: Why wouldn't you use a std::map for this purpose?

model/minsky.cc
```diff
@@ -288,78 +288,87 @@ namespace minsky
     GroupPtr g(new Group);
     g->self=g;
     m.populateGroup(*g);
+    // stash values of parameters in copied group, as they are reset for some unknown reason later on. for ticket 1258
+    vector<pair<string,string>> existingParms; 
+    for (auto& i: g->items) 
+        if (i->variableCast() && i->variableCast()->type()==VariableType::parameter) existingParms.push_back(make_pair(i->variableCast()->valueId(),i->variableCast()->init()));
```
highperformancecoder, line 294: i->variableCast() is not a trivial call - there is no need to call it 4 times.

model/minsky.cc
wdednam: Very good question. I was considering it last night, but because I stumbled across the solution so late, I decided to commit it. Changed it now :)

model/minsky.cc
highperformancecoder: Use if (auto v=i->variableCast())
          if (v->type() ... v->valueId(), v->init())
pattern

wdednam: Because "auto v=i->variableCast()" does not seem to always work well, I'm trying "VariablePtr v=i->variableCast()". As far as I've read this should be fine, because we know that the type variableCast() returns has to be derived from VariableBase?

highperformancecoder:
Two things bother me: 1) we haven't worked out why the init values are being reset - this solution basically papers over the problem; and 2) the use of a vector of pairs is the worst data structure to use - lookups are O(n), ie this algorithm runs in O(n^2) where n is the number of parameters. std::map is probably the simplest data structure to use, which has O(log n) complexity, but occasionally std::unordered_map is better (average O(1) lookup, worst case O(n), and a large multiplying constant due to the use of a hashing function.

wdednam:
`1) we haven't worked out why the init values are being reset - this solution basically papers over the problem; `

That's true. What I don't understand is why 

for (auto& i: g->items)
   if (auto v=i->variableCast()) cout << v->init() << endl;

returns zero for the parameters sometimes (when the bug occurs).

But  

for (auto& i: g->items)
   if (i->variableCast()) cout << i->variableCast()->init() << endl;

does not?

`2) the use of a vector of pairs is the worst data structure to use - lookups are O(n), ie this algorithm runs in O(n^2) where n is the number of parameters. std::map is probably the simplest data structure to use, which has O(log n) complexity, but occasionally std::unordered_map is better (average O(1) lookup, worst case O(n), and a large multiplying constant due to the use of a hashing function.`      

Thanks, I've just changed it now.

model/minsky.cc
highperformancecoder: Not why "auto v=" is a problem - we use it elsewhere just fine. But variableCast returns a VariableBase*, so that is what you should use if you go that route, not VariablePtr, which would completely muck up ownership.

wdednam: > Use if (auto v=i->variableCast())
> if (v->type() ... v->valueId(), v->init())
> pattern

It's the "if (auto ...)" that is the problem. If I instead do

auto v=i->variableCast()
if (v)
...

then it still works. I'll commit the updated code now.

wdednam: > Not sure why "auto v=" is a problem - we use it elsewhere just fine. But variableCast returns a VariableBase*, so that is what you should use if you go that route, not VariablePtr, which would completely muck up ownership.

Thank you for explaining that. I quickly ran into trouble (segfault) when I did that and tested copying and pasting in 1Free.mky.

highperformancecoder:
>>for (auto& i: g->items)
if (auto v=i->variableCast()) cout << v->init() << endl;

returns zero for the parameters sometimes (when the bug occurs).

But

for (auto& i: g->items)
if (i->variableCast()) cout << i->variableCast()->init() << endl;

does not?
>>

That is strange, for sure. I can't think of an explanation.

As for the original problem, I'd focus on VariableBase::ensureValueExists.

Maybe we can go with this suggested solution for now, for Steve's "Graeber release" but leave the ticket open, as it deserves further investigation.

model/minsky.cc
```diff
@@ -352,8 +354,8 @@ namespace minsky
        // ensure that initial values of pasted parameters are correct. for ticket 1258
        if (auto v=i->variableCast())
 		 if (v->type()==VariableType::parameter && !existingParms.empty()) 
-		   for (vector<pair<string,string>>::iterator it = existingParms.begin() ; it != existingParms.end(); ++it)
-		      if (v->valueId()==(*it).first) v->init((*it).second);
+		   for (auto& it: existingParms)
```
highperformancecoder, line 357: Please use "auto it=existingParms.find(v->valueId(); if (it!=existingParms.end()) ..."

wdednam: Done :)

wdednam:
`As for the original problem, I'd focus on VariableBase::ensureValueExists.`

Do you mean right at the top, before any pasting gets done, within the loop where I populate the map?

That's a very good suggestion that hadn't occurred to me. I tried it in the g->recursiveDo() after the defined vars had been dealt with and outside their conditional statement, but of course, as the values are zeroed when the group is populated at the top of paste(), it wouldn't have helped.

highperformancecoder:
Oh no - Travis had gone red. I'll see if I can fix the broken build to expedite this PR.

PR #221, opened 2020-09-16
fix scoping error in unittests

wdednam:
This change is [Reviewable](https://reviewable.io/reviews/highperformancecoder/minsky/221)

### DIFF
--- a/engine/variableValue.cc
+++ b/engine/variableValue.cc
@@ -232,7 +232,8 @@ namespace minsky
   int VariableValue::scope(const std::string& name) 
   {
     boost::smatch m;
-    if (boost::regex_search(utf_to_utf<char>(name), m, boost::regex(R"((\d*)]?:.*)")))
+    auto nm=utf_to_utf<char>(name);
+    if (boost::regex_search(nm, m, boost::regex(R"((\d*)]?:.*)")))
       if (m.size()>1 && m[1].matched && !m[1].str().empty())
         {
           int r;


engine/variableValue.cc
```diff
@@ -232,7 +232,8 @@ namespace minsky
   int VariableValue::scope(const std::string& name) 
   {
     boost::smatch m;
-    if (boost::regex_search(utf_to_utf<char>(name), m, boost::regex(R"((\d*)]?:.*)")))
+    auto nm=utf_to_utf<char>(name);
+    if (boost::regex_search(nm, m, boost::regex(R"((\d*)]?:.*)")))
```
highperformancecoder, line 236: How does this change fix the problem? The code seems equivalent to the old code...
Note this has been an intermittent problem for a while, so keen on an explanation.

wdednam: I don't understand why defining variables in an if statement, e..g, if (auto...) or calling more complex functions like boost::regex_search() in an if condition causes intermittent errors. Is it perhaps because different threads are spawned in these cases?

highperformancecoder: No

highperformancecoder: An interesting thing: I switched from using boost::regex to std::regex, and the compiler actually forced me to make the change you suggest here. It appears that return value optimisation (aka copy elision) somehow works incorrectly with regex - perhaps pointers are being kept that are invalidated. I don't have a full explanation for what's going on.

The only reason why there is so much boost::regex stuff in the Minsky codebase is that there used to be buggy std::regex implementations in gcc, but that hasn't been true for some time. It was needed while we need to support Ubuntu 14.04, but since we no longer do, perhaps a nice little refactoring task is to replace all usage of boost::regex with std::regex.

wdednam: > An interesting thing: I switched from using boost::regex to std::regex, and the compiler actually forced me to make the change you suggest here. It appears that return value optimisation (aka copy elision) somehow works incorrectly with regex - perhaps pointers are being kept that are invalidated. I don't have a full explanation for what's going on.

It is a helpful explanation, thank you :)

> The only reason why there is so much boost::regex stuff in the Minsky codebase is that there used to be buggy std::regex implementations in gcc, but that hasn't been true for some time. It was needed while we need to support Ubuntu 14.04, but since we no longer do, perhaps a nice little refactoring task is to replace all usage of boost::regex with std::regex.

Consider it done.

PR #222, opened 2020-09-16
inner product between tenors of rank 2 working, trying to get it to w…

wdednam:
…ork between rank 2 and rank 1, or rank 1 and rank 1, tensors too. I'm sure implementation can still be greatly improved

---
This change is [Reviewable](https://reviewable.io/reviews/highperformancecoder/minsky/222)


highperformancecoder:
I'm sure you're aware of the flaws here. It doesn't deal with sparsity. Even if the initial data is dense, you generally lose whatever cache advantage you get from level 3 blas for the overheads of allocating memory and copying the data to and fro.

engine/minskyTensorOps.cc
```diff
@@ -276,9 +280,59 @@ namespace minsky
...
+     if (cachedResult.size()==0) 
+        for (size_t i=0; i<arg1->size(); i++) 
+           cachedResult[i]=nan("");
     }
     Timestamp timestamp() const override {return max(arg1->timestamp(), arg2->timestamp());}
+    void setArguments(const TensorPtr& a1, const TensorPtr& a2) override {
+      arg1=a1; arg2=a2;
+      if (arg1 && arg1->rank()!=0 && arg2 && arg2->rank()!=0) {
+        if (arg1->hypercube().dims()[arg1->rank()-1]!=arg2->hypercube().dims()[0])
+          throw std::runtime_error("arguments not conformal");
```
highperformancecoder, line 329: Conformality is not required for inner product. "Incommensurate shape" might be a better term?

engine/minskyTensorOps.cc
```diff
@@ -276,9 +280,59 @@ namespace minsky
...
+        
+        // ensure size of final tensor is the same as the second argument, which can be rank 1 or 2  
+ 		cachedResult.hypercube(arg2->hypercube());
```
highperformancecoder, line 332: The resultant hypercube has shape [i1,i2,...i{n-1},j2,j3,...j{m-1}], where ii are the dimensions of the first tensor and ji the dimensions of the second, n the rank of the first, and m the rank of the second. Not [j1,...jm] as you have here.

highperformancecoder:
My question is - how do we specify which dimensions to sum over for the first and second tensors? Or do we always do last and first (like matrix multiplication), and provide extra operations for permuting the tensor indices?

engine/minskyTensorOps.cc
wdednam: Thanks, I've changed that :)

engine/minskyTensorOps.cc
wdednam: I realized this yesterday when I was trying to get the matrix dot vector operation to work. I need to resize the hypercube to be a concatenation of the outer dimensions of the two tensors that comprise the inner product. I'll go and look at the civita class and the hypercube struct, in particular, to see how best to accomplish this.

wdednam:
> My question is - how do we specify which dimensions to sum over for the first and second tensors? Or do we always do last and first (like matrix multiplication), and provide extra operations for permuting the tensor indices

In the case of rank 2 tensors I would use the matrix multiplication convention, but in the case of higher rank tensors it becomes more complicated (from Taha Sochi's "An introduction to tensor calculus"):

> For tensors whose outer product produces a tensor of rank > 2, various contraction operations between different sets of indices can occur and hence more than one inner product, which are different in general, can be defined. Moreover, when the outer product produces a tensor of rank > 3 more than one contraction can take place simultaneously. • 

Maybe we should just do rank 2 inner products and let the user choose the two dimensions they would like to multiply in the case of higher order ones?

wdednam:
> I'm sure you're aware of the flaws here. It doesn't deal with sparsity. Even if the initial data is dense, you generally lose whatever cache advantage you get from level 3 blas for the overheads of allocating memory and copying the data to and fro.

To deal with sparsity, I've found the following submodule: [https://github.com/uestla/Sparse-Matrix](https://github.com/uestla/Sparse-Matrix)

I could not get it to compile as a submodule, however. So, i just copied the (three) main files into a new folder inside the minsky top directory. i could get it to compile and expose the methods in the SparseMatrix class to minskyTensorOps.cc. There are a few kinks to iron out with the indices, which run from 1 to the size of the matrix and not starting from 0, but it's lighter than boost::ublas and we could probably modify it to suit our purposes since it has the MIT license.

engine/minskyTensorOps.cc
wdednam: Seems like I reverted to "non-conformal" by accident fixing that now...

engine/minskyTensorOps.cc
```diff
@@ -276,9 +276,67 @@ namespace minsky
...
+      double tmpSum;
+      for (size_t i=0; i< m; i++) {
+        if (arg1->rank()>1 && arg2->rank()>1)		
+          for (size_t j=0; j< n; j++)
+            {
+              tmpSum=0;
+              for (size_t k=0; k<stride; k++)  {
+				auto v1=arg1->atHCIndex(i*stride+k);  
+				auto v2=arg2->atHCIndex(k*stride+j);  
+                tmpSum+=(isnan(v1)? 0 : v1) * (isnan(v2)? 0 : v2);
```
highperformancecoder, line 294: May be marginally faster to do if (!isnan(v1) && !isnan(v2) tmpSum+=v1*v2;

engine/minskyTensorOps.cc
```diff
@@ -276,9 +276,67 @@ namespace minsky
   {
     std::shared_ptr<ITensor> arg1, arg2;
     void computeTensor() const override {//TODO
-      throw runtime_error("inner product not yet implemented");
+		
+      if (arg1->rank()>2 || arg2->rank()>2)
+        throw runtime_error("inner product of rank>2 tensors not yet implemented");
+   		 
+      size_t stride=arg2->hypercube().dims()[0];
```
highperformancecoder, line 283: Maybe check that arg1->hypercube().xvector[arg2->rank()-1].size()==stride

engine/minskyTensorOps.cc
```diff
@@ -276,9 +276,67 @@ namespace minsky
...
+      size_t stride=arg2->hypercube().dims()[0];	
+      size_t m=arg1->hypercube().dims()[0], n=arg2->hypercube().dims()[arg2->rank()-1];	 
+      double tmpSum;
+      for (size_t i=0; i< m; i++) {
+        if (arg1->rank()>1 && arg2->rank()>1)		
+          for (size_t j=0; j< n; j++)
+            {
+              tmpSum=0;
+              for (size_t k=0; k<stride; k++)  {
+				auto v1=arg1->atHCIndex(i*stride+k);
```
highperformancecoder, line 292: Should be k*m+i

engine/minskyTensorOps.cc
```diff
@@ -276,9 +276,67 @@ namespace minsky
...
+      size_t m=arg1->hypercube().dims()[0], n=arg2->hypercube().dims()[arg2->rank()-1];	 
+      double tmpSum;
+      for (size_t i=0; i< m; i++) {
+        if (arg1->rank()>1 && arg2->rank()>1)		
+          for (size_t j=0; j< n; j++)
+            {
+              tmpSum=0;
+              for (size_t k=0; k<stride; k++)  {
+				auto v1=arg1->atHCIndex(i*stride+k);  
+				auto v2=arg2->atHCIndex(k*stride+j);
```
highperformancecoder, line 293: Should be j*stride + k

engine/minskyTensorOps.cc
```diff
@@ -276,9 +276,67 @@ namespace minsky
...
+                tmpSum+=(isnan(v1)? 0 : v1) * (isnan(v2)? 0 : v2);
+			   }
+              cachedResult[j+i*stride]=tmpSum;
+            }
+        else {
+          for (size_t k=0; k<stride; k++)  {
+			auto v1=(arg1->rank()>1? arg1->atHCIndex(i*stride+k):(*arg1)[i]);   
+			auto v2=(*arg2)[k];
+            cachedResult[i]+= (isnan(v1)? 0 : v1)*(isnan(v2)? 0 : v2);
+		  }
```
highperformancecoder, line 303: Is there a significant performance advantage in having special case code for rank 1 arguments? Might be worth profiling.

engine/minskyTensorOps.cc
```diff
@@ -276,9 +276,67 @@ namespace minsky
...
+        else if (arg1->rank()>1 && arg2->rank()<2) { //case mxn dot nx1
+          vector<XVector>&& outerXVs{arg2->hypercube().xvectors[0]};
+          cachedResult.hypercube(outerXVs);
+        }
+        else {
+			//cachedResult.hypercube({});  // causes segfault: 0x00005555556d3926 in boost::any::~any (this=0x55555723f850, __in_chrg=<optimized out>) at /usr/local/include/boost/any.hpp:79 delete content; 
+			
+			// also causes the above segfault
+            auto hc=arg2->hypercube();
+            if (rank()==0) return;
```
highperformancecoder, line 332: hypercube has not been set yet, so rank(0 will always return 0 at this point.

engine/minskyTensorOps.cc
```diff
@@ -276,9 +276,67 @@ namespace minsky
...
+            
+            auto& xv=hc.xvectors[0];
```
highperformancecoder, line 334: Try the following:
auto xv1=arg1->hypercube().xvectors, xv2=arg2->hypercube().xvectors;
Hypercube hc;
hc.xvectors.insert(hc.xvectors.begin(), xv1.begin(), xv1.end()-1);
hc.xvectors.insert(hc.xvectors.begin(), xv2.begin()+1, xv2.end());

highperformancecoder: Then no need for the above special cases.

highperformancecoder: Of course, this densifies the result, so need to think about generating index arrays too.

engine/minskyTensorOps.cc
```diff
@@ -276,9 +276,67 @@ namespace minsky
...
+        if (arg1->rank()>1 && arg2->rank()>1)		
+          for (size_t j=0; j< n; j++)
+            {
+              tmpSum=0;
+              for (size_t k=0; k<stride; k++)  {
+				auto v1=arg1->atHCIndex(i*stride+k);  
+				auto v2=arg2->atHCIndex(k*stride+j);  
+                tmpSum+=(isnan(v1)? 0 : v1) * (isnan(v2)? 0 : v2);
+			   }
+              cachedResult[j+i*stride]=tmpSum;
```
highperformancecoder, line 296: i+m*j

engine/minskyTensorOps.cc
```diff
@@ -276,9 +276,67 @@ namespace minsky
   {
     std::shared_ptr<ITensor> arg1, arg2;
     void computeTensor() const override {//TODO
-      throw runtime_error("inner product not yet implemented");
+		
+      if (arg1->rank()>2 || arg2->rank()>2)
+        throw runtime_error("inner product of rank>2 tensors not yet implemented");
+   		 
+      size_t stride=arg2->hypercube().dims()[0];	
+      size_t m=arg1->hypercube().dims()[0], n=arg2->hypercube().dims()[arg2->rank()-1];
```
highperformancecoder, line 284: In general, m should be the product of dimensions 0..arg1->rank()-2, and n should be the product of dimensions 1..arg2->rank()-1

highperformancecoder: That way, you'll be able to remove the test on line 280.

engine/minskyTensorOps.cc
wdednam: > Might be worth profiling to see if the extra code complexity is worth it.

I remember we discussed profiling in February, but I don't remember which one you use or recommend for optimizing the code?

highperformancecoder: I use gperftools. Do make CPUPROFILE=1 to enable (just the link step is needed, so rather than building everything, just delete the minsky executable).

then run minsky with
env CPUPROFILE=prof.out gui-tk/minsky

The run "pprof --callgrind gui-tk/minsky prof.out >prof.cg"

Finally run "kcachegrind prof.cg"

You'll need to install the relevant packages for these tools, of course.

wdednam: Thank you :)

wdednam: On my machine pprof is google-pprof. is that correct? pprof seems to be something else, a part of a package called "tau".

I also had to replace -lprofiler in Minsky's Makefile with "LIBS+=-L/usr/lib/x86_64-linux-gnu/libprofiler" because as usual, Ubuntu has stuff in non-standard places...

So, maybe that's the wrong libprofiler.so because I only generate empty ".tmp.minsky.*.prof.out" files in the new "pprof" folder under my /home directory?

I've made all the other changes you suggested and the segfault related to destroying a non-existent boost::any pointer is gone, thank you :) I'll commit the code now.

wdednam:
I don't understand why compareWithOctave.sh fails online, it passes locally:
compareWithOctavePasses

wdednam:
Ran the test in Minsky's top directory and it passes, although I get some strange messages:
compareWithOctavePasses2

engine/minskyTensorOps.cc
```diff
@@ -275,32 +275,37 @@ namespace minsky
...
         else {
           for (size_t k=0; k<stride; k++)  {
-			auto v1=(arg1->rank()>1? arg1->atHCIndex(i*stride+k):(*arg1)[i]);   
-			auto v2=(*arg2)[k];
-            cachedResult[i]+= (isnan(v1)? 0 : v1)*(isnan(v2)? 0 : v2);
-		  }
+            auto v1=(arg1->rank()>1? arg1->atHCIndex(k*m+i):(*arg1)[i]);   
+            auto v2=(*arg2)[k];
+            if (!isnan(v1) && !isnan(v2)) cachedResult[i]+= v1*v2;
+          }
```
highperformancecoder, line 308: This section of code doesn't handle m=1, n>1 situation.

engine/minskyTensorOps.cc
```diff
@@ -315,26 +320,23 @@ namespace minsky
...
+        cachedResult.hypercube(move(hc));
+        
+        auto idx1=arg1->index(), idx2=arg2->index();
+        set<size_t> newIdx;        
+        for (auto& i: idx1)
+          for (auto& j: idx2)
+          {
+            // strip of any indices outside the output range
+            auto t=ssize_t(j)-ssize_t(i);
+            if (t>=0 && t<ssize_t(size()))  newIdx.insert(t);
```
highperformancecoder, line 336: This doesn't make sense

engine/minskyTensorOps.cc
```diff
@@ -315,26 +320,23 @@ namespace minsky
...
+          }
+          
+        cachedResult.index(Index(newIdx));
```
highperformancecoder, line 339: For now, computeTensor is not sparse aware, so the only way this can be consistent is to not set the index of cachedResult. I suggest you commit this "densifying" inner product for now (once you fix the above algorithm for n>1), and leave correct sparse handling for a later PR.

model/parVarSheet.cc
```diff
@@ -133,11 +133,11 @@ namespace minsky
 			    else
 			      {
                     cairo_move_to(cairo,x,y-1.5*rowHeight);
-                    pango.setMarkup(value->name+":");
+                    pango.setMarkup(latexToPango(value->name)+":");
                     pango.show();  					    
                     for (size_t k=0; k<rank-1; k++)  
                       {   
-                        y+=rowHeight; // allow room for header row
+                        if (k%2==0) y+=rowHeight; // allow room for header row
```
highperformancecoder, line 140: Just so you know - % is a relatively expensive operation (think long division), and modulo powers of two can be done via much cheaper bit masks. x%2==0 is equivalent to (x&1)==0. Be careful with operator precedence though when using these operators - it tends to be counterintuitive, so liberal use of parentheses is advisable. 

I don't know how crucial performance is in this bit of code.

engine/minskyTensorOps.cc
wdednam: The index() members of both arg1 and arg2 are empty in setArguments(), so I don't think I should be calculating it here, but rather in computeTensor()? In the dense case, the index is filled with the correct indices of the finite elements in the final matrix, but the index's size exceeds the number of finite elements. So, I guess the index() would be correct if it appeared filled up to only the last non-blank cell? (see attached screenshot
Index
)

engine/minskyTensorOps.cc
wdednam: > For now, computeTensor is not sparse aware, so the only way this can be consistent is to not set the index of cachedResult

Okay, this answers my question in the previous reply. Thanks.

model/parVarSheet.cc
wdednam: Thanks, I've replaced the 3 instances I added with this suggestion. There's still room for improvement in this feature, especially for higher rank parameters, but it will have to wait until I've sorted out bug 1258 that Steve raised recently.

engine/minskyTensorOps.cc
```diff
@@ -288,26 +288,18 @@ namespace minsky
...
-                auto v1=arg1->atHCIndex(k*m+i);  
-                auto v2=arg2->atHCIndex(j*stride + k);  
+      for (size_t i=0; i< m; i++)
+        for (size_t j=0; j< n; j++)
+          {
+            tmpSum=0;
+            for (size_t k=0; k<stride; k++)  
+              {
+                auto v1=m>1? arg1->atHCIndex(k*m+i) : (*arg1)[k];  
+                auto v2=n>1? arg2->atHCIndex(j*stride + k) : (*arg2)[k];
```
highperformancecoder, line 298: The ternary op seems like premature optimisation to me, but logic is correct.

PR #223, opened 2020-09-18
replaced boost::regex with std::regex, also smatch and cmatch

wdednam:
This change is [Reviewable](https://reviewable.io/reviews/highperformancecoder/minsky/223)

### DIFF
--- a/engine/derivative.cc
+++ b/engine/derivative.cc
@@ -23,10 +23,10 @@
 #include "minsky.h"
 #include "expr.h"
 #include "minsky_epilogue.h"
-#include <boost/regex.hpp>
+#include <regex> 
 
-using boost::regex;
-using boost::smatch;
+using std::regex;
+using std::smatch;
 using namespace minsky;
 
 namespace MathDAG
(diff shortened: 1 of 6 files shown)


wdednam:
I don't understand, I get this error for a line that does not exist in plotWdiget.cc on my machine:

model/plotWidget.cc: In member function 'void minsky::PlotWidget::addConstantCurves()':

model/plotWidget.cc:546:29: error: comparison of integer expressions of different signedness: 'long long int' and 'const size_t' {aka 'const long unsigned int'} [-Werror=sign-compare]

  546 |                 if (div.quot<maxNumTensorElementsToPlot)

      |                     ~~~~~~~~^~~~~~~~~~~~~~~~~~~~~~~~~~~

highperformancecoder:
Sorry about that - I broke the build in my rush to get a new version of Ravel out for Steve. It should be fixed now, so merge master again, and commit.

wdednam:
> Sorry about that - I broke the build in my rush to get a new version of Ravel out for Steve. It should be fixed now, so merge master again, and commit.

Ah okay, no worries. I thought I had broken something somehow...

PR #225, opened 2020-09-22
Bugfix 1256

wdednam:
This change is [Reviewable](https://reviewable.io/reviews/highperformancecoder/minsky/225)


engine/equationDisplay.cc
```diff
@@ -168,8 +168,9 @@ namespace MathDAG
       {
         if (eqnNo++ < baseEqn) continue;
         // initial conditions
+        cairo_move_to(dest.cairo(), x, y); // needed to define a current point on the equations tab. for ticket 1256
```
highperformancecoder, line 171: I don't understand how this fixes things. If the loop starting on line 153 runs through without line 163 being executed, then x and y still have their original values  as set on line 144. But then this is just the coordinates of the current point, so setting the current point to that value does nothing. OTOH, if there is no current point on entering this method, then presumably x and y are undefined, so calling this method with undefined value presumably just makes things worse.

wdednam: > I don't understand how this fixes things. If the loop starting on line 153 runs through without line 163 being executed, then x and y still have their original values as set on line 144. But then this is just the coordinates of the current point, so setting the current point to that value does nothing.

It [redefines](https://www.cairographics.org/manual/cairo-Paths.html#cairo-get-current-point) the current_point as being at x and y, If x and y are undefined, the they are both set to 0:

cairo_move_to ()

void
cairo_move_to (cairo_t *cr,
               double x,
               double y);

Begin a new sub-path. After this call the current point will be (x , y ).

> OTOH, if there is no current point on entering this method, then presumably x and y are undefined, so calling this method with undefined value presumably just makes things worse.

It may not be the most appropriate function to redefine the current point when there isn't one, but it solves the problem by making the sure _a_ current point exists. I will try some of the others discussed at the above link. We can also use cairo_has_current_point () to check for undefined current points, but I guess in that case nothing will be displayed on the equations tab. Maybe I could do the following check that will define a current point at x=0 and y=0?

if (!cairo_has_current_point (dest.cairo())) cairo_show_text (dest.cairo(),NULL)

highperformancecoder: Fair enough. I hadn't read the fine print on cairo_get_current_point. I suppose we could initialise x and y to 0 - it wouldn't do anything, except I suppose make clear that in the undefined situation, x & y are zero.

wdednam: > if (!cairo_has_current_point (dest.cairo())) cairo_show_text (dest.cairo(),NULL)

This doesn't work...

wdednam: > Fair enough. I hadn't read the fine print on cairo_get_current_point. I suppose we could initialise x and y to 0 - it wouldn't do anything, except I suppose make clear that in the undefined situation, x & y are zero.

Cairo's documentation doesn't invite one to read the fine print. I always find it very confusing/disorienting when I go to check anything there. There is enough information, but the order and emphasis of the presentation is not user friendly in my opinion.

PR #227, opened 2020-09-28
textinput containing positve and negative numbers creates constants o…

wdednam:
…n canvas, negative sign only in textinput and immediately pressing return creates minus op on canvas

---
This change is [Reviewable](https://reviewable.io/reviews/highperformancecoder/minsky/227)

### DIFF
--- a/gui-tk/wiring.tcl
+++ b/gui-tk/wiring.tcl
@@ -412,16 +412,24 @@ proc textOK {} {
     canvas.moveOffsX 0
     canvas.moveOffsY 0
     if {[lsearch [availableOperations] $textBuffer]>-1} {
-        addOperationKey $textBuffer
+		addOperationKey $textBuffer
+	} elseif {[llength $textBuffer]==1 && [string match "-" $textBuffer]} { # minus sign only creates subtract op. for ticket 145
+		addOperationKey subtract		
     } elseif [string match "\[%#\]*" $textBuffer] {
         addNote [string range $textBuffer 1 end]
     } else {
-        if [regexp "(.*)=(.*)" $textBuffer dummy name init] {            
-            minsky.addVariable $name flow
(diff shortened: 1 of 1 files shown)


gui-tk/wiring.tcl
```diff
@@ -412,16 +412,24 @@ proc textOK {} {
...
-        if [regexp "(.*)=(.*)" $textBuffer dummy name init] {            
-            minsky.addVariable $name flow
+        if [regexp "(.*)=(.*)" $textBuffer dummy name init] {
+			minsky.addVariable $name flow
 			minsky.canvas.itemFocus.init $init
             minsky.variableValues.reset
-        } else {
-            minsky.addVariable $textBuffer flow
+        } else {  # signed numbers create constant on the canvas. for ticket 145
+			if [regexp "(?:^|\s)(\[+-\]?\[\[:digit:\]\]*\.?\[\[:digit:\]\]+)(?=\$|\s)" $textBuffer] { #https://stackoverflow.com/questions/33520934/regex-with-only-numbers-in-a-string-c
```
highperformancecoder, line 426: Make this elseif to reduce the level of nesting

highperformancecoder: Did you really want to match +1.3 here?

gui-tk/wiring.tcl
```diff
@@ -432,7 +440,7 @@ proc textOK {} {
 
 # operation add shortcuts
 bind . <Key-plus> {addOperationKey add}
-bind . <Key-minus> {addOperationKey subtract}
+bind . <Key-minus> {textInput "-"; .wiring.canvas configure -cursor {}}
```
highperformancecoder, line 443: This line is probably redundant now, as the <Key> event will match

highperformancecoder:
A few minor code improvement suggestions above, but nothing that should affect the efficacy of the code.

PR #228, opened 2020-09-29
possible fix for parameter slider bug which is part of ticket 1258

wdednam:
This change is [Reviewable](https://reviewable.io/reviews/highperformancecoder/minsky/228)


highperformancecoder:
I'll take your word for it that this does improve things.

wdednam:
> I'll take your word for it that this does improve things.

It does seem better, but I'm not going to close the ticket. I've seen this problem more often when the model is not at initial zoom. After testing it thoroughly, I can't reproduce the problem, even when I expect it to occur.

PR #231, opened 2020-10-05
this change seems sufficient, but not sure if there is a better approach

wdednam:
This change is [Reviewable](https://reviewable.io/reviews/highperformancecoder/minsky/231)

### DIFF
--- a/model/variable.cc
+++ b/model/variable.cc
@@ -503,6 +503,8 @@ void VariableBase::draw(cairo_t *cairo) const
     try
       {
         auto val=engExp();
+        
+        if ((sliderMax-sliderMin)/sliderStep > 1.0e04) sliderStep=(sliderMax-sliderMin)/1.0e04;   // ensure there are at most 10000 steps between sliderMin and Max. for ticket 1255.        
   
         Pango pangoVal(cairo);
         if (!isnan(value())) {
--- a/model/variable.cc
+++ b/model/variable.cc
@@ -423,6 +423,7 @@ void VariableBase::sliderSet(double x)
 {
   if (x<sliderMin) x=sliderMin;
   if (x>sliderMax) x=sliderMax;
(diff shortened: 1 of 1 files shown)


model/variable.cc
```diff
@@ -503,6 +503,8 @@ void VariableBase::draw(cairo_t *cairo) const
     try
       {
         auto val=engExp();
+        
+        if ((sliderMax-sliderMin)/sliderStep > 1.0e04) sliderStep=(sliderMax-sliderMin)/1.0e04;   // ensure there are at most 10000 steps between sliderMin and Max. for ticket 1255.
```
highperformancecoder, line 507: Probably better to do this in VariableBase::adjustSliderBounds(), VariableBase::sliderSet() and initSliderBounds()

wdednam: Thank you. I wasn't sure. I wanted to add the fewest number of lines, and ended up settling for this after trying various options.

model/variable.cc
```diff
@@ -502,7 +505,9 @@ void VariableBase::draw(cairo_t *cairo) const
   if (type()!=constant && !ioVar() && (vv.size()==1) )
     try
       {
-        auto val=engExp();
+        auto val=engExp();    
+        
+        double tmpSliderStep=(sliderMax-sliderMin)/sliderStep > 1.0e04 ? (sliderMax-sliderMin)/1.0e04 : sliderStep;
```
highperformancecoder, line 510: Did this prove to be necessary>

model/variable.cc
```diff
@@ -423,6 +423,7 @@ void VariableBase::sliderSet(double x)
 {
   if (x<sliderMin) x=sliderMin;
   if (x>sliderMax) x=sliderMax;
+  if ((sliderMax-sliderMin)/sliderStep > 1.0e04) sliderStep=(sliderMax-sliderMin)/1.0e04;   // ensure there are at most 10000 steps between sliderMin and Max. for ticket 1255.
```
highperformancecoder, line 426: Probably worth extracting this calculation into a separate function/method, rather than repeating 3 times.

PR #232, opened 2020-10-08
implemented single method to replace code that limits the number of s…

wdednam:
…lider steps to 10000

---
This change is [Reviewable](https://reviewable.io/reviews/highperformancecoder/minsky/232)

### DIFF
--- a/model/variable.cc
+++ b/model/variable.cc
@@ -423,7 +423,7 @@ void VariableBase::sliderSet(double x)
 {
   if (x<sliderMin) x=sliderMin;
   if (x>sliderMax) x=sliderMax;
-  if ((sliderMax-sliderMin)/sliderStep > 1.0e04) sliderStep=(sliderMax-sliderMin)/1.0e04;   // ensure there are at most 10000 steps between sliderMin and Max. for ticket 1255.     
+  sliderStep=maxSliderSteps();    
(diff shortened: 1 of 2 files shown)


model/variable.cc
```diff
@@ -459,10 +459,17 @@ void VariableBase::adjustSliderBounds() const
...
         if (sliderMin>vv->value()) sliderMin=vv->value();
-        if ((sliderMax-sliderMin)/sliderStep > 1.0e04) sliderStep=(sliderMax-sliderMin)/1.0e04;   // ensure there are at most 10000 steps between sliderMin and Max. for ticket 1255.     
+        sliderStep=maxSliderSteps();   
       }
 }
 
+double VariableBase::maxSliderSteps() const
+{
+    // ensure there are at most 10000 steps between sliderMin and Max. for ticket 1255. 	
+	if ((sliderMax-sliderMin)/sliderStep > 1.0e04) sliderStep=(sliderMax-sliderMin)/1.0e04;
```
highperformancecoder, line 469: Kind of OK, but this method does lie about the fact that it mutates sliderStep. Better would be to return (sliderMax-sliderMin)/1.0e04 at this point, rather than setting it. Then it would truly live up to the const attribute. FWIW, I can't remember now why these attributes are mutable.

PR #234, opened 2020-10-14
initial commit for feature 147: element-wise integration of tensor vars

wdednam:
This change is [Reviewable](https://reviewable.io/reviews/highperformancecoder/minsky/234)


highperformancecoder:
Looks good. This code is mergeable.

wdednam:
Thanks :) I just need to make sure the flow vars are correct with the change. The indices looked correct when I printed them out, i.e., they were contiguous and ordered from lowest to highest within each stock and input VariableValue of length 5.

PR #235, opened 2020-10-15
now slider bounds are set by defintion and the user will have no choi…

wdednam:
…ce but to change them

---
This change is [Reviewable](https://reviewable.io/reviews/highperformancecoder/minsky/235)

### DIFF
--- a/model/cairoItems.cc
+++ b/model/cairoItems.cc
@@ -94,7 +94,7 @@ bool RenderVariable::inImage(float x, float y)
 
 double RenderVariable::handlePos() const
 {
-  var.initSliderBounds();
+  if (var.sliderStep<1.0e-100) var.initSliderBounds();   // this should only be used when sliderStep's value has not been set or is a nonsensical value. for tickets 1258/1263
   var.adjustSliderBounds();
   return (w<0.5*var.iWidth()? 0.5*var.iWidth() : w)*(var.value()-0.5*(var.sliderMin+var.sliderMax))/(var.sliderMax-var.sliderMin);
 }
--- a/model/cairoItems.cc
+++ b/model/cairoItems.cc
@@ -94,7 +94,7 @@ bool RenderVariable::inImage(float x, float y)
 
 double RenderVariable::handlePos() const
 {
-  if (var.sliderStep<1.0e-100) var.initSliderBounds();   // this should only be used when sliderStep's value has not been set or is a nonsensical value. for tickets 1258/1263
+  if (var.sliderStep<std::numeric_limits<double>::min() || std::isnan(var.sliderStep)) var.initSliderBounds();   // this should only be used when sliderStep's value has not been set or is a nonsensicl
   var.adjustSliderBounds();
   return (w<0.5*var.iWidth()? 0.5*var.iWidth() : w)*(var.value()-0.5*(var.sliderMin+var.sliderMax))/(var.sliderMax-var.sliderMin);
 }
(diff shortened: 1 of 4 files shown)


model/slider.h
```diff
@@ -32,7 +32,7 @@ namespace minsky
                           [sliderMin,sliderMax] */
 
     mutable double sliderMin, sliderMax, sliderStep;
-    Slider(): sliderBoundsSet(false), sliderStepRel(false) {}
+    Slider(): sliderBoundsSet(true), sliderStepRel(false) {}    // ensure sliderBoundsSet is true upon creation of a variable or parameter. for tickets 1258/1263
```
highperformancecoder, line 35: Is the sliderBoundsSet flag ever falsified?

wdednam: No, it's not. Maybe all this requires is a tweak in wiring.tcl, because initSliderBounds runs regardless of whether or not the slider bounds are set the first time?

model/cairoItems.cc
```diff
@@ -94,7 +94,7 @@ bool RenderVariable::inImage(float x, float y)
 
 double RenderVariable::handlePos() const
 {
-  var.initSliderBounds();
+  if (var.sliderStep<1.0e-100) var.initSliderBounds();   // this should only be used when sliderStep's value has not been set or is a nonsensical value. for tickets 1258/1263
```
highperformancecoder, line 97: Again, the problem with these sorts of hard constants is if var.value ~ 1e-100, then a sliderstep of that value may be appropriate. However 1e-100 probably close to epslion, though, where 1+epsilon == 1 (In IEEE arithmetic), so maybe its appropriate (until it's not - we cannot assume arithemetic will always be IEEE double precision, even though it's 99.9% likely to be so.

model/slider.h
highperformancecoder: I think the original intention of this flag was to track whether the slider bounds were set (originally variable didn't even display the slider). Perhaps now, we insist that they're set at parameter creation to something sensible, then this flag can be removed.

wdednam: I think the original intention of the flag is very useful, which why I've been striving so hard to try to keep it in my code commits. I think I've now come up with something more reasonable than my previous commit. Your comment about epsilon() was the hint I needed. Thanks.

model/cairoItems.cc
wdednam: Epsilon() still seems to be too large for what I had in mind [https://stackoverflow.com/questions/13698927/compare-double-to-zero-using-epsilon](https://stackoverflow.com/questions/13698927/compare-double-to-zero-using-epsilon) 

So, I'm using std::numeric_limits<double>::min() instead, which seems to be a good cutoff value, because most of the time sliderStep is initially either 0 or values like 2.61361e-321.

model/variable.cc
```diff
@@ -455,18 +455,19 @@ void VariableBase::adjustSliderBounds() const
...
+        sliderBoundsSet=true;	                    
       }
 }
 
 double VariableBase::maxSliderSteps() const
 {
     // ensure there are at most 10000 steps between sliderMin and Max. for ticket 1255. 	
-	if ((sliderMax-sliderMin)/sliderStep > 1.0e04) sliderStep=(sliderMax-sliderMin)/1.0e04;    
+	if ((sliderMax-sliderMin)/sliderStep > 1.0e04) return (sliderMax-sliderMin)/1.0e04;    
 	return sliderStep;
```
highperformancecoder, line 471: Unreachable code

PR #236, opened 2020-10-19
added outer product of dense tensors. for feature 39

wdednam:
This change is [Reviewable](https://reviewable.io/reviews/highperformancecoder/minsky/236)

### DIFF
--- a/engine/minskyTensorOps.cc
+++ b/engine/minskyTensorOps.cc
@@ -327,9 +327,38 @@ namespace minsky
   {
     std::shared_ptr<ITensor> arg1, arg2;
     void computeTensor() const override {//TODO
-      throw runtime_error("outer product not yet implemented");
+      //throw runtime_error("outer product not yet implemented");
+      size_t m=1, n=1;   
+      for (size_t i=0; i<arg1->rank(); i++)
+        m*=arg1->hypercube().dims()[i];
+        
+      for (size_t i=0; i<arg2->rank(); i++)
+        n*=arg2->hypercube().dims()[i];     
+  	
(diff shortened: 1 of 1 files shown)


engine/minskyTensorOps.cc
```diff
@@ -327,9 +327,38 @@ namespace minsky
   {
     std::shared_ptr<ITensor> arg1, arg2;
     void computeTensor() const override {//TODO
-      throw runtime_error("outer product not yet implemented");
+      //throw runtime_error("outer product not yet implemented");
```
highperformancecoder, line 330: Dead code, please remove

engine/minskyTensorOps.cc
```diff
@@ -327,9 +327,38 @@ namespace minsky
...
+      size_t m=1, n=1;   
+      for (size_t i=0; i<arg1->rank(); i++)
+        m*=arg1->hypercube().dims()[i];
```
highperformancecoder, line 333: This loop recapitulates arg1->hypercube().numElements()

engine/minskyTensorOps.cc
```diff
@@ -327,9 +327,38 @@ namespace minsky
...
+        
+      for (size_t i=0; i<arg2->rank(); i++)
+        n*=arg2->hypercube().dims()[i];
```
highperformancecoder, line 336: ditto

highperformancecoder:
A couple of minor improvements, but basically mergeable. Obviously, we need to redo it as a sparse algorithm, as this will not fare well with high rank sparse tensors.

PR #238, opened 2020-10-22
added code to add \ to % in var names in node_latex.cc

wdednam:
This change is [Reviewable](https://reviewable.io/reviews/highperformancecoder/minsky/238)

### DIFF
--- a/engine/node_latex.cc
+++ b/engine/node_latex.cc
@@ -44,6 +44,11 @@ namespace MathDAG
     string::size_type ss;
     if ((ss=nm.find_first_of("_^"))!=string::npos)
       return mathrm(nm.substr(0, ss)) + nm[ss] + "{"+mathrm(nm.substr(ss+1))+"}";
+      
+    // process % chars
+    string::size_type pc;
+    if ((pc=nm.find_first_of("%"))!=string::npos)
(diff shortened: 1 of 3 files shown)


engine/node_latex.cc
```diff
@@ -44,6 +44,11 @@ namespace MathDAG
     string::size_type ss;
     if ((ss=nm.find_first_of("_^"))!=string::npos)
       return mathrm(nm.substr(0, ss)) + nm[ss] + "{"+mathrm(nm.substr(ss+1))+"}";
+      
+    // process % chars
+    string::size_type pc;
+    if ((pc=nm.find_first_of("%"))!=string::npos)
+      return mathrm(nm.substr(0, pc)) + "\\" + nm[pc] + mathrm(nm.substr(pc+1));
```
highperformancecoder, line 51: This only modifies the first % found in the string. Shouldn't you be quoting all of them? regex_replace would be the appropriate tool.

highperformancecoder: I'm becoming suspicious of line 46 too. Was that me?

highperformancecoder, line 51: Forget below. It's recursive, and I didn't see that. All instances are being processed.

PR #239, opened 2020-10-31
Feature 72

wdednam:
This change is [Reviewable](https://reviewable.io/reviews/highperformancecoder/minsky/239)

### DIFF
--- a/model/item.cc
+++ b/model/item.cc
@@ -161,6 +161,8 @@ namespace minsky
  
   bool Item::visible() const 
   {
+	if (auto i=dynamic_cast<const IntOp*>(this))
+	  if (i->intVar->attachedToDefiningVar()) return false; 
 	if (auto o=operationCast()) 
 	  if (o->attachedToDefiningVar()) return false;
 	if (auto v=variableCast()) 
--- a/model/item.cc
+++ b/model/item.cc
@@ -161,13 +161,7 @@ namespace minsky
  
   bool Item::visible() const 
   {
-	// needed in some models, such as fundamentalConstants.mky to hide coupled intop along with other defining variables. for ticket 72/73  	  
-	if (auto i=dynamic_cast<const IntOp*>(this))
-	  if (i->intVar->attachedToDefiningVar()) return false; 
-	if (auto o=operationCast()) 
-	  if (o->attachedToDefiningVar()) return false;
-	if (auto v=variableCast()) 
-	  if (v->attachedToDefiningVar()) return false;	  
+	if (attachedToDefiningVar()) return false;   
     auto g=group.lock();
     return (!g || g->displayContents());
   }
--- a/model/item.cc
+++ b/model/item.cc
@@ -279,6 +279,21 @@ namespace minsky
     cairo_stroke(cairo);
   }
   
+  bool Item::attachedToDefiningVar() const
+  {
+    //if (auto g=group.lock())
+    //  if (any_of(g->items.begin(),g->items.end(), [&](ItemPtr i){if (auto v=i->variableCast()) return v->varTabDisplay; else return false;})) return true;
+      //else return true;
+      //  for (auto& i: g->items) {
+		//   auto v=i->variableCast();	 
+      //     if (v && v->defined() && v->varTabDisplay) return true; 
+	  // }
+	if (ports.size()>1)
+	  for (size_t i=0; i<ports.size(); i++)
+        for (auto w: ports[i]->wires())
+          if (w->attachedToDefiningVar()) return true;
+    return false;
+  }    
   
   // default is just to display the detailed text (ie a "note")
   void Item::draw(cairo_t* cairo) const
--- a/model/item.cc
+++ b/model/item.cc
@@ -161,8 +161,12 @@ namespace minsky
  
   bool Item::visible() const 
   {
+	if (auto o=operationCast()) 
+	  if (o->attachedToDefiningVar()) return false;
+	if (auto v=variableCast()) 
+	  if (v->attachedToDefiningVar()) return false;	  
     auto g=group.lock();
-    return !g || g->displayContents();
+    return (!g || g->displayContents());
   }
   
 
@@ -281,10 +285,9 @@ namespace minsky
   
   bool Item::attachedToDefiningVar() const
   {
-    if (ports.size()>1)
-      for (size_t i=0; i<ports.size(); i++)
-        for (auto w: ports[i]->wires())
-          if (w->attachedToDefiningVar()) return true;
+    for (size_t i=0; i<ports.size(); i++)
+      for (auto w: ports[i]->wires())
+        if (w->attachedToDefiningVar()) return true;
     return false;
   }    
   
--- a/model/item.cc
+++ b/model/item.cc
@@ -161,10 +161,13 @@ namespace minsky
  
   bool Item::visible() const 
   {
-	// needed in some models, such as fundamentalConstants.mky to hide coupled intop along with other defining variables. for tickets 72/73  
+	// needed in some models, such as fundamentalConstants.mky to hide coupled intop along with other defining variables. for ticket 72/73  	  
 	if (auto i=dynamic_cast<const IntOp*>(this))
-	  if (i->intVar->attachedToDefiningVar()) return false;	  
-	if (attachedToDefiningVar()) return false;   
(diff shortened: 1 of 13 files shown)


model/item.cc
```diff
@@ -279,6 +279,14 @@ namespace minsky
     cairo_stroke(cairo);
   }
   
+  bool Item::attachedToDefiningVar() const
+  {
+    if (ports.size()>1)
```
highperformancecoder, line 284: guard clause unnecessary - the following loop with zero iterations is equivalent.

model/variable.h
```diff
@@ -167,6 +167,9 @@ namespace minsky
     bool isStock() const {return type()==stock || type()==integral;}
 
     virtual ~VariableBase();
+    
+    bool varTabDisplay=false;
+    void toggleVarTabDisplay() {varTabDisplay=!varTabDisplay;}
```
highperformancecoder, line 172: You should overrride attachedToDefiningVar() here, and return the value of varTabDisplay. This will simplify the Wire::attachedToDefiningVar() implementation later.

model/wire.cc
```diff
@@ -363,6 +363,19 @@ namespace
...
     cairo_path_destroy (path);              
   }
+  
+  bool Wire::attachedToDefiningVar() const
+  {
+    auto f=from(), t=to();
+    assert(f && t);
+    bool attachedToFrom=false, attachedToTo=false;                
+    if (auto vf=f->item().variableCast()) attachedToFrom=vf->varTabDisplay;
+    //else return f->item().attachedToDefiningVar();                                // attempt at recursion...
```
highperformancecoder, line 373: See my note above. If Variable::attachedToDefiningVar() is an overridden implementation, this implementation can just be the result of to()->attachedToDefiningVar(). I don't think the status of the from() item is relevant to determine visibility on the canvas, only the to() item..

model/variable.cc
```diff
@@ -480,139 +480,141 @@ bool VariableBase::handleArrows(int dir,bool reset)
...
-  // parameters of icon in userspace (unscaled) coordinates
-  float w, h, hoffs, scaleFactor;
-  w=rv.width()*z; 
-  h=rv.height()*z;
-  scaleFactor=max(1.0f,min(0.5f*iWidth()*z/w,0.5f*iHeight()*z/h));
-  if (rv.width()<0.5*iWidth()) w=0.5*iWidth()*z;
-  if (rv.height()<0.5*iHeight()) h=0.5*iHeight()*z;
-  rv.setFontSize(12*scaleFactor*z);
-  hoffs=rv.top()*z;
+  if (!varTabDisplay) {
```
highperformancecoder, line 483: This is not quite right. A variable should only be invisible if varTabDisplay and all outgoing wires are attachedToDefiningVar(). Other items should have their visibility depend on all their outgoing wires too.

It would be reasonable to modify the visible() method for this purpose, as visible() already means whether the item is visible on the canvas.

model/item.cc
wdednam: Thanks, I've removed it.

model/variable.h
wdednam: Done, thanks.  But struggling to get all the items to disappear /reappear when I toggleVarTabDisplay()

model/wire.cc
wdednam: Thank you. Yes, it doesn't make sense to include from() here :)

model/variable.cc
wdednam: I understand and was confused about whether or not to use visible(). I thought it just hides variables when they are inside groups and displayContents is false.

model/item.cc
```diff
@@ -279,6 +283,13 @@ namespace minsky
     cairo_stroke(cairo);
   }
   
+  bool Item::attachedToDefiningVar() const
+  {
+    for (size_t i=0; i<ports.size(); i++)
```
highperformancecoder, line 288: Only port 0 (the output port) is relevant here (except for switches, which have more than one output port, and can be handle via an override).

model/wire.cc
```diff
@@ -363,6 +364,14 @@ namespace
     }
     cairo_path_destroy (path);              
   }
+  
+  bool Wire::attachedToDefiningVar() const
+  {
+    auto t=to();
+    assert(t);             
+    if (auto v=t->item().variableCast()) return v->attachedToDefiningVar();
```
highperformancecoder, line 372: This should still be a call to t->item().attachedToDefiningVar()

model/item.cc
wdednam: Thanks, this change and the next makes all the items attached to the defining disappear when I toggle var tab display :)

I just have to make them come back when I do the reverse, which they don't for some reason and also deal with int ops which seem to crash Minsky (example exponentialGrowth.mky crashes). After that, this feature should be ready to go and I can work on the second part of it.

model/item.cc
```diff
@@ -161,8 +161,14 @@ namespace minsky
  
   bool Item::visible() const 
   {
+	if (auto i=dynamic_cast<const IntOp*>(this))
+	  if (i->intVar->attachedToDefiningVar()) return false;
```
highperformancecoder, line 165: Isn't an intVar a defined variable by definition?

model/item.cc
```diff
@@ -161,8 +161,14 @@ namespace minsky
  
   bool Item::visible() const 
   {
+	if (auto i=dynamic_cast<const IntOp*>(this))
+	  if (i->intVar->attachedToDefiningVar()) return false; 
+	if (auto o=operationCast()) 
+	  if (o->attachedToDefiningVar()) return false;
+	if (auto v=variableCast()) 
+	  if (v->attachedToDefiningVar()) return false;
```
highperformancecoder, line 169: If attachedToDefiningVar() is virtual, which I think it is, then there is no need to do the *Cast() calls.

highperformancecoder: Note pay attention to Switches, which have multiple outputs.

highperformancecoder:
Sorry about my comment on switches - they do indeed have a single output port at ports[0], just like all other items except for GodleyIcons and PlotWidgets. These latter two should never be made invisible, of course.

model/item.cc
wdednam: If i->coupled(), then the intop remains visible, which is why I added this condition.

wdednam:
> Sorry about my comment on switches - they do indeed have a single output port at ports[0], just like all other items except for GodleyIcons and PlotWidgets. These latter two should never be made invisible, of course. That is currently handled, because attachedToDefiningVar() can only be true for operations and variables. I'm not convinced we should include switches in that list, though, so a better condition might be that it has a port[0] and that it is an output port, rather than being an operation or variable.
> 
> Anyway - very close now - I'd merge with those minor fixups mentioned above, and we can work out what needs to change by running the code.

There is still a problem. If I hide a var individually I cannot reveal it again via the background context menu. Working on fixing that now.

wdednam:
I've decided to disable the ability to hide vars individually, because I couldn't think of an obvious way to reveal them on the canvas again.

model/wire.cc
```diff
@@ -363,6 +365,15 @@ namespace
     }
     cairo_path_destroy (path);              
   }
+  
+  bool Wire::attachedToDefiningVar() const
+  {
+    auto t=to();
+    assert(t);             
+    if (auto i=dynamic_cast<IntOp*>(t->item().operationCast())) return i->intVar->varTabDisplay;
```
highperformancecoder, line 373: Could we not override attachedToDefiningVar() in IntOp to implement this code there? That way, this code, and the code in Item::visible() is simplified.

wdednam: I'm busy fixing this now.

model/variableSheet.h
```diff
@@ -28,7 +28,7 @@ namespace minsky
   class VariableSheet: public ParVarSheet
   {	  
   public:
-    bool variableSelector(ItemPtr i) override {return i->variableCast() && i->variableCast()->type()!=VariableType::parameter;}
+    bool variableSelector(ItemPtr i) override {return i->variableCast() && i->variableCast()->type()!=VariableType::parameter && i->variableCast()->attachedToDefiningVar();}
```
highperformancecoder, line 31: I prefer
if (auto v=i->variableCast())
   return v->type()!=VariableType::parameter && v->attachedToDefiningVariable();
return false;
It eliminates 2 unnecessary calls to variableCast()

PR #242, opened 2020-11-04
proposed fix for ticket 1268

wdednam:
This change is [Reviewable](https://reviewable.io/reviews/highperformancecoder/minsky/242)

### DIFF
--- a/model/godleyTableWindow.cc
+++ b/model/godleyTableWindow.cc
@@ -162,7 +162,6 @@ namespace minsky
     if (!godleyIcon) return;
     CairoSave cs(cairo);
     cairo_scale(cairo,zoomFactor,zoomFactor);
-    ZoomablePango::zoomFactor=zoomFactor;
     ZoomablePango pango(cairo);
     pango.setMarkup("Flows ↓ / Stock Vars →");
     rowHeight=pango.height()+2;


model/godleyTableWindow.cc
```diff
@@ -162,7 +162,6 @@ namespace minsky
     if (!godleyIcon) return;
     CairoSave cs(cairo);
     cairo_scale(cairo,zoomFactor,zoomFactor);
-    ZoomablePango::zoomFactor=zoomFactor;
```
highperformancecoder, line 165: I do think this solves the problem. ZoomablePango was a solution to a design problem in Pango that has now been corrected. But a ZoomablePango with zoomfactor=1 is just a regular Pango, so we should retire the whole class, to keep the code tidy.

highperformancecoder:
Thanks for diagnosing this. Steve wanted this fixed pronto, so I couldn't wait for the final code changes, so I committed the fix myself.

wdednam:
> Thanks for diagnosing this. Steve wanted this fixed pronto, so I couldn't wait for the final code changes, so I committed the fix myself.

Sure. :) I've relinquished ownership at Sourceforge, so the ticket is all yours :)

PR #243, opened 2020-11-10
for pars with vValues of rank 2 or higher, user can now choose which …

wdednam:
…pair of xvectors to display in sheet on tab, for vars, only the first two are displayed

---
This change is [Reviewable](https://reviewable.io/reviews/highperformancecoder/minsky/243)

### DIFF
--- a/gui-tk/wiring.tcl
+++ b/gui-tk/wiring.tcl
@@ -710,7 +710,7 @@ proc contextMenu {x y X Y} {
             .wiring.context add command -label "Flip" -command "$item.flip; flip_default"                     
             if {[$item.type]=="parameter"} {
                 .wiring.context add command -label "Import CSV" -command {CSVImportDialog}
-                .wiring.context add command -label "Select dimensions to display" -command {setupPickDimMenu}                 
+                .wiring.context add command -label "Display CSV values on tab" -command {setupPickDimMenu}                 
             }
             .wiring.context add command -label "Export as CSV" -command exportItemAsCSV
         }
@@ -830,11 +830,7 @@ proc setupPickDimMenu {} {
             foreach i [.wiring.context.pick.select.lb curselection] {
                 lappend pick [lindex $dimLabelPicked $i]
             }
-            if {[lindex $pick 1]==""} {
-				minsky.canvas.item.setDimLabelsPicked [lindex $pick 0] ""
-			} else {
-				minsky.canvas.item.setDimLabelsPicked [lindex $pick 0] [lindex $pick 1]
-			}
+			minsky.canvas.item.setDimLabelsPicked [lindex $pick 0] [lindex $pick 1]
             reset
         }
         button .wiring.context.pick.buttonBar.clear -text "Clear" -command {
--- a/gui-tk/wiring.tcl
+++ b/gui-tk/wiring.tcl
@@ -841,9 +841,6 @@ proc setupPickDimMenu {} {
     }
         
     set dimLabelPicked [minsky.canvas.item.dimLabels]
-    for {set i 0} {$i<[llength $dimLabelPicked]} {incr i} {
-        set idx([lindex $dimLabelPicked $i]) $i
-    }
     wm transient .wiring.context.pick
     wm geometry .wiring.context.pick +[winfo pointerx .]+[winfo pointery .]
     ensureWindowVisible .wiring.context.pick
--- a/gui-tk/wiring.tcl
+++ b/gui-tk/wiring.tcl
@@ -814,7 +814,7 @@ proc setupPickDimMenu {} {
     global dimLabelPicked
     if {![winfo exists .wiring.context.pick]} {
         toplevel .wiring.context.pick
-        wm title .wiring.context.pick "Pick two dimensions"
+        wm title .wiring.context.pick "Pick any two dimensions"
         frame .wiring.context.pick.select
         scrollbar .wiring.context.pick.select.vscroll -orient vertical -command {
             .wiring.context.pick.select.lb yview}
@@ -830,7 +830,11 @@ proc setupPickDimMenu {} {
             foreach i [.wiring.context.pick.select.lb curselection] {
                 lappend pick [lindex $dimLabelPicked $i]
             }
-            minsky.canvas.item.setDimLabelsPicked [lindex $pick 0] [lindex $pick 1]
+            if {[lindex $pick 1]==""} {
+				minsky.canvas.item.setDimLabelsPicked [lindex $pick 0] ""
+			} else {
+				minsky.canvas.item.setDimLabelsPicked [lindex $pick 0] [lindex $pick 1]
+			}
             reset
         }
         button .wiring.context.pick.buttonBar.clear -text "Clear" -command {
--- a/gui-tk/wiring.tcl
+++ b/gui-tk/wiring.tcl
@@ -710,6 +710,7 @@ proc contextMenu {x y X Y} {
             .wiring.context add command -label "Flip" -command "$item.flip; flip_default"                     
             if {[$item.type]=="parameter"} {
                 .wiring.context add command -label "Import CSV" -command {CSVImportDialog}
+                .wiring.context add command -label "Select dimensions to display" -command {setupPickDimMenu}                 
             }
             .wiring.context add command -label "Export as CSV" -command exportItemAsCSV
         }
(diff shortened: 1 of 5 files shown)


model/variable.h
```diff
@@ -129,7 +130,21 @@ namespace minsky
       if (auto v=vValue()) return v->hypercube().dims();
       else return {};
     }
-      
+
+    std::vector<std::string> dimLabels() const {
+      std::vector<std::string> tmpLabels;
```
highperformancecoder, line 135: Decl should be moved to just before line 137

model/variable.h
```diff
@@ -76,6 +76,7 @@ namespace minsky
   private:
     CLASSDESC_ACCESS(VariableBase);
     std::string m_name; 
+    std::pair<std::string,std::string> m_dimLabelsPicked=std::make_pair("","");
```
highperformancecoder, line 79: You could just use the default initializer for pair here.

gui-tk/wiring.tcl
```diff
@@ -809,6 +810,46 @@ proc contextMenu {x y X Y} {
...
+        listbox .wiring.context.pick.select.lb -listvariable dimLabelPicked \
+            -selectmode extended -selectforeground blue \
+            -width 35 \
+            -yscrollcommand {.wiring.context.pick.select.vscroll set} 
+        pack .wiring.context.pick.select.lb -fill both  -expand y -side left
+        pack .wiring.context.pick.select.vscroll -fill y -expand y -side left
+        pack .wiring.context.pick.select
+        buttonBar .wiring.context.pick {
+            set pick {}
+            foreach i [.wiring.context.pick.select.lb curselection] {
```
highperformancecoder, line 830: curelection returns a list, so no need to iteratively build a new list, just assign it with a set statement

gui-tk/wiring.tcl
```diff
@@ -809,6 +810,46 @@ proc contextMenu {x y X Y} {
...
+                lappend pick [lindex $dimLabelPicked $i]
+            }
+            minsky.canvas.item.setDimLabelsPicked [lindex $pick 0] [lindex $pick 1]
```
highperformancecoder, line 833: You need to deal with the possibility that the user has selected less than two dimensions

model/variable.h
wdednam: Thanks, done.

model/variable.h
wdednam: Wasn't sure. Also fixed that up now.

gui-tk/wiring.tcl
wdednam: I don't why, but when I assign [.wiring.context.pick.select.lb curselection] directly to "dimLabelPick", and then append it to "pick", only the indices I picked appear, which obviously don't agree with the labels themselves, so the first two labels are always chosen. I've tried to do this in several different way but the result is also the same... For some reason dimLabelPicked is not assigned correctly...

gui-tk/wiring.tcl
wdednam: I've dealt with this, but don't know if a switch statement would be more appropriate? (Not sure if I need only two or three cases?)

gui-tk/wiring.tcl
highperformancecoder: My apologies - I thought curselection returned the selected values. I misread the documentation - I see now it returns indices, so your proposed code is as good as any.

gui-tk/wiring.tcl
highperformancecoder: I was concerned that lindex might throw an error if the list had fewer than 2 elements. Actually, the documentation of lindex is silent on what happens here. Currently the behaviour is that an empty string will be passed to minsky.canvas.item.setDimLabelsPicked, which knowing how its used, is not disasterous. Maybe leave it for now, hopefully we'll be off TCL before lindex's error behaviour changes...

gui-tk/wiring.tcl
```diff
@@ -809,6 +810,50 @@ proc contextMenu {x y X Y} {
...
+        buttonBar .wiring.context.pick {
+            set pick {}
+            foreach i [.wiring.context.pick.select.lb curselection] {
+                lappend pick [lindex $dimLabelPicked $i]
+            }
+            if {[lindex $pick 1]==""} {
+				minsky.canvas.item.setDimLabelsPicked [lindex $pick 0] ""
+			} else {
+				minsky.canvas.item.setDimLabelsPicked [lindex $pick 0] [lindex $pick 1]
+			}
```
highperformancecoder, line 837: I would have done a check on llength of $pick, but given how things stand, we should just go back to your original code.

gui-tk/wiring.tcl
```diff
@@ -809,6 +810,50 @@ proc contextMenu {x y X Y} {
...
+            .wiring.context.pick.select.lb selection clear 0 end}
+        pack .wiring.context.pick.buttonBar.clear -side left
+    } else {
+        deiconify .wiring.context.pick
+    }
+        
+    set dimLabelPicked [minsky.canvas.item.dimLabels]
+    for {set i 0} {$i<[llength $dimLabelPicked]} {incr i} {
+        set idx([lindex $dimLabelPicked $i]) $i
+    }
```
highperformancecoder, line 850: What is this for?

model/parVarSheet.cc
```diff
@@ -52,7 +52,9 @@ namespace minsky
             double w=0,h=0,w_prev, h_prev,lh,lw; 
             for (auto& it: itemVector)
               {
-                auto value=it->variableCast()->vValue();
+                auto v=it->variableCast();
+                auto value=v->vValue();
```
highperformancecoder, line 56: If it is not a variable, then you have introduced a null dereference. It it  is guaranteed to be a variable, then the original code was fine. (maybe add an assertion to that fact)

model/variable.h
```diff
@@ -129,7 +130,21 @@ namespace minsky
       if (auto v=vValue()) return v->hypercube().dims();
       else return {};
     }
-      
+
+    std::vector<std::string> dimLabels() const {
+      if (auto v=vValue()) {
+        std::vector<std::string> tmpLabels;			  
+        for (auto& i: v->hypercube().xvectors) tmpLabels.push_back(static_cast<std::string>(i.name));
+        return tmpLabels;
```
highperformancecoder, line 138: Maybe useful enough operation to make a member of Hypercube - perhaps call it Hypercube::dimNames()?

highperformancecoder: BTW - the static_cast is not required here.

model/parVarSheet.cc
```diff
@@ -129,106 +131,115 @@ namespace minsky
...
                    
                     y0=h+3.1*rowHeight;                 
-			      }
-			    else
-			      {
+                  }
+                else
+                  { 
                     cairo_move_to(cairo,x,y-1.5*rowHeight);
                     pango.setMarkup(latexToPango(value->name)+":");
```
highperformancecoder, line 138: The remainder of the code in this function is too difficult to review, but IIUC it is just reformatted from what was before.

gui-tk/wiring.tcl
wdednam: Okay, done.

gui-tk/wiring.tcl
wdednam: Also done.

gui-tk/wiring.tcl
wdednam: This populates the list that appears in the selection box. Without it, the selection box remains empty.

model/parVarSheet.cc
wdednam: The only reason I added this is because I use the methods I added to the the VariableBase class further down in this code. But, maybe I can move those methods to the Hypercube class and then I won't have to do the above...

model/variable.h
wdednam: > Maybe useful enough operation to make a member of Hypercube - perhaps call it Hypercube::dimNames()?

Very good suggestion, thanks. Going to try to move all the new methods I added to VariableBase to Hypercube.

wdednam: > BTW - the static_cast is not required here.

True, getting rid of it :)

model/parVarSheet.cc
wdednam: Yes. I've seen you've made improvements in the Sheet class and I've been comparing that code with mine to make improvements here too.

gui-tk/wiring.tcl
highperformancecoder: dimLabelPicked is the list of values in the listBox. What is idx? It doesn't appear to be used anywhere I can see.

model/parVarSheet.cc
highperformancecoder: OK - Probably  missed that...

gui-tk/wiring.tcl
wdednam: True, I thought you were referring to these lines and the one immediately before it... So, when I commented them out, the list box was empty. These 3 lines contribute nothing. Sorry.

wdednam:
I think this is ready for review and merging now. Next task is to finalize const version of variable tab.

PR #244, opened 2020-11-16
first attempt at fixing bugs 1254 and 1272

wdednam:
This change is [Reviewable](https://reviewable.io/reviews/highperformancecoder/minsky/244)

### DIFF
--- a/model/minsky.cc
+++ b/model/minsky.cc
@@ -1393,13 +1393,21 @@ namespace minsky
              if (type!=VariableType::flow)
                for (auto v: g->flowVars())
                  if (v->valueId()==name)
-                   try {newName=v->name()+"^{Flow}"; g->table.renameFlows(v->name(),newName);}
-                   catch (...) {throw error("flow variables in Godley tables cannot be converted to a different type");}                  
+                   {
+					   newName=v->name()+"^{Flow}";
+                       VariableValues::iterator iv=variableValues.find(newName);
+                       if (iv==variableValues.end()) g->table.renameFlows(v->name(),newName);
+					   else throw error("flow variables in Godley tables cannot be converted to a different type");
+					}
              if (type!=VariableType::stock)
                for (auto v: g->stockVars())
                  if (v->valueId()==name)
-                   try {newName=v->name()+"^{Stock}"; g->table.renameStock(v->name(),newName);}
-                   catch (...) {throw error("stock variables in Godley tables cannot be converted to a different type");}
+                   {
+					   newName=v->name()+"^{Stock}";
+                       VariableValues::iterator iv=variableValues.find(newName);
+                       if (iv==variableValues.end()) g->table.renameStock(v->name(),newName);
+					   else throw error("stock variables in Godley tables cannot be converted to a different type");
+				  }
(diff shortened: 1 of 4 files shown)


model/minsky.cc
```diff
@@ -1389,22 +1389,26 @@ namespace minsky
        {
          if (auto g=dynamic_cast<GodleyIcon*>(i->get()))
            {
+			 string newName;  
              if (type!=VariableType::flow)
                for (auto v: g->flowVars())
                  if (v->valueId()==name)
-                   throw error("flow variables in Godley tables cannot be converted to a different type");
+                   try {newName=v->name()+"^{Flow}"; g->table.renameFlows(v->name(),newName);}
```
highperformancecoder, line 1396: Current renameFlows does not throw, so this try/catch block will not catch anything.

highperformancecoder: Don't you want to check somewhere that the new name is not an existing flow?

test/testModel.cc
```diff
@@ -1099,8 +1099,8 @@ SUITE(Integrate)
       model->addItem(a);
       model->addWire(new Wire(intop->ports[0],a->ports[1]));
       intop->description("a");
-      // should cowardly refuse, and give a different name
-      CHECK_EQUAL("a1",intop->description());
+      // should cowardly refuse, and give a different name. Wynand: I don't really understand this. why can't the description be the same as the intop name?
```
highperformancecoder, line 1102: Because there is already a flow variable called "a". If it didn't "cowardly refuse", then the model ends up in an inconsistent state of a being both a flow variable and and integral variable.

model/minsky.cc
```diff
@@ -1389,22 +1389,33 @@ namespace minsky
...
+			 string newName;  
              if (type!=VariableType::flow)
                for (auto v: g->flowVars())
                  if (v->valueId()==name)
-                   throw error("flow variables in Godley tables cannot be converted to a different type");
+                   {
+					   newName=v->name()+"^{Flow}";
+                       VariableValues::iterator iv=variableValues.find(newName);
+                       if (iv==variableValues.end()) g->table.renameFlows(v->name(),newName);
+					   else throw error("flow variables in Godley tables cannot be converted to a different type");
```
highperformancecoder, line 1400: OK for now, but you can also create a loop, such that it tries ^{Flow}, ^{Flow1}, ... until it finds a name that doesn't clash.

model/minsky.cc
```diff
@@ -1389,22 +1389,33 @@ namespace minsky
...
              if (type!=VariableType::stock)
                for (auto v: g->stockVars())
                  if (v->valueId()==name)
-                   throw error("stock variables in Godley tables cannot be converted to a different type");
+                   {
+					   newName=v->name()+"^{Stock}";
+                       VariableValues::iterator iv=variableValues.find(newName);
+                       if (iv==variableValues.end()) g->table.renameStock(v->name(),newName);
+					   else throw error("stock variables in Godley tables cannot be converted to a different type");
+				  }
```
highperformancecoder, line 1410: ditto

PR #245, opened 2020-11-17
why can't I construct jacobian this way?

wdednam:
This change is [Reviewable](https://reviewable.io/reviews/highperformancecoder/minsky/245)


model/minsky.cc
```diff
@@ -1078,17 +1078,27 @@ namespace minsky
         ds[j]=1;
         for (size_t i=0; i<equations.size(); ++i)
           equations[i]->deriv(&df[0], df.size(), &ds[0], sv, &flow[0]);
-        vector<double> d(stockVars.size());
-        evalGodley.eval(&d[0], &df[0]);
+        double d[stockVars.size()];
```
highperformancecoder, line 1081: This is not standard C++.

highperformancecoder:
You might need to check that the d1 & d2 definitions are correct for tensor operations. They may not be.

model/minsky.cc
wdednam: Whoops. i meant to change it back and forgot...

wdednam:
> You might need to check that the d1 & d2 definitions are correct for tensor operations. They may not be.

Thank you, I will. At first I thought it was just the size of the Jacobian that may be the problem, but it is not as simple as that.

wdednam:
Crap, I thought I had a separate branch for this. Going to revert now and submit a different PR. Sorry...

highperformancecoder:
What's happening with this PR?

wdednam:
> What's happening with this PR?

Unfortunately. not much. I spent more time on this after your suggestion but did not make much progress. So, you can close it if you want to.

PR #246, opened 2020-11-19
const version of variable tab as described at feature 15 on the way. …

wdednam:
…Don't understand why in-row spacing is not the same for column headings and values of var attributes...

---
This change is [Reviewable](https://reviewable.io/reviews/highperformancecoder/minsky/246)


highperformancecoder:
Code looks good to me!

wdednam:
> Code looks good to me!

Thanks :)

I still need to align the attribute headings and values vertically... and draw in vertical separation lines too...

model/parVarSheet.cc
```diff
@@ -40,6 +40,41 @@ namespace minsky
...
                                        });   	
   }
+  
+  int ParVarSheet::colX(double x) const
+  { 
+	if (itemVector.empty()) return -1;
+    if ((x-offsx)<colLeftMargin[0]) return -1;
+    if ((x-offsx)<colLeftMargin[1]) return 0;
+    auto p=std::upper_bound(colLeftMargin.begin(), colLeftMargin.end(), (x-offsx));
+    size_t r=p-colLeftMargin.begin()-2+scrollColStart;
```
highperformancecoder, line 50: Assuming scrollColStart=1, then this formula will reproduce lines 47 and 48 anyway.

model/parVarSheet.cc
```diff
@@ -40,6 +40,41 @@ namespace minsky
...
+  }
+  
+  ParVarSheet::ClickType ParVarSheet::clickType(double x, double y) const
+  {
+    int c=colX(x), r=rowY(y);
+        
+    //if (c>=0 && c<int(cols()))
+    //  if (r>=0 && r<int(rows()))
+    if (c>=0 && c<varAttribVals.size())
+      if (r>=0 && r<2*itemVector.size())
```
highperformancecoder, line 73: the two conditionals can be combined by &&

highperformancecoder:
Only minor stuff, which I'm happy to merge, except you'll need to sort out why Travis is failing it.

model/parVarSheet.cc
wdednam: Thanks. I removed these, because they are not necessary on the tabs, the offsets offsx and offsy are enough.

model/parVarSheet.cc
wdednam: Done :)

wdednam:
> Only minor stuff, which I'm happy to merge, except you'll need to sort out why Travis is failing it.

It works for rank 0 vars, but not when they are mixed. I thought I could do without a rowTopMargin vector of doubles to keep track of the var positions on the tab, but when there vars of mixed ranks, my implementation is not up to scratch... I'm busy fixing this now.

model/parVarSheet.cc
```diff
@@ -40,6 +40,38 @@ namespace minsky
                                          return false;
                                        });   	
   }
+  
+  int ParVarSheet::colX(double x) const
+  { 
+	if (itemVector.empty()) return -1;
+	size_t c;
+    for (auto& i: colLeftMargin)
```
highperformancecoder, line 48: Code returns undefined value if colLeftMargin is empty. Is this possible?

model/parVarSheet.cc
```diff
@@ -40,6 +40,38 @@ namespace minsky
...
+       c=p-i.second.begin();
+       }
+    if (c<0) c=-1; // out of bounds, invalidate
+    return c;
+  }
+
+  int ParVarSheet::rowY(double y) const
+  {
+	if (itemVector.empty()) return -1;     
+    auto p=std::upper_bound(rowTopMargin.begin(), rowTopMargin.end(), (y-offsy));
```
highperformancecoder, line 60: rowY returns 0 if rowTopMargin is empty, so long as this is OK.

wdednam:
This should be fine for the purposes of tickets 72 and 73, but editing the individual cells on the tab is going to be difficult because the colX(x) method does not correctly find the column numbers when there are several vars of different rank on the tab. So, more work is required...

PR #249, opened 2020-11-27
Bugfix 1275

wdednam:
This change is [Reviewable](https://reviewable.io/reviews/highperformancecoder/minsky/249)

### DIFF
--- a/model/variable.cc
+++ b/model/variable.cc
@@ -416,7 +416,7 @@ bool VariableBase::visible() const
   //toplevel i/o items always visible
   if ((!g || !g->group.lock()) && g==controller.lock()) return true;
   // ensure pars and constants with invisible out wires are made invisible. for ticket 1275  
-  if (type()==constant || type()==parameter)
+  if (attachedToDefiningVar() && (type()==constant || type()==parameter))
   {
     for (auto w: ports[0]->wires())
       if (w->visible()) return true;
--- a/model/variable.cc
+++ b/model/variable.cc
@@ -418,12 +418,8 @@ bool VariableBase::visible() const
   // ensure pars and constants with invisible out wires are made invisible. for ticket 1275  
   if (type()==constant || type()==parameter)
   {
-    for (auto w: ports[0]->wires())
-      if (w->attachedToDefiningVar()) 
-      {
-        if (w->visible()) return true;
-        else return false;  
-	  }
+	if (std::any_of(ports[0]->wires().begin(),ports[0]->wires().end(), [](Wire* w){return w->attachedToDefiningVar() && !w->visible();})) return false;
+	else return true;
   }  
   // ensure flow vars with out wires remain visible. for ticket 1275
   if (attachedToDefiningVar() && !ports[0]->wires().empty()) return true;  
--- a/model/variable.cc
+++ b/model/variable.cc
@@ -415,6 +415,10 @@ bool VariableBase::visible() const
   auto g=group.lock();
   //toplevel i/o items always visible
   if ((!g || !g->group.lock()) && g==controller.lock()) return true;
+  // ensure vars attached to invisible ops are also made invisible. for ticket 1275
+  if (!ports[0]->wires().empty())
+    if (auto i=ports[0]->wires()[0]->to()->item().operationCast())
+      if (!i->visible()) return false;
   return !controller.lock() && Item::visible();
 }
 
--- a/model/variable.cc
+++ b/model/variable.cc
@@ -416,11 +416,14 @@ bool VariableBase::visible() const
   //toplevel i/o items always visible
   if ((!g || !g->group.lock()) && g==controller.lock()) return true;
   // ensure pars and constants with invisible out wires are made invisible. for ticket 1275  
-  if (attachedToDefiningVar() && (type()==constant || type()==parameter))
+  if (type()==constant || type()==parameter)
   {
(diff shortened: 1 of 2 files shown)


model/variable.cc
```diff
@@ -415,6 +415,10 @@ bool VariableBase::visible() const
   auto g=group.lock();
   //toplevel i/o items always visible
   if ((!g || !g->group.lock()) && g==controller.lock()) return true;
+  // ensure vars attached to invisible ops are also made invisible. for ticket 1275
+  if (!ports[0]->wires().empty())
+    if (auto i=ports[0]->wires()[0]->to()->item().operationCast())
+      if (!i->visible()) return false;
```
highperformancecoder, line 421: This doesn't sound right. If a variable has been moved to the var tab, then it should still be visible if downstream items are visible.

An item becomes invisible if all downstream items are invisible or a downstream item is  variable move to the variable (regardless of visibility).

These rules need to be applied recursively, of course.

model/wire.cc
```diff
@@ -103,6 +103,7 @@ namespace minsky
   {
     auto f=from(), t=to();
     if (attachedToDefiningVar()) return false;
+    if (f->item().attachedToDefiningVar()) return false;      // ensure wires attached to out port do not dangle. for ticket 1275
```
highperformancecoder, line 106: Maybe for wires, visibility is simply the conjunction (&&) of the visibility of the to() and from() objects.

wdednam: I tried these combinations but they lead to segfaults in Wire::attachedToDefiningVar() and more dangling wires in Wire::visible().

model/variable.cc
wdednam: > This doesn't sound right. If a variable has been moved to the var tab, then it should still be visible if downstream items are visible.

Doesn't the code I added only make the var invisible if an op attached to its outport is invisible by virtue of another downstream var already being invisible?

I think my comment above the code may be misleading. I meant to say "ensure vars with invisible ops attached to their outports are also made invisible. for ticket 1275"

model/wire.cc
highperformancecoder: Maybe you're conflating visibility with attachedToDefiningVar...

model/variable.cc
highperformancecoder: Doesn't the code I added only make the var invisible if an op attached to its outport is invisible by virtue of another downstream var already being invisible?

Yes - but this is not what you want to do. A variable is invisible if it moved to the variable tab and no item attached to its output port is visible

wdednam: So, the variable itself should stay visible but not the items attached to it? I quote from ticket 72:

> The defining groupm should then be removed from the canvas (or made invisible).

model/wire.cc
wdednam: Yes, I think we should discuss this on Monday morning during our meeting. I think I'm missing the point of the feature. I thought it was to reduce clutter on the canvas by moving a var to the tab.

model/variable.cc
```diff
@@ -415,6 +415,20 @@ bool VariableBase::visible() const
   auto g=group.lock();
   //toplevel i/o items always visible
   if ((!g || !g->group.lock()) && g==controller.lock()) return true;
+  // ensure pars and constants with invisible out wires are made invisible. for ticket 1275  
+  if (type()==constant || type()==parameter)
+  {
+    for (auto w: ports[0]->wires())
+      if (w->attachedToDefiningVar()) 
+      {
+        if (w->visible()) return true;
```
highperformancecoder, line 424: Can be simplified to "return w->visible()"

highperformancecoder: But I think the code is still wrong. This will return false when the first outgoing wire is invisible, even if the remaning wires are visible = dangling wires.

What you should do is set a boolean flag that tracks whether any wire is attachedToDefiningVar. Then return false on line 428 if that flag is set. Also I think the code on line 429 is now obsolete.

wdednam: Thanks for the tips :)

> What you should do is set a boolean flag that tracks whether any wire is attachedToDefiningVar.

I tried using std::any_of, but it didn't work previously. Now it seemingly does with this:

```
if (std::any_of(ports[0]->wires().begin(),ports[0]->wires().end(), [](Wire* w){return w->attachedToDefiningVar() && !w->visible();})) return false;
	else return true;
```

> Also I suspect the code on line 429 is now obsolete.

I thought so too, but when I remove it vars attached to plots still disappear in the fundamentalConstants.mky example.

model/variable.cc
```diff
@@ -415,6 +415,16 @@ bool VariableBase::visible() const
   auto g=group.lock();
   //toplevel i/o items always visible
   if ((!g || !g->group.lock()) && g==controller.lock()) return true;
+  // ensure pars and constants with invisible out wires are made invisible. for ticket 1275  
+  if (type()==constant || type()==parameter)
+  {
+	if (std::any_of(ports[0]->wires().begin(),ports[0]->wires().end(), [](Wire* w){return w->attachedToDefiningVar() && !w->visible();})) return false;
+	else return true;
```
highperformancecoder, line 422: Clever, but I think you should be using all_of rather than any_of

wdednam: Hmmm. I knew there was a reason why I used any_of instead of all_of. When I use all_of all parameters disappear from the canvas... Reverting now...

PR #250, opened 2020-12-01
variable definitions now appear on var tab

wdednam:
This change is [Reviewable](https://reviewable.io/reviews/highperformancecoder/minsky/250)

### DIFF
--- a/model/parVarSheet.cc
+++ b/model/parVarSheet.cc
@@ -23,7 +23,9 @@
 #include <pango.h>
 #include "minsky_epilogue.h"
 #include "minsky.h"
+#include "equations.h"
 using namespace std;
+using namespace MathDAG;
 using ecolab::cairo::Surface;
 using ecolab::Pango;
 using ecolab::cairo::CairoSave;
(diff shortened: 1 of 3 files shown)


model/parVarSheet.cc
```diff
@@ -72,6 +74,38 @@ namespace minsky
...
+
+    for (const VariableDAG* i: system.integrationVariables)
+      {
+		if (i->valueId==valueId) 
+		{  
+            VariableDAGPtr input=system.expressionCache.getIntegralInput(i->valueId);
+            if (input && input->rhs)
+              input->rhs->matlab(o);
+	    }
+      }
```
highperformancecoder, line 104: This does work, but I'm sure you appreciate won't scale well for large models. There's two features you could use to look up the nodes defining a variable in O(log(N)), rather than O(N). The first is SystemOfEquations::getNodeFromWire, which requires a wire to look up. The other is to use the subexression cache directly: You can look up a variable's definition by name - you might need to know whether you're starting with a stock or a flow variable first.

Currently, permissions won't allow you to do this from parVarSheet, so you'll need to rearrange things a bit. Maybe add another method to SystemOfEquations?

highperformancecoder:
I'll merge this for now, but please refactor definition to use direct lookup of the definitions.

PR #251, opened 2020-12-02
works the same as previous implementation, except I can't get integra…

wdednam:
…l definitions to show

---
This change is [Reviewable](https://reviewable.io/reviews/highperformancecoder/minsky/251)


engine/equations.cc
```diff
@@ -760,7 +760,25 @@ namespace MathDAG
           }
       }
     return r;
-  }     
+  }
+  
+  VariableDAGPtr SystemOfEquations::getNodeFromVar(const VariablePtr v)
+  {
+    NodePtr r;
+    if (expressionCache.exists(*v))
```
highperformancecoder, line 768: From what I remember, integral inputs occupy a different "namespace" within the expressionCache. Need to look at the code...

engine/equations.cc
```diff
@@ -760,7 +760,29 @@ namespace MathDAG
...
+    else if (v && v->type()!=VariableBase::undefined) r=makeDAG(*v);
+    return dynamic_pointer_cast<VariableDAG>(r);
+  }         
+
+  ostringstream SystemOfEquations::getDefFromIntVar(const VariablePtr v)
+  {
+    ostringstream o;
+    
+	std::vector<VariableDAG*>::const_iterator it1=std::find_if(integrationVariables.begin(),integrationVariables.end(),[&](VariableDAG* i){return i->valueId==v->valueId();});    
+    if (*it1)
```
highperformancecoder, line 779: This bit of code is unnecessary. v already has a valueId() method, and getIntegralInput will return a null pointer if nothing has been cached. It won't mutate expressionCache's state.

model/parVarSheet.cc
```diff
@@ -77,32 +77,21 @@ namespace minsky
...
-	    }
-      }
-   
+    ostringstream o;
+	
+	auto varDAG=system.getNodeFromVar(*v);
+    
+    if (v->type()!=VariableType::parameter)
+    {
+      if (varDAG && varDAG->rhs && varDAG->type!=VariableType::constant && !dynamic_cast<const IntegralInputVariableDAG*>(varDAG.get()))
```
highperformancecoder, line 90: It would suffice to check the type tag for being VariableType::integral, rather than dynamic_casting.

engine/equations.cc
```diff
@@ -760,7 +760,29 @@ namespace MathDAG
...
+  VariableDAGPtr SystemOfEquations::getNodeFromVar(const VariablePtr v)
+  {
+    NodePtr r;
+    if (expressionCache.exists(*v))
+      return dynamic_pointer_cast<VariableDAG>(expressionCache[*v]);
+    else if (v && v->type()!=VariableBase::undefined) r=makeDAG(*v);
+    return dynamic_pointer_cast<VariableDAG>(r);
+  }         
+
+  ostringstream SystemOfEquations::getDefFromIntVar(const VariablePtr v)
```
highperformancecoder, line 774: I would make the parameter const VariableBase&

engine/equations.cc
```diff
@@ -760,7 +760,29 @@ namespace MathDAG
           }
       }
     return r;
-  }     
+  }
+  
+  VariableDAGPtr SystemOfEquations::getNodeFromVar(const VariablePtr v)
```
highperformancecoder, line 765: I would make the parameter const VariableBase&

model/parVarSheet.cc
```diff
@@ -77,32 +77,21 @@ namespace minsky
   
 namespace
 {
-  std::string definition(const string& valueId)
+ 
+  std::string definition(const VariablePtr v)
```
highperformancecoder, line 81: parameter should be const VariableBase&

engine/equations.cc
```diff
@@ -774,7 +774,27 @@ namespace MathDAG
...
+    else if (v.type()!=VariableBase::undefined) r=makeDAG(const_cast<VariableBase&>(v));
+    return dynamic_pointer_cast<VariableDAG>(r);
+  }         
+
+  ostringstream SystemOfEquations::getDefFromIntVar(const VariableBase& v)
+  {
+    ostringstream o;
+    
+	std::vector<VariableDAG*>::const_iterator it1=std::find_if(integrationVariables.begin(),integrationVariables.end(),[&](VariableDAG* i){return i->valueId==v.valueId();});         
+    VariableDAGPtr input=expressionCache.getIntegralInput((*it1)->valueId);
```
highperformancecoder, line 793: No need for the find_if on line 792. We already have valueId as v.valueId()

engine/equations.cc
```diff
@@ -774,7 +774,27 @@ namespace MathDAG
...
+
+  ostringstream SystemOfEquations::getDefFromIntVar(const VariableBase& v)
+  {
+    ostringstream o;
+    
+	std::vector<VariableDAG*>::const_iterator it1=std::find_if(integrationVariables.begin(),integrationVariables.end(),[&](VariableDAG* i){return i->valueId==v.valueId();});         
+    VariableDAGPtr input=expressionCache.getIntegralInput((*it1)->valueId);    
+    if (input && input->rhs) input->rhs->matlab(o);    
+    
+    return o;
```
highperformancecoder, line 796: Clever use of return value optimisation (RVO). Such a statement would not be possible in old-fashioned C++!

engine/equations.cc
wdednam: Removed, thanks.

engine/equations.cc
wdednam: Thanks :)

PR #252, opened 2020-12-04
reverted all_of to any_of in VariableBase::visible() because it led t…

wdednam:
…o invisible unwired pars and consts

---
This change is [Reviewable](https://reviewable.io/reviews/highperformancecoder/minsky/252)

### DIFF
--- a/model/variable.cc
+++ b/model/variable.cc
@@ -418,7 +418,7 @@ bool VariableBase::visible() const
   // ensure pars and constants with invisible out wires are made invisible. for ticket 1275  
   if (type()==constant || type()==parameter)
   {
-	if (std::all_of(ports[0]->wires().begin(),ports[0]->wires().end(), [](Wire* w){return w->attachedToDefiningVar() && !w->visible();})) return false;
+	if (std::any_of(ports[0]->wires().begin(),ports[0]->wires().end(), [](Wire* w){return w->attachedToDefiningVar() && !w->visible();})) return false;
 	else return true;
   }  
   // ensure flow vars with out wires remain visible. for ticket 1275
--- a/model/variable.cc
+++ b/model/variable.cc
@@ -415,8 +415,8 @@ bool VariableBase::visible() const
   auto g=group.lock();
   //toplevel i/o items always visible
   if ((!g || !g->group.lock()) && g==controller.lock()) return true;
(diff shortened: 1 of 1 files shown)


highperformancecoder:
master-canvas
PR252-canvas
PR252-vartab

See attached images of the same model - a parameter foo feeding into two variables var2 and var3. The current master head (using all_of) works correctly, any_of does not.

BTW - when creating a parameter on the canvas, it is automatically placed on the parameter tab, and invisible on the canvas, because it has no visible wires emanating from it.There is no way to correct this. By default, when adding parameters to the canvas, they should not be marked as on the parameter tab.

model/variable.cc
```diff
@@ -415,8 +415,8 @@ bool VariableBase::visible() const
   auto g=group.lock();
   //toplevel i/o items always visible
   if ((!g || !g->group.lock()) && g==controller.lock()) return true;
-  // ensure pars and constants with invisible out wires are made invisible. for ticket 1275  
-  if (type()==constant || type()==parameter)
+  // ensure pars, constants and flows with invisible out wires are made invisible. for ticket 1275  
+  if ((type()==constant || type()==parameter) && !ports[0]->wires().empty())
```
highperformancecoder, line 419: You need to work out why varTabDisplay is being set on parameters on their creation. I can't see the smoking gun by eye - you'll need to debug it.

highperformancecoder: Let's discuss this some more. Could you please outline what you think the rules should be, and we walk through the consequences.

highperformancecoder: Actually, on second thoughts, unwired parameters should return false on attachedToDefiningVar() being true, so I will merge this change. But that and the the next clause should apply for all items, not just constants and variables, so should be moved into Item::visible. Once done, line 425 is no longer relevant. Line 426 also needs to be moved into IntOp (avoiding the dynamic_cast).  My thought is that an IntOp->attachedToDefiningVar() should delegate to its intVar.

wdednam: Thanks for these tips :) I can get most of what you suggest to work, but not the IntOp. It defies my every attempt when I move the code to the IntOp class... I'll keep trying and  hopefully by tomorrow morning (my time) have something working properly...

PR #253, opened 2020-12-04
trying to populate index() for result of outer product, but the index…

wdednam:
…() objects always seem to be empty regardless of whether I load sparse or dense tensors?

---
This change is [Reviewable](https://reviewable.io/reviews/highperformancecoder/minsky/253)

### DIFF
--- a/engine/minskyTensorOps.cc
+++ b/engine/minskyTensorOps.cc
@@ -368,17 +368,23 @@ namespace minsky
   {
     std::shared_ptr<ITensor> arg1, arg2;
     void computeTensor() const override {//TODO Sparse implementation
-      size_t m=arg1->hypercube().numElements(), n=arg2->hypercube().numElements();   
+      size_t m=arg1->size(), n=arg2->size();   
   	
-      for (size_t i=0; i< m; i++)
-      {
-        auto v1=(*arg1)[i];  			
-        for (size_t j=0; j< n; j++) 
-        {
-            auto v2=(*arg2)[j];			
-            if (!isnan(v1) && !isnan(v2)) cachedResult[i+j*m]=v1*v2;			
-		}
-	  }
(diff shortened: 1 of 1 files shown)


engine/minskyTensorOps.cc
```diff
@@ -393,6 +399,15 @@ namespace minsky
       hc.xvectors.insert(hc.xvectors.begin(), xv1.begin(), xv1.end());           
       cachedResult.hypercube(move(hc));
         
+      set<size_t> newIdx;
+      for (auto& i: arg1->index())
+        for (auto& j: arg2->index()) 
+        {
+            newIdx.insert(i+arg1->index().size()*j);
```
highperformancecoder, line 406: This doesn't look right. The new index should be i+arg1->hypercube().numElements()*j, but obviously move the numElements() call outside the loop.

engine/minskyTensorOps.cc
```diff
@@ -368,17 +368,23 @@ namespace minsky
...
-		}
-	  }
+      //if (cachedResult.index().empty())  	
+       for (size_t i=0; i< m; i++)
+       {
+         auto v1=(*arg1)[i];  			
+         for (size_t j=0; j< n; j++) 
+         {
+             auto v2=(*arg2)[j];			
+             if (!isnan(v1) && !isnan(v2)) cachedResult[i+j*m]=v1*v2;
```
highperformancecoder, line 380: This should be fine. Note the test for isnan is not necessary, a nan multiplied by anything else is still a nan.

engine/minskyTensorOps.cc
wdednam: Thanks, done!

engine/minskyTensorOps.cc
wdednam: Ok, I got rid of the conditional.

PR #254, opened 2020-12-08
first attempt at displaying plots on tab. tried using canvas in tcl, …

wdednam:
…but don't know how to broadcast it to a tab. the examples I've looked at are not very helpful: https://wiki.tcl-lang.org/page/canvas

---
This change is [Reviewable](https://reviewable.io/reviews/highperformancecoder/minsky/254)

### DIFF
--- a/gui-tk/minsky.tcl
+++ b/gui-tk/minsky.tcl
@@ -937,6 +937,8 @@ addTab parameters "Parameters" minsky.parameterSheet
 pack .parameters.canvas -fill both -expand 1
 addTab variables "Variables" minsky.variableSheet
 pack .variables.canvas -fill both -expand 1
+addTab plts "Plots" minsky.plotSheet
+pack .plts.canvas -fill both -expand 1
 .tabs select 0
 
 bind .variables.canvas <<contextMenu>> "variableContext %x %y %X %Y"  
@@ -972,6 +974,7 @@ bind .wiring.canvas <Configure> {setScrollBars; minsky.panopticon.width %w; mins
 bind .equations.canvas <Configure> {setScrollBars}
 bind .parameters.canvas <Configure> {setScrollBars}
 bind .variables.canvas <Configure> {setScrollBars}
+bind .plts.canvas <Configure> {setScrollBars}
 
 set helpTopics(.wiring.panopticon) Panopticon
 
@@ -1004,7 +1007,13 @@ proc setScrollBars {} {
             set y0 [expr (10000-[variableSheet.offsy])/20000.0]
             .hscroll set $x0 [expr $x0+[winfo width .variables.canvas]/20000.0]
             .vscroll set $y0 [expr $y0+[winfo height .variables.canvas]/20000.0]                 
-        }        
+        }
+        .plts {
+            set x0 [expr (10000-[plotSheet.offsx])/20000.0]
+            set y0 [expr (10000-[plotSheet.offsy])/20000.0]
+            .hscroll set $x0 [expr $x0+[winfo width .plts.canvas]/20000.0]
+            .vscroll set $y0 [expr $y0+[winfo height .plts.canvas]/20000.0]                 
+        }                 
     }
 }
 
@@ -1033,6 +1042,11 @@ proc panCanvas {offsx offsy} {
             variableSheet.offsy $offsy						
             variableSheet.requestRedraw
         }           
+        .plts {
+            plotSheet.offsx $offsx
+            plotSheet.offsy $offsy						
+            plotSheet.requestRedraw
+        }            
     }
     setScrollBars
 }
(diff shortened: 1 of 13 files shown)


model/plotSheet.h
```diff
@@ -0,0 +1,36 @@
+/*
+  @copyright Steve Keen 2019
```
highperformancecoder, line 2: Need to update the copyright date

model/parVarSheet.cc
```diff
@@ -370,7 +371,13 @@ namespace
                 else y0+=4.1*rowHeight;   
                 iC++;
                
-              }
+              } else if (auto p=dynamic_cast<PlotWidget*>(it.get()))
```
highperformancecoder, line 374: This smells like it could be handled via an override.

model/plotSheet.h
```diff
@@ -0,0 +1,36 @@
...
+#define PLOTSHEET_H
+#include <parVarSheet.h>
+
+namespace minsky
+{
+	 
+  class PlotSheet: public ParVarSheet
+  {	  
+  public:
+    bool variableSelector(ItemPtr i) override {if (auto p=dynamic_cast<PlotWidget*>(i.get())) return p->plotOnTab(); return false;}
```
highperformancecoder, line 31: This smells like it should be a virtual function and an override.

gui-tk/minsky.tcl
```diff
@@ -953,7 +977,20 @@ proc variableContext {x y X Y} {
 		}
     }
     tk_popup .variables.context $X $Y
-}  
+}
+
+# Don't understand why this doesn't work???
```
highperformancecoder, line 982: What happens currently?

gui-tk/plots.tcl
```diff
@@ -160,6 +160,8 @@ proc doPlotOptions {plot} {
 proc plotDoubleClick {plotId} {
     toplevel .plot$plotId
     wm title .plot$plotId [$plotId.title]
+    
+    #pack [canvas .plots.canvas$plotId] -fill both -expand 1
```
highperformancecoder, line 164: Dead code?

model/parVarSheet.cc
```diff
@@ -359,7 +450,17 @@ namespace
...
                 iC++;
                
-              }
+              } else if (auto p=it->plotWidgetCast())
+		       {
+				cairo::CairoSave cs(cairo);   
+				if (it==itemFocus) {
+					cairo_translate(cairo,xItem,yItem);  		    				   
+                    itemCoords.erase(itemFocus);   
+                    itemCoords.emplace(make_pair(itemFocus,make_pair(xItem,yItem)));
```
highperformancecoder, line 459: I'm guessing that itemCoords is a map (or similar) type. Simplest code would be

itemCoords[itemFocus]=make_pair(xItem,yItem);

model/parVarSheet.h
```diff
@@ -56,12 +57,24 @@ namespace minsky
...
     /// row at \a y in unzoomed coordinates
-    int rowY(double y) const;    
+    int rowY(double y) const;
+    void moveTo(float x, float y);  
+         
+    float moveOffsX, moveOffsY,xItem,yItem;
+    ItemPtr itemFocus;        
     enum ClickType {background, internal};    
-    ClickType clickType(double x, double y) const;         
+    ClickType clickType(double x, double y);
```
highperformancecoder, line 66: Why is clickType not const?

model/plotSheet.h
```diff
@@ -0,0 +1,36 @@
...
+#define PLOTSHEET_H
+#include <parVarSheet.h>
+
+namespace minsky
+{
+	 
+  class PlotSheet: public ParVarSheet
+  {	  
+  public:
+    bool variableSelector(ItemPtr i) override {if (auto p=i->plotWidgetCast()) {return p->plotOnTab();}; return false;}
```
highperformancecoder, line 31: variableSelector is getting increasingly misnamed. Can be fixed on a later PR, though.

gui-tk/minsky.tcl
wdednam: Nothing at all. I wonder if should create a tabs.tcl file and source it in minsky.tcl?

gui-tk/plots.tcl
wdednam: Oops, I missed that, sorry.

model/parVarSheet.h
wdednam: Because I initially had itemAt(x,y) inside it instead of itemFocus, but I'm changing it now...

model/plotSheet.h
wdednam: Yes, I'm thinking about changing the ParVarSheet class to the ItemTab class, which means variableSheet will become variableTab, etc. I'll do so when I submit the Godley tab code.

PR #255, opened 2020-12-14
Feature 17

wdednam:
This change is [Reviewable](https://reviewable.io/reviews/highperformancecoder/minsky/255)


model/godleyTab.h
```diff
@@ -0,0 +1,36 @@
...
+#define GODLEYTAB_H
+#include <itemTab.h>
+
+namespace minsky
+{
+	 
+  class GodleyTab: public ItemTab
+  {	  
+  public:
+    bool itemSelector(ItemPtr i) override {if (auto g=dynamic_cast<GodleyIcon*>(i.get())) {return g->editorMode();}; return false;}
```
highperformancecoder, line 31: Wouldn't we want the godley table displayed always, not just when they're in "editor mode"?

model/itemTab.h
```diff
@@ -73,13 +73,13 @@ namespace minsky
     void mouseUp(float x, float y);
     void mouseMove(float x, float y);    
     ItemPtr itemAt(float x, float y);
-    void togglePlotDisplay();
+    void togglePlotDisplay() const;
     void displayDelayedTooltip(float x, float y);        
        
-    ~ParVarSheet() {}
+    ~ItemTab() {}
```
highperformancecoder, line 79: destructor should be marked virtual. I think CairoSurface has a virtual destructor, so this isn't a bug yet, but it is safest to always declare a virtual destructor if any other method is declared virtual. I don't know why C++ doesn't make this a feature of the language, maybe there some code out there that relies on non-virtual destructors in polymorphic types.

model/godleyTab.h
wdednam: Yes, that's true. I'll change it. I also forgot to assign the itemCoords object using the [] operator instead of emplace...

model/itemTab.h
wdednam: Done, thanks :)

PR #256, opened 2020-12-15
Feature 17

wdednam:
I tried using the simpler [] assign operator to populate itemCoords, but the coordinates do not persist when I do so and revert to the values on the original canvas after moving a plot or Godley table on the tab.

---
This change is [Reviewable](https://reviewable.io/reviews/highperformancecoder/minsky/256)

### DIFF
--- a/model/godleyTab.h
+++ b/model/godleyTab.h
@@ -28,7 +28,7 @@ namespace minsky
   class GodleyTab: public ItemTab
   {	  
   public:
-    bool itemSelector(ItemPtr i) override {if (auto g=dynamic_cast<GodleyIcon*>(i.get())) {return g->editorMode();}; return false;}    
+    bool itemSelector(ItemPtr i) override {if (auto g=dynamic_cast<GodleyIcon*>(i.get())) return true; return false;}    
   };
   
 }
--- a/model/godleyTab.h
+++ b/model/godleyTab.h
@@ -28,7 +28,7 @@ namespace minsky
   class GodleyTab: public ItemTab
   {	  
   public:
-    bool itemSelector(ItemPtr i) override {if (auto g=dynamic_cast<GodleyIcon*>(i.get())) return true; return false;}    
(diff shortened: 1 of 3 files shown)


wdednam:
I either have to use the Godley table in its editor form, then it has its title and can be selected on the Godley tab very precisely, but then the Godley table also has to be in editor mode on the main canvas... OR, I can use GodleyTableWindow instead, via a "new GodleyTableWindow". But in this case I can't get the title positioned nicely above the table or get its dimensions correct in itemAt(x,y), even though the table displays nicely. I'm stumped...

model/godleyTab.h
```diff
@@ -28,7 +28,7 @@ namespace minsky
   class GodleyTab: public ItemTab
   {	  
   public:
-    bool itemSelector(ItemPtr i) override {if (auto g=dynamic_cast<GodleyIcon*>(i.get())) {return g->editorMode();}; return false;}    
+    bool itemSelector(ItemPtr i) override {if (dynamic_cast<GodleyIcon*>(i.get())) return true; return false;}
```
highperformancecoder, line 31: Or even simpler: "return i;"

model/itemTab.cc
```diff
@@ -458,19 +459,21 @@ namespace minsky
...
                   {
                     cairo::CairoSave cs(cairo);   
                     if (it==itemFocus) {
                       cairo_translate(cairo,xItem,yItem);  		    				   
                       itemCoords.erase(itemFocus);   
                       itemCoords.emplace(make_pair(itemFocus,make_pair(xItem,yItem)));         
-                    } else cairo_translate(cairo,itemCoords[it].first,itemCoords[it].second);      
-                    g->draw(cairo);
+                    } else cairo_translate(cairo,itemCoords[it].first,itemCoords[it].second);
+                    auto godley=GodleyTableWindow(g);
```
highperformancecoder, line 474: Cleaner would be GodleyTableWindow godley(g); The above may invoke an unnecessary copy constructor.

PR #257, opened 2020-12-17
Feature 17

wdednam:
This change is [Reviewable](https://reviewable.io/reviews/highperformancecoder/minsky/257)

### DIFF
--- a/model/godleyTab.h
+++ b/model/godleyTab.h
@@ -28,7 +28,7 @@ namespace minsky
   class GodleyTab: public ItemTab
   {	  
   public:
-    bool itemSelector(ItemPtr i) override {if (dynamic_cast<GodleyIcon*>(i.get())) return true; return false;}    
+    bool itemSelector(ItemPtr i) override {if (dynamic_cast<GodleyIcon*>(i.get())) return i.get(); return false;}
   };
   
 }
--- a/model/itemTab.cc
+++ b/model/itemTab.cc
@@ -148,10 +148,30 @@ namespace minsky
     auto minD=numeric_limits<float>::max();
     for (auto& i: itemCoords)
       {
-        float d=sqr((i.second).first+offsx-x)+sqr((i.second).second+offsy-y);
+		float xx=(i.second).first+offsx, yy=(i.second).second+offsy;  
+        float d=sqr(xx-x)+sqr(yy-y);
         float z=i.first->zoomFactor();
         float w=0.5*i.first->iWidth()*z,h=0.5*i.first->iHeight()*z;
-        if (d<minD && fabs((i.second).first+offsx-x)<w && fabs((i.second).second+offsy-y)<h)
+        // improve grabbing of Godley tables on the tab.
+        if (auto g=dynamic_pointer_cast<GodleyIcon>(i.first))
+        {
+			GodleyTableWindow godley(g);
+            ecolab::cairo::Surface surf
+               (cairo_recording_surface_create(CAIRO_CONTENT_COLOR_ALPHA,NULL));			
+            try
+              {
+                godley.draw(surf.cairo());
+              }
+            catch (const std::exception& e) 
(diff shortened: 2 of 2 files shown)


model/godleyTab.h
```diff
@@ -28,7 +28,7 @@ namespace minsky
   class GodleyTab: public ItemTab
   {	  
   public:
-    bool itemSelector(ItemPtr i) override {if (dynamic_cast<GodleyIcon*>(i.get())) return true; return false;}    
+    bool itemSelector(ItemPtr i) override {if (dynamic_cast<GodleyIcon*>(i.get())) return i.get(); return false;}
```
highperformancecoder, line 31: I was thinking more `return dynamic_cast<GodleyIcon*>(i.get());` The if construct is redundant.

model/itemTab.cc
```diff
@@ -148,10 +148,30 @@ namespace minsky
...
-        float d=sqr((i.second).first+offsx-x)+sqr((i.second).second+offsy-y);
+        float xx=(i.second).first+offsx, yy=(i.second).second+offsy;  
+        float d=sqr(xx-x)+sqr(yy-y);
         float z=i.first->zoomFactor();
         float w=0.5*i.first->iWidth()*z,h=0.5*i.first->iHeight()*z;
-        if (d<minD && fabs((i.second).first+offsx-x)<w && fabs((i.second).second+offsy-y)<h)
+        // improve grabbing of Godley tables on the tab.
+        if (auto g=dynamic_pointer_cast<GodleyIcon>(i.first))
+          {
+            GodleyTableWindow godley(g);
```
highperformancecoder, line 158: This seems a little heavyweight for itemAt, which is potentially called many times as a mouse is moved over the canvas.

I would have thought constructing the GodleyTableWindows and leaving them extent for the lifetime of the tab would be the way to go - that way they only need to be rendered when the tab is rendered, not every time the mouse moves.

model/itemTab.cc
```diff
@@ -471,9 +491,21 @@ namespace minsky
                       itemCoords.erase(itemFocus);   
                       itemCoords.emplace(make_pair(itemFocus,make_pair(xItem,yItem)));         
                     } else cairo_translate(cairo,itemCoords[it].first,itemCoords[it].second);
-                    auto godley=GodleyTableWindow(g);
-                    godley.disableButtons();   
+                    GodleyTableWindow godley(g);
```
highperformancecoder, line 494: Like my last comment - GodleyTableWindow is a little heavyweight for the draw method too.

My suggestion is to add another GodleyTableWindow member to GodleyIcon, like I did for the editor  mode

highperformancecoder: Also lines 476 and 486 are a code smell that virtual function overrides should be used.

wdednam: > Like my last comment - GodleyTableWindow is a little heavyweight for the draw method too.
> 
> My suggestion is to add another GodleyTableWindow member to GodleyIcon, like I did for the editor mode

I tried this, but the best I get is when I declare a public "CopiableUniquePtr godley" in class GodlyeIcon, but which when I call it in ItemTab, returns a nullptr pointer and gives a segfault. 

I also tried initializing "CopiableUniquePtr godley" with *this in GodleyIcon, but no cigar, I'm afraid.

Then I tried adding AnotherCopiableUniquePtr with GodleyTableWindow as the base class, but this would not compile at all either.

Outside of simple types and classes, I really struggle with inheritance and polymorphism... :(

wdednam: > Also lines 476 and 486 are a code smell that virtual function overrides should be used.

What would be the return type of the overridden members in plotTab and godleyTab? ItemPtr?

Again, when the target function is of type void, bool, string, double, int, etc., I understand how to deal with virtual functions, but I don't know how to specialize to more general user defined classes or types.

model/godleyTab.h
wdednam: Done, thank you :)

PR #258, opened 2020-12-18
Fix vector integration

wdednam:
This change is [Reviewable](https://reviewable.io/reviews/highperformancecoder/minsky/258)

### DIFF
--- a/tensor/interpolateHypercube.cc
+++ b/tensor/interpolateHypercube.cc
@@ -76,9 +76,16 @@ namespace civita
             auto dst=find_if(targetHC.begin(), targetHC.end(),
                              [&](const XVector& i){return i.name==src.name;});
             if (dst==targetHC.end())
-              throw runtime_error("no matching dimension found");
-            interimHC.xvectors.push_back(*dst);
-            rotation[dst-targetHC.begin()]=i;
+            {
+			  // possible alternative when targetHC has no xvectors or undefined ones. for feature 147
+              interimHC.xvectors.push_back(src);
+              rotation[i]=i;
+		    }
+		    else
+		    {
+		       interimHC.xvectors.push_back(*dst);
+               rotation[dst-targetHC.begin()]=i;
+		    }
           }
         strides.push_back(stride);
         stride*=targetHC[i].size();
--- a/tensor/interpolateHypercube.cc
+++ b/tensor/interpolateHypercube.cc
@@ -71,7 +71,7 @@ namespace civita
                 src.dimension.units!=dst.dimension.units) //TODO handle conversion between different units
               throw runtime_error("mismatch between unnamed dimensions");
             interimHC.xvectors.push_back(dst);
-            tmpRotation[i]=i;
+            tmpRotation.emplace(make_pair(i,i));
           }
         else // find matching names dimension
           {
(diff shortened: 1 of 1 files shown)


tensor/interpolateHypercube.cc
```diff
@@ -76,9 +76,16 @@ namespace civita
...
                              [&](const XVector& i){return i.name==src.name;});
             if (dst==targetHC.end())
-              throw runtime_error("no matching dimension found");
-            interimHC.xvectors.push_back(*dst);
-            rotation[dst-targetHC.begin()]=i;
+            {
+              // possible alternative when targetHC has no xvectors or undefined ones. for feature 147
+              interimHC.xvectors.push_back(src);
+              rotation[i]=i;
+            }
```
highperformancecoder, line 83: I would be OK with this, provided you can guarantee that rotation is a permutation (ie it is filled with the numbers 0..rotation.size(), each appearing only once in the sequence, but not necessarily in order).

I don't think the current code guarantees that, as your chosen position for the ith vector may get overwritten later on on line 87, leading to one of the spots in the vector being uninitialised, or left at its default value of 0.

One solution might be to check that the xvector dimension name appearing as targetHC[i] does not appear in the src, and to choose another index if it is found to be. This is hardly efficient or optimal though - perhaps you have a special use case that might be handled by a different code branch.

wdednam: > I don't think the current code guarantees that, as your chosen position for the ith vector may get overwritten later on on line 87, leading to one of the spots in the vector being uninitialised, or left at its default value of 0.

Ah, I see that now too.

> perhaps you have a special use case that might be handled by a different code branch.

Well, any time one builds a model in which the variables have tensor parameters as initial values this will occur.

I tried to change the dimensions in Edit-> Dimensions after checking that they were mismatched in "dimensions" when browsing the object, but to no avail. If it is possible, it is not simple and the user will likely struggle like I did. Then I also tried to format my CSV file such that the dimensions are (0.000000, 1.000000, 2.000000, etc.,) as when a parameter is initialised using a zeros(5), iota (5),eye(5), etc. But the dimensions were still mismatched and my model wouldn't run.

I might have to go with the solution you suggest, which I will work on now. Yesterday, I tried to improve the other PR.

tensor/interpolateHypercube.cc
```diff
@@ -69,20 +71,28 @@ namespace civita
...
+                interimHC.xvectors.push_back(*dst);
+                tmpRotation[dst-targetHC.begin()]=i;
+              }
           }
         strides.push_back(stride);
         stride*=targetHC[i].size();
       }
+    for (size_t i=0; i< tmpRotation.size(); i++) rotation[i]=tmpRotation[i];  
 #ifndef NDEBUG
     for (auto& i: rotation) assert(i<rank()); // check that no indices have been doubly assigned.
```
highperformancecoder, line 97: For it to be a permutation, tmpRotation.size()==rank() must also be true, and this is cheaper than checking all rotation elements are < rank(). I think we need to throw in the case it is not a permutation.

Also line 95 can be replaced by for (auto& i: tmpRotation) rotation[i.first]=i.second, or alternatively, use push_back() (but need to remove line 59 first). It is cheaper to iterate over the elements of a map (which will in key order), than it is to look up the map elements each time around the loop (roughly O(N) instead of O(N log N).

tensor/interpolateHypercube.cc
```diff
@@ -69,20 +71,28 @@ namespace civita
...
-            rotation[dst-targetHC.begin()]=i;
+            {
+              // possible alternative when targetHC has no xvectors or undefined ones. for feature 147
+              interimHC.xvectors.push_back(src);
+              tmpRotation[i]=i;
+            }
+            else
+              {
+                interimHC.xvectors.push_back(*dst);
+                tmpRotation[dst-targetHC.begin()]=i;
```
highperformancecoder, line 89: using emplace is cheaper here and lines 74 and 84, as you don't need to search for an existing key first. emplace will not overwrite an existing value, but we don't care about that, as we're going to throw in that circumstance.

tensor/interpolateHypercube.cc
wdednam: Done and done :)

tensor/interpolateHypercube.cc
wdednam: Done, thanks :)

highperformancecoder: Thanks - looks good, at least to visual inspection.

PR #259, opened 2020-12-21
again, not sure this is really much more efficient than what went bef…

wdednam:
…ore. don't know what else to virtualise...

---
This change is [Reviewable](https://reviewable.io/reviews/highperformancecoder/minsky/259)

### DIFF
--- a/model/canvas.cc
+++ b/model/canvas.cc
@@ -837,6 +837,10 @@ namespace minsky
       {
         redrawUpdateRegion();
       }
+    catch (std::exception& ex)
+      {
+        cerr << ex.what() << endl;
+      }
     catch (...)
       {
         // consume exception and try redrawing
--- a/model/canvas.cc
+++ b/model/canvas.cc
@@ -503,7 +503,22 @@ namespace minsky
            });
 	}      
     itemVector.clear();  
-  }  
+  }
+  
+  void Canvas::showPlotsOnTab()
+  {
+    if (itemVector.empty())
+	{
+        model->recursiveDo
+          (&GroupItems::items, [&](const Items&,Items::const_iterator i)
+           {
+             if (auto p=(*i)->plotWidgetCast())
+               if (p->plotOnTab()==false)
+				 p->togglePlotTabDisplay();	 
+             return false;
+           });
+	} 
+  }    
   
   void Canvas::deleteItem()
   {
--- a/model/canvas.cc
+++ b/model/canvas.cc
@@ -507,17 +507,13 @@ namespace minsky
   
   void Canvas::showPlotsOnTab()
   {
-    if (itemVector.empty())
-	{
-        model->recursiveDo
-          (&GroupItems::items, [&](const Items&,Items::const_iterator i)
-           {
-             if (auto p=(*i)->plotWidgetCast())
-               if (p->plotOnTab()==false)
-				 p->togglePlotTabDisplay();	 
-             return false;
-           });
-	} 
+     model->recursiveDo
+       (&GroupItems::items, [&](const Items&,Items::const_iterator i)
+        {
+          if (auto p=(*i)->plotWidgetCast())
+            if (p->plotOnTab()==false) p->togglePlotTabDisplay();	 
+          return false;
+        });
   }    
   
   void Canvas::deleteItem()
--- a/model/godleyIcon.h
+++ b/model/godleyIcon.h
@@ -80,8 +80,6 @@ namespace minsky
     void setEditorDisplayValues();
     
     CopiableUniquePtr godleyT;
-    
-    std::string classType() const override {return "GodleyIcon";}    
 
     /// scale icon until it's height or width matches \a h or \a w depending on which is minimum             
     void scaleIcon(float w, float h);         
--- a/model/godleyIcon.h
+++ b/model/godleyIcon.h
@@ -78,6 +78,10 @@ namespace minsky
 
     /// sets editor's display values attributes to current global preferences
     void setEditorDisplayValues();
+    
+    CopiableUniquePtr godleyT;
+    
+    std::string classType() const override {return "GodleyIcon";}    
 
     /// scale icon until it's height or width matches \a h or \a w depending on which is minimum             
     void scaleIcon(float w, float h);         
--- a/model/godleyTab.cc
+++ b/model/godleyTab.cc
@@ -1,0 +1,110 @@
+/*
+  @copyright Steve Keen 2020
+  @author Russell Standish
+  @author Wynand Dednam
+  This file is part of Minsky.
+
+  Minsky is free software: you can redistribute it and/or modify it
+  under the terms of the GNU General Public License as published by
+  the Free Software Foundation, either version 3 of the License, or
+  (at your option) any later version.
+
+  Minsky is distributed in the hope that it will be useful,
+  but WITHOUT ANY WARRANTY; without even the implied warranty of
+  MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE.  See the
+  GNU General Public License for more details.
+
+  You should have received a copy of the GNU General Public License
+  along with Minsky.  If not, see <http://www.gnu.org/licenses/>.
+*/
+#include "godleyTab.h"
+#include "godleyTableWindow.h"
+#include <pango.h>
+#include "minsky_epilogue.h"
+#include "minsky.h"
+using namespace std;
+using ecolab::cairo::Surface;
+using ecolab::Pango;
+using ecolab::cairo::CairoSave;
+
+namespace minsky
+{
+  
+  ItemPtr GodleyTab::itemAt(float x, float y)
+  {
+    ItemPtr item;                    
+    auto minD=numeric_limits<float>::max();
+    for (auto& i: itemCoords)
+      {
+		if (i.first->classType()=="GodleyIcon")
+		{
+          float xx=(i.second).first+offsx, yy=(i.second).second+offsy;  
+		  auto g=dynamic_pointer_cast<GodleyIcon>(i.first);  
+		  g->godleyT.reset(new GodleyTableEditor(g));
+          ecolab::cairo::Surface surf
+            (cairo_recording_surface_create(CAIRO_CONTENT_COLOR_ALPHA,NULL));			
+          try
+            {
+              g->godleyT.get()->draw(surf.cairo());
+            }
+          catch (const std::exception& e) 
+            {cerr<<"illegal exception caught in draw(): "<<e.what()<<endl;}
+          catch (...) {cerr<<"illegal exception caught in draw()";}
+          float w=0.5*g->godleyT.get()->colLeftMargin[g->godleyT.get()->colLeftMargin.size()-1];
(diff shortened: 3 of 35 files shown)


model/canvas.cc
```diff
@@ -503,7 +503,22 @@ namespace minsky
            });
 	}      
     itemVector.clear();  
-  }  
+  }
+  
+  void Canvas::showPlotsOnTab()
+  {
+    if (itemVector.empty())
```
highperformancecoder, line 510: Why would you only do this if itemVector is empty? Wouldn't the user expect that all remaining plots be shown on the tab in this action?

model/canvas.cc
```diff
@@ -503,7 +503,22 @@ namespace minsky
...
+	{
+        model->recursiveDo
+          (&GroupItems::items, [&](const Items&,Items::const_iterator i)
+           {
+             if (auto p=(*i)->plotWidgetCast())
+               if (p->plotOnTab()==false)
+				 p->togglePlotTabDisplay();
```
highperformancecoder, line 517: This is something you can virtualise as tabDisplay or toggleTabDisplay (as appropriate), which would be an empty implementation in Item, and overridden in variable, godleyIcon and plotWidget.

model/itemTab.cc
```diff
@@ -41,7 +41,7 @@ namespace minsky
                                          if (itemSelector(*i)) 
                                            {		                                 
                                              itemVector.emplace_back(*i);
-                                             if (auto p=(*i)->plotWidgetCast()) itemCoords.emplace(make_pair(*i,make_pair(p->x(),p->y()))); 
+                                             if ((*i)->classType()=="PlotWidget") itemCoords.emplace(make_pair(*i,make_pair((*i)->x(),(*i)->y())));
```
highperformancecoder, line 44: plotWidgetCast() was better. String comparisons are more expensive.

highperformancecoder: IIRC, itemSelector is virtual, and returns true only if i points to the right item type. So the type test on this line and next is redundant. Just insert the coordinates into the itemCoords vector for all such tabbed items, even though we're not going to use them in the variable case.

model/plotTab.cc
```diff
@@ -0,0 +1,70 @@
...
+#include "minsky.h"
+using namespace std;
+using ecolab::cairo::Surface;
+using ecolab::Pango;
+using ecolab::cairo::CairoSave;
+
+namespace minsky
+{
+
+  bool PlotTab::itemSelector(ItemPtr i)
```
highperformancecoder, line 33: const ref the shared_ptr

model/plotTab.cc
```diff
@@ -0,0 +1,70 @@
...
+                      cairo_translate(cairo,xItem,yItem);  		    				   
+                      itemCoords.erase(itemFocus);   
+                      itemCoords.emplace(make_pair(itemFocus,make_pair(xItem,yItem)));
+                    } else cairo_translate(cairo,itemCoords[it].first,itemCoords[it].second);      
+                    p->draw(cairo);
+                  }
+              }              
+          }
+      }
+    catch (...) {throw;/* exception most likely invalid variable value */}
```
highperformancecoder, line 67: You may as well not have the try/catch if you're just going to rethrow the exception.

model/plotTab.cc
```diff
@@ -0,0 +1,70 @@
...
+  bool PlotTab::itemSelector(ItemPtr i)
+  {
+	if (auto p=i->plotWidgetCast()) return p->plotOnTab();
+	return false;
+  }	
+
+  void PlotTab::togglePlotDisplay() const      
+  {
+    if (auto p=itemFocus->plotWidgetCast()) p->togglePlotTabDisplay();
+    else return;
```
highperformancecoder, line 42: redundant line.

model/plotWidget.h
```diff
@@ -86,7 +86,9 @@ namespace minsky
     float height() const {return Item::height();}
     
     const PlotWidget* plotWidgetCast() const override {return this;}
-    PlotWidget* plotWidgetCast() override {return this;}        
+    PlotWidget* plotWidgetCast() override {return this;}
+    
+    std::string classType() const override {return "PlotWidget";}
```
highperformancecoder, line 91: Shouldn't be necessary. ItemT and classdesc ensure this happens anyway, automatically for all such ItemT derived types.

model/plotTab.cc
```diff
@@ -0,0 +1,70 @@
...
+  }	
+	 
+  void PlotTab::draw(cairo_t* cairo)
+  {   
+    try
+      {	
+      		
+        if (!itemVector.empty())
```
highperformancecoder, line 50: This check is redundant. The following for loop is equivalent when the vector is empty.

model/plotTab.cc
```diff
@@ -0,0 +1,70 @@
...
+          {
+            for (auto& it: itemVector)
+              {
+                if (auto p=it->plotWidgetCast())
```
highperformancecoder, line 54: All items in itemVector should be plotWidgets, so this test is redundant. Assert it if you must.

model/plotTab.cc
```diff
@@ -0,0 +1,70 @@
...
+                  {
+                    cairo::CairoSave cs(cairo);   
+                    if (it==itemFocus) {
+                      cairo_translate(cairo,xItem,yItem);  		    				   
+                      itemCoords.erase(itemFocus);   
+                      itemCoords.emplace(make_pair(itemFocus,make_pair(xItem,yItem)));
```
highperformancecoder, line 60: More efficient to do itemCoords[itemFocus]=move(make_pair(xItem,yItem))

highperformancecoder:
There's now some conflicts to sort out as well!

highperformancecoder:
Sorry accidentally closed this.

model/godleyIcon.h
```diff
@@ -78,6 +78,10 @@ namespace minsky
 
     /// sets editor's display values attributes to current global preferences
     void setEditorDisplayValues();
+    
+    CopiableUniquePtr godleyT;
+    
+    std::string classType() const override {return "GodleyIcon";}
```
highperformancecoder, line 84: Oops - you removed this in Feature 17, but this PR looks like it will reinstate it.

model/godleyTab.cc
```diff
@@ -0,0 +1,110 @@
...
+  {
+    ItemPtr item;                    
+    auto minD=numeric_limits<float>::max();
+    for (auto& i: itemCoords)
+      {
+		if (i.first->classType()=="GodleyIcon")
+		{
+          float xx=(i.second).first+offsx, yy=(i.second).second+offsy;  
+		  auto g=dynamic_pointer_cast<GodleyIcon>(i.first);  
+		  g->godleyT.reset(new GodleyTableEditor(g));
```
highperformancecoder, line 43: Need to qualify this with if (!g->godleyT)

model/godleyTab.cc
```diff
@@ -0,0 +1,110 @@
...
+          ecolab::cairo::Surface surf
+            (cairo_recording_surface_create(CAIRO_CONTENT_COLOR_ALPHA,NULL));			
+          try
+            {
+              g->godleyT.get()->draw(surf.cairo());
```
highperformancecoder, line 48: No need for .get(()

model/godleyTab.cc
```diff
@@ -0,0 +1,110 @@
...
+					g->godleyT.reset(new GodleyTableEditor(g));  
+                    cairo::CairoSave cs(cairo);   
+                    if (it==itemFocus) {
+                      cairo_translate(cairo,xItem,yItem);  		    				   
+                      itemCoords.erase(itemFocus);   
+                      itemCoords.emplace(make_pair(itemFocus,make_pair(xItem,yItem)));         
+                    } else cairo_translate(cairo,itemCoords[it].first,itemCoords[it].second);
+                    g->godleyT.get()->disableButtons();
+                    g->godleyT.get()->displayValues=true;   
+                    g->godleyT.get()->draw(cairo);
```
highperformancecoder, line 91: ditto - no need for .get()

model/godleyTab.h
```diff
@@ -28,7 +28,9 @@ namespace minsky
   class GodleyTab: public ItemTab
   {	  
   public:
-    bool itemSelector(ItemPtr i) override {if (dynamic_cast<GodleyIcon*>(i.get())) return i.get(); return false;}
+    bool itemSelector(const ItemPtr i) override {if (i->classType()=="GodleyIcon") return true; return false;}
```
highperformancecoder, line 31: Should be return dynamic_cast<GodleyIcon*>(i.get())

model/itemTab.h
```diff
@@ -48,8 +48,8 @@ namespace minsky
     /// computed positions of the variable rows
     std::vector<double> rowTopMargin;                      
 
-    void populateItemVector();
-    virtual bool itemSelector(ItemPtr i) = 0;
+    virtual void populateItemVector();
+    virtual bool itemSelector(const ItemPtr i) = 0;
```
highperformancecoder, line 52: Should be const ItemPtr&  (aka const ref).

model/parameterTab.h
```diff
@@ -28,7 +28,7 @@ namespace minsky
   class ParameterTab: public ItemTab
   {
   public:
-    bool itemSelector(ItemPtr i) override {if (auto v=i->variableCast()) return v->type()==VariableType::parameter; return false;} 
+    bool itemSelector(const ItemPtr i) override {if (auto v=i->variableCast()) return v->type()==VariableType::parameter; return false;}
```
highperformancecoder, line 31: Ditto

model/plotTab.cc
```diff
@@ -0,0 +1,59 @@
...
+	 
+  void PlotTab::draw(cairo_t* cairo)
+  {   
+     for (auto& it: itemVector)
+       {
+         assert(it->plotWidgetCast());
+         cairo::CairoSave cs(cairo);   
+         if (it==itemFocus) {
+           cairo_translate(cairo,xItem,yItem);  		    				   
+           itemCoords.erase(itemFocus);
```
highperformancecoder, line 52: No need to erase here.

model/plotTab.h
```diff
@@ -28,7 +28,9 @@ namespace minsky
   class PlotTab: public ItemTab
   {	  
   public:
-    bool itemSelector(ItemPtr i) override {if (auto p=i->plotWidgetCast()) {return p->plotOnTab();}; return false;}    
+    bool itemSelector(const ItemPtr i) override;
```
highperformancecoder, line 31: const ref

model/variableTab.h
```diff
@@ -28,7 +28,7 @@ namespace minsky
   class VariableTab: public ItemTab
   {	  
   public:
-    bool itemSelector(ItemPtr i) override {if (auto v=i->variableCast()) return v->type()!=VariableType::parameter && v->attachedToDefiningVar(); return false;}  
+    bool itemSelector(const ItemPtr i) override {if (auto v=i->variableCast()) return v->type()!=VariableType::parameter && v->attachedToDefiningVar(); return false;}
```
highperformancecoder, line 31: const ref

highperformancecoder:
Now that I've merged Feature 17, it looks like this PR will cause regressions. Could you please merge master back into this branch, fix up the conflicts and check my comments above

PR #260, opened 2020-12-21
Feature 17

wdednam:
This change is [Reviewable](https://reviewable.io/reviews/highperformancecoder/minsky/260)

### DIFF
--- a/model/godleyTab.cc
+++ b/model/godleyTab.cc
@@ -1,0 +1,110 @@
+/*
+  @copyright Steve Keen 2020
+  @author Russell Standish
+  @author Wynand Dednam
+  This file is part of Minsky.
+
+  Minsky is free software: you can redistribute it and/or modify it
+  under the terms of the GNU General Public License as published by
+  the Free Software Foundation, either version 3 of the License, or
+  (at your option) any later version.
+
+  Minsky is distributed in the hope that it will be useful,
+  but WITHOUT ANY WARRANTY; without even the implied warranty of
+  MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE.  See the
+  GNU General Public License for more details.
+
+  You should have received a copy of the GNU General Public License
+  along with Minsky.  If not, see <http://www.gnu.org/licenses/>.
+*/
+#include "godleyTab.h"
+#include "godleyTableWindow.h"
+#include <pango.h>
+#include "minsky_epilogue.h"
+#include "minsky.h"
+using namespace std;
+using ecolab::cairo::Surface;
+using ecolab::Pango;
+using ecolab::cairo::CairoSave;
+
+namespace minsky
+{
+  
+  ItemPtr GodleyTab::itemAt(float x, float y)
+  {
+    ItemPtr item;                    
+    auto minD=numeric_limits<float>::max();
+    for (auto& i: itemCoords)
+      {
+		if (i.first->classType()=="GodleyIcon")
+		{
+          float xx=(i.second).first+offsx, yy=(i.second).second+offsy;  
+		  auto g=dynamic_pointer_cast<GodleyIcon>(i.first);  
+		  g->godleyT.reset(new GodleyTableEditor(g));
+          ecolab::cairo::Surface surf
+            (cairo_recording_surface_create(CAIRO_CONTENT_COLOR_ALPHA,NULL));			
+          try
+            {
+              g->godleyT.get()->draw(surf.cairo());
+            }
+          catch (const std::exception& e) 
+            {cerr<<"illegal exception caught in draw(): "<<e.what()<<endl;}
+          catch (...) {cerr<<"illegal exception caught in draw()";}
+          float w=0.5*g->godleyT.get()->colLeftMargin[g->godleyT.get()->colLeftMargin.size()-1];
+          float h=0.5*(g->godleyT.get()->godleyIcon->table.rows())*g->godleyT.get()->rowHeight;
+          xx+=w;
+          yy+=h;
+          float d=sqr(xx-x)+sqr(yy-y);
+          
+          if (d<minD && fabs(xx-x)<w && fabs(yy-y)<h)
+            {
+              minD=d;
+              item=i.first;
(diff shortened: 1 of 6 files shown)


model/godleyTab.cc
```diff
@@ -0,0 +1,110 @@
...
+  ItemPtr GodleyTab::itemAt(float x, float y)
+  {
+    ItemPtr item;                    
+    auto minD=numeric_limits<float>::max();
+    for (auto& i: itemCoords)
+      {
+		if (i.first->classType()=="GodleyIcon")
+		{
+          float xx=(i.second).first+offsx, yy=(i.second).second+offsy;  
+		  auto g=dynamic_pointer_cast<GodleyIcon>(i.first);
```
highperformancecoder, line 42: Move this to line 39 and use as the test. It's got to be quicker than string comparison.

model/godleyTab.cc
```diff
@@ -0,0 +1,110 @@
...
+  {
+    ItemPtr item;                    
+    auto minD=numeric_limits<float>::max();
+    for (auto& i: itemCoords)
+      {
+		if (i.first->classType()=="GodleyIcon")
+		{
+          float xx=(i.second).first+offsx, yy=(i.second).second+offsy;  
+		  auto g=dynamic_pointer_cast<GodleyIcon>(i.first);  
+		  g->godleyT.reset(new GodleyTableEditor(g));
```
highperformancecoder, line 43: Shouldn't need to do this every time. Perhaps do
if (!g->godleyT) g->godleyT.reset(new ...)

model/godleyTab.cc
```diff
@@ -0,0 +1,110 @@
...
+          ecolab::cairo::Surface surf
+            (cairo_recording_surface_create(CAIRO_CONTENT_COLOR_ALPHA,NULL));			
+          try
+            {
+              g->godleyT.get()->draw(surf.cairo());
```
highperformancecoder, line 48: get() shouldn't be required.

model/godleyTab.cc
```diff
@@ -0,0 +1,110 @@
...
+            }
+          catch (const std::exception& e) 
+            {cerr<<"illegal exception caught in draw(): "<<e.what()<<endl;}
+          catch (...) {cerr<<"illegal exception caught in draw()";}
+          float w=0.5*g->godleyT.get()->colLeftMargin[g->godleyT.get()->colLeftMargin.size()-1];
+          float h=0.5*(g->godleyT.get()->godleyIcon->table.rows())*g->godleyT.get()->rowHeight;
+          xx+=w;
+          yy+=h;
+          float d=sqr(xx-x)+sqr(yy-y);
+
```
highperformancecoder, line 58: Another thought. What about stashing a list of GodleyIcons in the tab, and initialise them in editor mode? That way you can use the existing bounding box code.
The only slight wrinkle is where the global minsky object is referenced - but I don't think that will cause any problems for this application.

model/godleyTab.cc
```diff
@@ -0,0 +1,110 @@
...
+      {	
+      		
+        if (!itemVector.empty())
+          {               
+            for (auto& it: itemVector)
+              {
+                if (it->classType()=="GodleyIcon")
+                  {
+					auto g=dynamic_pointer_cast<GodleyIcon>(it);
+					g->godleyT.reset(new GodleyTableEditor(g));
```
highperformancecoder, line 82: See comment above. We're trying to get away from creating one these every draw() call.

model/godleyTab.h
```diff
@@ -28,7 +28,9 @@ namespace minsky
   class GodleyTab: public ItemTab
   {	  
   public:
-    bool itemSelector(ItemPtr i) override {if (dynamic_cast<GodleyIcon*>(i.get())) return i.get(); return false;}
+    bool itemSelector(ItemPtr i) override {if (i->classType()=="GodleyIcon") return true; return false;}
```
highperformancecoder, line 31: Wouldn't return dynamic_cast<GodleyIcon*>(i.get())) be better?

model/itemTab.cc
```diff
@@ -42,7 +41,7 @@ namespace minsky
                                            {		                                 
                                              itemVector.emplace_back(*i);
                                              if (auto p=(*i)->plotWidgetCast()) itemCoords.emplace(make_pair(*i,make_pair(p->x(),p->y()))); 
-                                             if (auto g=dynamic_cast<GodleyIcon*>(i->get())) itemCoords.emplace(make_pair(*i,make_pair(g->x(),g->y())));
+                                             if ((*i)->classType()=="GodleyIcon") itemCoords.emplace(make_pair(*i,make_pair((*i)->x(),(*i)->y())));
```
highperformancecoder, line 44: ditto

model/itemTab.h
```diff
@@ -48,7 +48,7 @@ namespace minsky
     /// computed positions of the variable rows
     std::vector<double> rowTopMargin;                      
 
-    void populateItemVector();
+    virtual void populateItemVector();
```
highperformancecoder, line 51: Don't forget to add a virtual destructor. It will be virtual anyway, because CairoSurface has a virtual destructor, but such things are easily overlooked if ever the inheritance hierarchy is changed,

wdednam: Haven't I already?  See   virtual ~ItemTab() {} on line 79

model/godleyTab.cc
wdednam: Got rid of all string comparisons, thanks :)

model/godleyTab.cc
wdednam: Thank you, I've implemented this.

model/godleyTab.cc
wdednam: Great idea, thanks :) I will do this in a subsequent PR. I would like to fix the column widths of rank 0 pars and vars first.

model/godleyTab.h
wdednam: Reverted to the dynamic_cast, thanks :)

PR #261, opened 2020-12-21
Still need to fix column widths on variables tab.

wdednam:
This change is [Reviewable](https://reviewable.io/reviews/highperformancecoder/minsky/261)


model/itemTab.cc
```diff
@@ -244,35 +246,38 @@ namespace minsky
...
-                          }
+                        if (it==itemVector[0]) {
+                          for (auto& i:varAttrib) 
+                            {
+                              cairo_move_to(cairo,x,y-1.5*rowHeight);                    
+                              pango.setMarkup(i);
+                              pango.show();                  
+                              colWidth=std::max(colWidth,5+pango.width());  
+                              x+=colWidth;	
+                              colLeftMargin[iC].push_back(x);
```
highperformancecoder, line 257: Wouldn't this be a right margin?

highperformancecoder: How is iC calculated?

model/itemTab.cc
```diff
@@ -244,35 +246,38 @@ namespace minsky
...
-                        for (auto& i:varAttrib) 
-                          {
-                            cairo_move_to(cairo,x,y-1.5*rowHeight);                    
-                            pango.setMarkup(i);
-                            pango.show();                  
-                            colWidth=std::max(colWidth,5+pango.width());  
-                            x+=colWidth;	
-                            colLeftMargin[iC].push_back(x);                        				    
-                          }
+                        if (it==itemVector[0]) {
```
highperformancecoder, line 249: OK, because itemVector no doubt has unique elements, but better to compare the address (&it==&itemVector[0]).

model/itemTab.cc
```diff
@@ -244,35 +246,38 @@ namespace minsky
...
-                            x+=colWidth;		
+                            x+=colWidth;
                           }
                         x=x0;                      
                         h_prev=h;
                         w=0;h=0;      
                         cairo_get_current_point (cairo,&w,&h);   
                         if (h<h_prev) h+=h_prev;                                                                         
                         // draw grid
+                        float y1=it==itemVector[0]?-1.5*rowHeight: rowHeight;
```
highperformancecoder, line 275: Ditto re comparing addresses.

model/itemTab.cc
```diff
@@ -282,23 +287,27 @@ namespace minsky
                         { // draw vertical grid lines
                           cairo::CairoSave cs(cairo);
                           cairo_set_source_rgba(cairo,0,0,0,0.5);
+                          y1=it==itemVector[0]? 0.5*rowHeight: 0;
```
highperformancecoder, line 290: Ditto re addresses.

model/itemTab.cc
```diff
@@ -282,23 +287,27 @@ namespace minsky
...
-                        { // draw horizontal grid line
-                          cairo::CairoSave cs(cairo);
-                          cairo_set_source_rgba(cairo,0,0,0,0.5);
-                          cairo_move_to(cairo,x0,y0-0.5*rowHeight);
-                          cairo_line_to(cairo,w+colWidth,y0-0.5*rowHeight);
-                          cairo_stroke(cairo);
-                        }                                  
+                        }
+                        
+                        if (it==itemVector[0])
```
highperformancecoder, line 299: ditto

model/itemTab.cc
```diff
@@ -282,23 +287,27 @@ namespace minsky
...
+                            cairo_set_source_rgba(cairo,0,0,0,0.5);
+                            cairo_move_to(cairo,x0,y0-0.5*rowHeight);
+                            cairo_line_to(cairo,w+colWidth,y0-0.5*rowHeight);
+                            cairo_stroke(cairo);
+                          }                                  
                         cairo::CairoSave cs(cairo);
                         // make sure rectangle has right height
-                        cairo_rectangle(cairo,x0,y0-1.5*rowHeight,w+colWidth,y-y0+2*rowHeight);    
+                        if (it==itemVector[0]) cairo_rectangle(cairo,x0,y0-1.5*rowHeight,w+colWidth,y-y0+2*rowHeight);    
+                        else cairo_rectangle(cairo,x0,y0-rowHeight,w+colWidth,y-y0+rowHeight);
```
highperformancecoder, line 310: Could we not reuse the previous y1 calculation here?

model/itemTab.cc
wdednam: It's incremented at the end of loop (for auto& it: itemVector).

model/itemTab.cc
wdednam: Done, thanks :)

wdednam:
Obviously, further fixes will be required when itemVector contains rank 1 or higher tensors...

model/itemTab.cc
wdednam: Done :)

model/itemTab.cc
wdednam: > Wouldn't this be a right margin?

I think it is the left margin of a column. Now that I think about it, I may be able to use to adjust the column widths.

PR #268, opened 2021-01-05
removed redefinition of t0 in minsky.h and start time now also persis…

wdednam:
…ted correctly

---
This change is [Reviewable](https://reviewable.io/reviews/highperformancecoder/minsky/268)

### DIFF
--- a/model/minsky.h
+++ b/model/minsky.h
@@ -284,7 +284,6 @@ namespace minsky
     void jacobian(Matrix& jac, double t, const double vars[]);
     
     double t{0}; ///< time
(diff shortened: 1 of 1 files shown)


model/minsky.h
```diff
@@ -284,7 +284,6 @@ namespace minsky
     void jacobian(Matrix& jac, double t, const double vars[]);
     
     double t{0}; ///< time
-    double t0{0}; ///< simulation start time
```
highperformancecoder, line 287: Whoops! Good catch!

PR #272, opened 2021-01-08
Bugfix 1243

wdednam:
Canvas::unGroupItem() does a bit more heavy-lifting now by using auto-layout on the newly ungrouped items. This necessarily requires leaving them in selection to avoid them ending up under/over other items of the canvas. Some of the code from Minsky::paste() has also been repurposed to prevent parameter values from being reset to zero.

---
This change is [Reviewable](https://reviewable.io/reviews/highperformancecoder/minsky/272)

### DIFF
--- a/model/canvas.cc
+++ b/model/canvas.cc
@@ -664,53 +664,55 @@ namespace minsky
     if (auto g=dynamic_cast<Group*>(item.get()))
       {		  
         if (auto p=g->group.lock())
-		  if (!g->empty()) // minskly crashes if group empty and ungrouped subsequently. for ticket 1243
-		    {  	    				 
-               // stash values of parameters in copied group, as they are reset for some unknown reason later on. for tickets 1243/1258
-               map<string,string> existingParms; 
-               for (auto& i: g->items) {
-                   auto v=i->variableCast(); 
-                   if (v && v->type()==VariableType::parameter) 
-	           		existingParms.emplace(v->valueId(),v->init());
-	           }
+          {
+            if (!g->empty()) // minskly crashes if group empty and ungrouped subsequently. for ticket 1243
+              {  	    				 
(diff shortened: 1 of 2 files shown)


model/canvas.cc
```diff
@@ -662,12 +662,58 @@ namespace minsky
...
+                    if (v->type()==VariableType::parameter && !existingParms.empty()) 
+                      {
+                        auto it=existingParms.find(v->valueId());
+                        if (it!=existingParms.end()) v->init(it->second);
+                      }
+                }
+	           
+                selection.autoLayout();
+	           
+                if (!existingParms.empty()) existingParms.clear();
```
highperformancecoder, line 700: No need to test for emptiness here.

model/canvas.cc
```diff
@@ -662,12 +662,58 @@ namespace minsky
...
+	           
+                // Attach mouse focus only to first visible item in selection. For ticket 1098.      
+                for (auto& i: selection.items)
+                  if (i->visible())
+                    {
+                      setItemFocus(i);
+                      break;
+                    }
+                                   
+                if (!copyOfGroups.empty()) setItemFocus(copyOfGroups[0]);
```
highperformancecoder, line 710: This assumes that all groups will be visible. I think that is the case currently, but who knows.

model/canvas.cc
```diff
@@ -662,12 +662,58 @@ namespace minsky
...
+                for (auto& i: selection.items)
+                  if (i->visible())
+                    {
+                      setItemFocus(i);
+                      break;
+                    }
+                                   
+                if (!copyOfGroups.empty()) setItemFocus(copyOfGroups[0]);           
+              } else {
+              p->moveContents(*g);
```
highperformancecoder, line 712: IIUC, g is empty at this point, so moveContents does nothing.

PR #273, opened 2021-01-09
implemented suggestions on previous PR

wdednam:
This change is [Reviewable](https://reviewable.io/reviews/highperformancecoder/minsky/273)

### DIFF
--- a/model/canvas.cc
+++ b/model/canvas.cc
@@ -696,6 +696,8 @@ namespace minsky
                 }
 	           
                 selection.autoLayout();
+		    
+		existingParms.clear();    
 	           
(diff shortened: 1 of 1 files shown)


model/canvas.cc
```diff
@@ -697,8 +697,6 @@ namespace minsky
 	           
                 selection.autoLayout();
 	           
-                if (!existingParms.empty()) existingParms.clear();
```
highperformancecoder, line 700: Didn't you still want to delete any entries that might be in this map?

wdednam: Ah, I see: Don't test for emptiness but still clear exisitngParms. Sorry, I misunderstood you.

PR #274, opened 2021-01-12
very much a WIP, but would be much more efficient if I could get it t…

wdednam:
…o work for arbitrary rotations. The resize handles appear where they should on the rotated item, but resizing becomes impossible because the bounding box is not updating properly

---
This change is [Reviewable](https://reviewable.io/reviews/highperformancecoder/minsky/274)

### DIFF
--- a/model/godleyIcon.cc
+++ b/model/godleyIcon.cc
@@ -357,8 +357,8 @@ namespace minsky
       {
         // top justification at bottom of icon if displayed, bottom justified otherwise
         v->rotation(90);
-        v->moveTo(x, y + v->y() - (variableDisplay? v->top(): v->bottom()));
-        x+=v->width();
+        v->moveTo(x,y + v->x() - (variableDisplay? v->left(): v->right()));  // need to use x after stock vars have been rotated. for tickets 1230/1230
+        x+=v->height();
       }
   }
 
--- a/model/item.cc
+++ b/model/item.cc
@@ -45,8 +45,8 @@ namespace minsky
     x.onResizeHandles=false;
     try
       {
-		cairo_rotate(surf.cairo(),x.rotation()*M_PI/180);  // perform transformation after drawing, otherwise ink extents not calculated correctly below. For ticket 1232  
 		x.draw(surf.cairo());                    
+		cairo_rotate(surf.cairo(),-x.rotation()*M_PI/180);  // perform transformation after drawing, otherwise ink extents not calculated correctly below. For ticket 1232    
       }
     catch (const std::exception& e) 
       {cerr<<"illegal exception caught in draw(): "<<e.what()<<endl;}
@@ -144,7 +144,7 @@ namespace minsky
   { 
      double angle=rotatedPoints().first;		  
      Point p=rotatedPoints().second;		    
-     Rotate r(0,0,0); // rotate into variable's frame of reference
+     Rotate r(0,0,0);
      return near(x,y,p.x(),p.y(),resizeHandleSize(),r);
   }
  
--- a/model/item.cc
+++ b/model/item.cc
@@ -31,7 +31,6 @@
 #include <exception>
 
 using ecolab::Pango;
-using namespace std;
 
 namespace minsky
 {
@@ -45,8 +44,8 @@ namespace minsky
     x.onResizeHandles=false;
     try
       {
+		cairo_rotate(surf.cairo(),-x.rotation()*M_PI/180); 
 		x.draw(surf.cairo());                    
-		cairo_rotate(surf.cairo(),-x.rotation()*M_PI/180);  // perform transformation after drawing, otherwise ink extents not calculated correctly below. For ticket 1232    
       }
     catch (const std::exception& e) 
       {cerr<<"illegal exception caught in draw(): "<<e.what()<<endl;}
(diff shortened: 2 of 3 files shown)


model/item.cc
```diff
@@ -39,14 +39,14 @@ namespace minsky
...
-       (cairo_recording_surface_create(CAIRO_CONTENT_COLOR_ALPHA,NULL));
+       (cairo_recording_surface_create(CAIRO_CONTENT_COLOR_ALPHA,NULL));    
     auto savedMouseFocus=x.mouseFocus;
     x.mouseFocus=false; // do not mark up icon with tooltips etc, which might invalidate this calc
     x.onResizeHandles=false;
     try
       {
-        x.draw(surf.cairo());
-		cairo_rotate(surf.cairo(),-x.rotation()*M_PI/180);  // perform transformation after drawing, otherwise ink extents not calculated correctly below. For ticket 1232      
+		cairo_rotate(surf.cairo(),x.rotation()*M_PI/180);  // perform transformation after drawing, otherwise ink extents not calculated correctly below. For ticket 1232
```
highperformancecoder, line 48: Firstly, the comment is no longer correct. Secondly, is ticket #1232 still relevant?

wdednam: > Firstly, the comment is no longer correct

I know, that is why it is still only a draft PR.

For the  life of me, I can't understand why the resize arrows can appear correctly on vars and ops (not intop though, but that is why the override exists, so I will have to specialize in its case), work two or three times at any angle, and then stop working.

Whether I use -x.rotation() or +x.rotation() does not seem to make a difference...

> Secondly, is ticket #1232 still relevant?

I can probably fix that using a new Rotate member for the stock variables in GodleyIcon::positionVariables(). 

I might have to move on though, because I already got bogged down on ticket 1264, where I could not even get the needle to move. At least on this one I've been making very slow incremental progress.

wdednam: > For the life of me, I can't understand why the resize arrows can appear correctly on vars and ops (not intop though, but that is why the override exists, so I will have to specialize in its case), work two or three times at any angle, and then stop working.

The bounding box is wrong, so if a var name is long (more than one letter), in any angle other than 0 or 180 degrees, the ports and resize handle can no longer be selected. This in infuriating, because the ports and resize handle appear when the mouse cursor moves over them. I don't understand why the bounding box does not update to the correct shape...

> Secondly, is ticket #1232 still relevant?

I have managed to fix this another way, but the ports of the stocks on Godleys can't be selected if its name is longer than a single letter.

IntOp still a WIP too. Very tricky this one, but I feel like I'm slowly, painfully so, moving towards a better solution than I had previously.

WorksForVarsWithShortNames

model/godleyIcon.cc
```diff
@@ -357,8 +357,8 @@ namespace minsky
       {
         // top justification at bottom of icon if displayed, bottom justified otherwise
         v->rotation(90);
-        v->moveTo(x, y + v->y() - (variableDisplay? v->top(): v->bottom()));
-        x+=v->width();
+        v->moveTo(x,y + v->x() - (variableDisplay? v->left(): v->right()));  // need to use x after stock vars have been rotated. for tickets 1230/1230
```
highperformancecoder, line 360: That doesn't make much sense. Rotation only affects the coordinates of the icon components relative to the item's origin, not the meaning of Item::x and Item::y.

model/godleyIcon.cc
```diff
@@ -357,8 +357,8 @@ namespace minsky
...
+        x+=v->height();
```
highperformancecoder, line 361: Currently width/height also refer to variables after rotation, so this also doesn't make sense:
    double Item::rotation(const double& r) {
      m_rotation=r;
      bb.update(*this);
      return m_rotation;
    }

model/item.cc
```diff
@@ -31,22 +31,21 @@
 #include <exception>
 
 using ecolab::Pango;
-using namespace std;
```
highperformancecoder, line 34: Any particular reason for not importing std namespace here?

model/godleyIcon.cc
wdednam: Maybe the problem is the bounding box update in the iWidth and iHeight setters? Maybe they shouldn't be there, and only in resize and rotate? I'll try that.

The problem here is that the stocks are positioned relative to the Godly icon and not the canvas background, which is why the stocks were spaced irregularly in ticket 1232.

model/item.cc
wdednam: Oops, I was working on both operation.cc and item.cc at the same time and added it to operation.cc but then deleted it from item.cc by accident. I will put it back in.

model/godleyIcon.cc
wdednam: I know, but I think it is because the variables are positioned relative to the Godley icon and not the canvas, so using the y coordinates here led to the irregular spacing in ticket 1232.

model/item.cc
wdednam: I have tried several different approaches, but I cannot reconcile BottomRightResizerItem::onResizeHandles() with BottomRightResizerItem::drawResizeHandles() for a general rotation and I don't understand why. If I rotate cairo in BoundingBox::update before drawing, then the resize arrows appear where they are supposed to but cannot be selected for variables that have names longer than one letter. If I rotate after drawing in BoundingBox::update, I cannot get the resize arrows to appear where they are supposed to. So, I'm closing this PR which might have to go on the backlog if I can't fix it before this iteration finalises. :(

PR #277, opened 2021-01-18
A-L-E now displayed above last column and symbolic sum or stock var r…

wdednam:
…ow displayed in top cell of last column

---
This change is [Reviewable](https://reviewable.io/reviews/highperformancecoder/minsky/277)

### DIFF
--- a/model/godleyTable.cc
+++ b/model/godleyTable.cc
@@ -199,79 +199,74 @@ string GodleyTable::rowSum(int row) const
 {
   // accumulate the total for each variable
   map<string,double> sum;
-
-  for (size_t c=1; c<cols(); ++c)
-    {
-      FlowCoef fc(cell(row,c));
-      if (row>0)
+  
+  if (row>0)
+  {
+    for (size_t c=1; c<cols(); ++c)
       {
-      if (!fc.name.empty()||initialConditionRow(row))
-        {
-          // apply accounting relation to the initial condition row
-          if (signConventionReversed(c))
-            sum[fc.name]-=fc.coef;
-          else
-            sum[fc.name]+=fc.coef;
-        }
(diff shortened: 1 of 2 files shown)


model/godleyTableWindow.cc
```diff
@@ -309,15 +309,14 @@ namespace minsky
...
     pango.setMarkup("A-L-E");
     pango.show();
     double colWidth=pango.width();
-    y+=rowHeight;
   
-    for (unsigned row=1; row<godleyIcon->table.rows(); ++row)
+    for (unsigned row=0; row<godleyIcon->table.rows(); ++row)    // perform row sum on stock var heading column too. for ticket 1285
       {
-        if (row>0 && row<scrollRowStart) continue;
+        if (row>-1 && row<scrollRowStart) continue;
```
highperformancecoder, line 319: Two comments, row will always be greater than -1, but more importantly this line is about ensuring row 0 is rendered when scrollRowStart>0.

highperformancecoder:
Need to fix Travis.

model/godleyTable.cc
```diff
@@ -211,20 +213,64 @@ string GodleyTable::rowSum(int row) const
           else
             sum[fc.name]+=fc.coef;
         }
+      } else   // display numerical sum of stock vars in first cell of last column. for ticket 1285
```
highperformancecoder, line 216: It might be simpler just to have a separate loop summing up the values of the stock variables for row 0, rather than this rather tortuous dealing with row 0 in a separate branch in the loops dealing with the other rows.

wdednam: Done, but values of stock vars did not display well. Numerical inaccuracies grew beyond 1.1e-15 after integrating a model for sufficient time. This caused values such as 1e-14 and greater to be displayed all the time in the first cell of the A-L-E column.

PR #278, opened 2021-01-19
gather now xvector aware like other tensor binops and units are also …

wdednam:
…transferred through like add, subtract, max and min binops

---
This change is [Reviewable](https://reviewable.io/reviews/highperformancecoder/minsky/278)

### DIFF
--- a/engine/minskyTensorOps.cc
+++ b/engine/minskyTensorOps.cc
@@ -499,7 +499,29 @@ namespace minsky
     }
     Timestamp timestamp() const override {return max(arg1->timestamp(), arg2->timestamp());}
     void setArguments(const TensorPtr& a1, const TensorPtr& a2) override {
-      arg1=a1; arg2=a2;
+	
+	// ensures that gather operation is x-vector aware, just like other tensor binops for ticket 1271	
+      if (a1 && a1->rank()>0)
+        hypercube(a1->hypercube());
+      else if (a2)
+        hypercube(a2->hypercube());
+      else
+        hypercube(Hypercube());
+      
+      arg1=a1;
+
+      if (a2)
+        {
+          if (a2->rank()==0 || a2->hypercube()==hypercube())
+            arg2=a2;
+          else
+            {
+              arg2=make_shared<InterpolateHC>();
+              arg2->hypercube(hypercube());
+              arg2->setArgument(a2);
+            }
+        }    
+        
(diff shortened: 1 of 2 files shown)


engine/minskyTensorOps.cc
```diff
@@ -499,7 +499,29 @@ namespace minsky
     }
     Timestamp timestamp() const override {return max(arg1->timestamp(), arg2->timestamp());}
     void setArguments(const TensorPtr& a1, const TensorPtr& a2) override {
-      arg1=a1; arg2=a2;
+	
+	// ensures that gather operation is x-vector aware, just like other tensor binops for ticket 1271
```
highperformancecoder, line 503: This won't work, because gather is not like other binops. The other binops are a generalisation of elementwise binary ops, whereas as gather is a generalisation of an index. The type (or units) of the second argument must match the type of first argument's x-vector. The InterpolateHC will come into play, but probably attached to the first argument.

wdednam: I understand, thank you. I have a few questions if you don't mind: will the implementation require modifications outside of minskyTensorOp.cc and operation.cc? Would it not be more appropriate to change the "Gather" op's base class to DimensionedArgCachedOp to achieve the desired result?

highperformancecoder:
On Thu, Jan 21, 2021 at 10:57:25AM -0800, Wynand wrote:
> I understand, thank you. I have a few questions if you don't mind: will the
> implementation require modifications outside of minskyTensorOp.cc and
> operation.cc? Would it not be more appropriate to change the "Gather" op's base
> class to DimensionedArgCachedOp to achieve the desired result?

I don't think so. CachedOp is used where you need to calculate the
full tensor just to work out one component. With gather, each
component of the result is independent of all the others.

You can probably precalculate the lookup tables, though.

Leave this for now if it is looking too complicated.


-- 

                      http://www.hpcoders.com.au

engine/minskyTensorOps.cc
highperformancecoder: I thought I'd responded to this. No -CachedOp is not approprate, as this each element's calculation will be independent of all of the others.

wdednam: Yes, you did, below. I haven't made any progress here and will probably have to go back to the drawing board on this one.. You can close it if you like.

PR #279, opened 2021-01-25
trying to get the columns in which pango.width()+5 exceeds colWidth j…

wdednam:
…ust once, to all adjust to the widest value globalColMaxWidth, even if in some rows that condition is not met

---
This change is [Reviewable](https://reviewable.io/reviews/highperformancecoder/minsky/279)

### DIFF
--- a/model/itemTab.cc
+++ b/model/itemTab.cc
@@ -274,7 +274,7 @@ namespace minsky
                           cairo_stroke(cairo);                                                        
                           for (auto& i : colWidths)
                             {						
-							  x+=i;	
+                              x+=i;	
                               cairo_move_to(cairo,x,y-2*rowHeight);
                               cairo_line_to(cairo,x,y+y1);
                               cairo_stroke(cairo);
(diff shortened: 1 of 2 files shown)


wdednam:
Arggg. This is not working 100% yet. Sometimes the last horizontal line of the rank 0 objects is not drawn... Working to fix it now... Sorry...

model/itemTab.cc
```diff
@@ -209,37 +209,43 @@ namespace minsky
...
-                        varAttribVals.push_back(to_string(v->sliderStep));
-                        varAttribVals.push_back(to_string(v->sliderMin));
-                        varAttribVals.push_back(to_string(v->sliderMax));
-                        varAttribVals.push_back(to_string(v->value()));
+                        varAttribVals.push_back(str(v->sliderStep));
+                        varAttribVals.push_back(str(v->sliderMin));
+                        varAttribVals.push_back(str(v->sliderMax));
+                        varAttribVals.push_back(str(v->value()));
+                        
+                        size_t vACtr=0;
```
highperformancecoder, line 224: What does vACtr stand for? Something counter?

model/itemTab.cc
```diff
@@ -254,7 +260,7 @@ namespace minsky
                           cairo_set_source_rgba(cairo,0,0,0,0.2);
                           for (y=y0+y1; y<h+rowHeight; y+=2*rowHeight)
                             {
-                              cairo_rectangle(cairo,x0,y,w+colWidth,rowHeight);
+                              cairo_rectangle(cairo,x0,y,w+colWidths.back(),rowHeight);
```
highperformancecoder, line 263: I'm little puzzled why the last column width would be appropriate for all columns here...

model/itemTab.cc
wdednam: Yes, it is the variable attribute counter. Not a very good name I'm afraid. It could have just been Ctr, I guess.

model/itemTab.cc
wdednam: I've generalised the colWidth=std::max(colWidth,pango.width()+5) approach you used so that it is done for every column individually. Now, the width of a given column on the tab will be the maximum pango.width()+5 in that column and not the global maximum in all columns.

PR #280, opened 2021-01-28
lower case gamma reinstated as par or var. Gamma used for the operation

wdednam:
This change is [Reviewable](https://reviewable.io/reviews/highperformancecoder/minsky/280)


highperformancecoder:
Yes - it is reasonable to change this to Gamma, as it is a capital symbol, even though it contradicts the usual naming style, because the name of the enum tag appears on the user interface.

PR #281, opened 2021-01-29
Bugfix 1291

wdednam:
This is supposed to pick up names such as \Gamma, \sin, etc. and then check the substring to convert it into an op or flow var as the case may be. I checked the regex at regex101.com and it works as it is supposed to, but I don't understand why it is not working in tcl/tk. At regex101: https://regex101.com/r/FZgMbO/1

---
This change is [Reviewable](https://reviewable.io/reviews/highperformancecoder/minsky/281)

### DIFF
--- a/gui-tk/wiring.tcl
+++ b/gui-tk/wiring.tcl
@@ -407,16 +407,13 @@ proc textOK {} {
     destroy .textInput
     canvas.moveOffsX 0
     canvas.moveOffsY 0
-	if [regexp "^\\\[a-zA-Z\]+$" $textBuffer] {  # ensure that names escaped with \ can be used as ops or pars and vars as the case may be. for ticket 1291
+	if {[lsearch [availableOperations] $textBuffer]>-1} {
 		  set subTextBuffer [string range $textBuffer 1 end]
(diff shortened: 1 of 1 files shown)


gui-tk/wiring.tcl
```diff
@@ -407,8 +407,13 @@ proc textOK {} {
     destroy .textInput
     canvas.moveOffsX 0
     canvas.moveOffsY 0
-    if {[lsearch [availableOperations] $textBuffer]>-1} {
-		addOperationKey $textBuffer
+	if {[lsearch [availableOperations] $textBuffer]>-1} {
+		  set subTextBuffer [string range $textBuffer 1 end]
+		  if {[lsearch [availableOperations] $subTextBuffer]>-1} { # ensure that names escaped with \ can be used as ops or pars and vars as the case may be. for ticket 1291
```
highperformancecoder, line 412: I like your approach, but I don't think it will work for backslashes appearing in the middle of textBuffer. Bit of a contrived example, but suppose someone wanted to name their parameter m\in (m∈). They might be surprised to get the min operator.

Another approach might be to do search and replace on \, replacing it by \\.

PR #282, opened 2021-02-01
I know it has to be index from first tensor plus some stride times in…

wdednam:
…dex from second tensor, but which stride???

---
This change is [Reviewable](https://reviewable.io/reviews/highperformancecoder/minsky/282)


engine/minskyTensorOps.cc
```diff
@@ -400,6 +400,18 @@ namespace minsky
...
         hc.xvectors.insert(hc.xvectors.begin(), xv1.begin(), xv1.end()-1);
         cachedResult.hypercube(move(hc));
+        
+        auto idx1=arg1->index(), idx2=arg2->index();
+        set<size_t> newIdx;
+        
+        for (auto& i: idx1)
+           for (auto& j: idx2)
+             {
+               newIdx.insert(i+arg1->hypercube().splitIndex(i).size()*j);    // I know it has to be index from first tensor plus some stride times index from second tensor, but what stride???
```
highperformancecoder, line 410: splitIndex is an expensive operation. What your calculating here is arg1->rank(), which is much cheaper.

highperformancecoder: This line seems more appropriate of outproduct than inner product

engine/minskyTensorOps.cc
```diff
@@ -433,6 +433,38 @@ namespace minsky
...
+        
+	    vector<size_t> tmpIdx;
+	    size_t tmpLinealIdx;        
+        
+        // first element of split index of first tensor and last element of split index of second tensor is the split index of product tensor. Convert to lineal index to get new index vector????    
+        for (size_t i=0; i< min(hcIdx1.size(),hcIdx2.size());i++)
+          {
+			 if (i==0) tmpIdx.push_back(hcIdx1[i]);
+			 if (i == min(hcIdx1.size(),hcIdx2.size())-1) tmpIdx.push_back(hcIdx2[i]);
+		 }
```
highperformancecoder, line 458: I don't think this is correct. Yes - the approach is valid, but the new split index should be something like [hcIdx1[0],...hcIdx1[hcIdx.size()-2,hcIdx2[1],...hcIdx2[hcIdx2.size()-1]]. You can use insert operations:

vector<size_t> tmpIdx(hcIdx1.begin(), hcIdx1.end()-1);
tmpIdx.insert(tmpIdx.end(),hcIdx2.begin()+1,hcIdx2.end());

engine/minskyTensorOps.cc
```diff
@@ -433,6 +433,38 @@ namespace minsky
...
+        tmpLinealIdx = hc.linealIndex(tmpIdx);		 	 
+        newIdx.insert(tmpLinealIdx);		 
+        tmpIdx.clear();		 
+         
+		cachedResult.index(Index(newIdx));               
+		
+		hcIdx1.clear();
+		hcIdx2.clear();
```
highperformancecoder, line 466: Unnecessary, as the destructors take care of this for you.

engine/minskyTensorOps.cc
```diff
@@ -433,6 +433,41 @@ namespace minsky
...
         cachedResult.hypercube(move(hc));
+        
+        set<size_t> newIdx;
+        vector<size_t> hcIdx1,hcIdx2;       
+        size_t tmpLinealIdx;        
+        
+        if (!arg1->index().empty() && !arg2->index().empty())
+        {
+        for (auto& i: arg1->index()) 
+            if (i!=arg1->index()[arg1->index().size()-1])
```
highperformancecoder, line 444: What are you trying to do here?

engine/minskyTensorOps.cc
```diff
@@ -433,6 +433,41 @@ namespace minsky
...
+            {
+			   for (auto& k: arg1->hypercube().splitIndex(i))
+			      hcIdx1.push_back(k);
```
highperformancecoder, line 447: This is suboptimal. Just use assignment, the compiler will probably use move assignment.

engine/minskyTensorOps.cc
```diff
@@ -433,6 +433,41 @@ namespace minsky
...
+              for (auto& l: arg2->hypercube().splitIndex(j)) 
+                    hcIdx2.push_back(l); 
+              tmpLinealIdx = hc.linealIndex(hcIdx2);		 	 
+              newIdx.insert(tmpLinealIdx);        
+              hcIdx2.clear(); 
+            }
+        
+        //// first element of split index of first tensor and last element of split index of second tensor is the split index of product tensor. Convert to lineal index to get new index vector????    
+		//vector<size_t> tmpIdx(hcIdx1.begin(), hcIdx1.end()-1);
+        //tmpIdx.insert(tmpIdx.end(),hcIdx2.begin()+1,hcIdx2.end());
```
highperformancecoder, line 467: This is what you should be doing for innerProduct.

PR #284, opened 2021-02-07
nearly good to go, I think, but there may be ways to improve upon thi…

wdednam:
…s and make it more effcient

---
This change is [Reviewable](https://reviewable.io/reviews/highperformancecoder/minsky/284)


wdednam:
Making another small improvement (I hope), so please don't review yet...

PR #285, opened 2021-02-17
Creating child window for rendering

digiperfect-janak:
We have created a struct `WindowInformation` that stores the `childWindowId` along with other details like display and window attributes. This information is reused across multiple calls to `renderFrame`. 

The flow for code will be -- when minsky starts:
A call to` /minsky/canvas/initializeNativeWindow ` will be made, with parentWindowId (and offsets) as the parameters (creating child window in electron did not work as expected, so we need to work with offsets). 
Subsequent repaints can be requested with ` /minsky/canvas/renderFrame`

As of now, we create a cairo surface with each call to `renderFrame`, though I think the surface can also be reused. I tried having pointer to cairo::SurfacePtr (not sure we should have pointer to pointer) but it didn't work as expected (maybe syntax related usage issue), so for now I reverted to recreating the surface in the call to `renderFrame`

Please especially review the lifecycle (constructors, desctructors and copy constructors) that I have defined in `renderNativeWindow.cc `. I think the `WindowInformation` object that is destroyed in the destructor for `RenderNativeWindow` can also be reused (perhaps it can be made a static object?). 

Also - am not sure how to distinguish between destructor for `RenderNativeWindow` that will be called with each call to load model (or undo/redo as you mentioned), and the final call when minsky is closed.

---
This change is [Reviewable](https://reviewable.io/reviews/highperformancecoder/minsky/285)

### DIFF
--- a/model/renderNativeWindow.cc
+++ b/model/renderNativeWindow.cc
@@ -77,12 +77,38 @@ namespace minsky
     int err=XGetWindowAttributes(display, window, &wAttr);
     if (err>1)
       throw runtime_error("Invalid window: "+to_string(window));
-    cairo::SurfacePtr r(new cairo::Surface(cairo_xlib_surface_create(display,window,wAttr.visual,wAttr.width,wAttr.height),wAttr.width,wAttr.height));
+
+
+    int padding = 10;
+    int yOffset = 150;
+    int childWindowWidth = wAttr.width - 2*padding;
+    int childWindowHeight = wAttr.height - yOffset - padding;
+
+    unsigned long childWindow = XCreateSimpleWindow(display, window, padding, yOffset, childWindowWidth, childWindowHeight, 0, 0, 0);
+    XMapWindow(display, childWindow);
+
+    cout << "Child Window ID ::"<< childWindow << endl;
+    //XSelectInput(d, da, ButtonPressMask | KeyPressMask);
+    cout << "wAttr(x, y)::" << wAttr.x << "," << wAttr.y << endl;
+
+    cairo::SurfacePtr r(new cairo::Surface(cairo_xlib_surface_create(display, childWindow,wAttr.visual, childWindowWidth, childWindowHeight), childWindowWidth, childWindowHeight));
     cairo_surface_set_device_offset(r->surface(), -wAttr.x, -wAttr.y);
+
+    cairo_move_to(r->cairo(), 0, 0);
+    cairo_set_source_rgb(r->cairo(), 1, 1, 1);
+    cairo_paint(r->cairo());
     return r;
   }
 #endif
-  
+
+
+  void RenderNativeWindow::initializeNativeWindow(unsigned long window) {
+
+  }
+
+  void RenderNativeWindow::renderFrame() {
+
+  }
 
   void RenderNativeWindow::renderToNativeWindow(unsigned long window)
   {
--- a/model/renderNativeWindow.cc
+++ b/model/renderNativeWindow.cc
@@ -30,29 +30,6 @@ Please especially review the lifecycle (constructors, desctructors and copy cons
 #include "windowInformation.h"
 #include "minsky_epilogue.h"
 
-#if defined(CAIRO_HAS_XLIB_SURFACE) && !defined(MAC_OSX_TK)
-#include <cairo/cairo-xlib.h>
-#include <X11/Xlib.h>
-#endif
-
-#if defined(CAIRO_HAS_WIN32_SURFACE) && !defined(__CYGWIN__)
-#define USE_WIN32_SURFACE
-#endif
-
-#ifdef _WIN32
-#undef Realloc
-#include <windows.h>
-#include <wingdi.h>
-#ifdef USE_WIN32_SURFACE
-#include <cairo/cairo-win32.h>
-#endif
-#endif
-
-#if defined(MAC_OSX_TK)
-#include <Carbon/Carbon.h>
-#include <cairo/cairo-quartz.h>
-#include "getContext.h"
-#endif
 
 #include <stdexcept>
 #include <string>
@@ -62,29 +39,14 @@ using namespace ecolab;
 
 namespace minsky
 {
-#ifdef USE_WIN32_SURFACE
-  inline cairo::SurfacePtr createNativeWindowSurface(WindowInformation &winInfo)
-  { /* TODO */
-  }
-#elif defined(MAC_OSX_TK)
-  inline cairo::SurfacePtr createNativeWindowSurface(WindowInformation &winInfo)
-  { /* TODO */
-  }
-#else
-  inline cairo::SurfacePtr createNativeWindowSurface(WindowInformation &wi)
-  {
-    cairo::SurfacePtr childSurface(new cairo::Surface(cairo_xlib_surface_create(wi.getDisplay(), wi.getChildWindowId(), wi.wAttr.visual, wi.childWidth, wi.childHeight), wi.childWidth, wi.childHeight));
-    cairo_surface_set_device_offset(childSurface->surface(), -wi.wAttr.x, -wi.wAttr.y);
-    return childSurface;
-  }
-#endif
-
   void RenderNativeWindow::renderFrame(unsigned long parentWindowId, int offsetLeft, int offsetTop, int childWidth, int childHeight)
   {
     if (!winInfoPtr)
      winInfoPtr=std::make_shared<WindowInformation>(parentWindowId, offsetLeft, offsetTop, childWidth, childHeight);
 
-    auto tmp = createNativeWindowSurface(*winInfoPtr);
+    auto tmp = winInfoPtr->getSurface();
+    
+    //auto tmp = createNativeWindowSurface(*winInfoPtr);
 
     //TODO:: Review if this paint (below 3 lines) is really needed with each frame
     cairo_move_to(tmp->cairo(), 0, 0);
--- a/model/renderNativeWindow.cc
+++ b/model/renderNativeWindow.cc
@@ -181,4 +181,9 @@ namespace minsky
     offsetTop = max(180, offsetTop);
     this->winInfo->initialize(parentWindowId, offsetLeft, offsetTop, childWidth, childHeight);
   }
+
+
+  void RenderNativeWindow::resizeWindow(int offsetLeft, int offsetTop, int childWidth, int childHeight) {
+    // TODO:: To be implemented... need to recreate child window
+  }
 } // namespace minsky
--- a/model/renderNativeWindow.cc
+++ b/model/renderNativeWindow.cc
@@ -54,67 +54,110 @@ using namespace ecolab;
 namespace minsky
 {
 #ifdef USE_WIN32_SURFACE
-  inline cairo::SurfacePtr nativeWindowSurface(unsigned long window)
-  {/* TODO */}
+  inline cairo::SurfacePtr createNativeWindowSurface(WindowInformation *winInfo)
+  { /* TODO */
+  }
 #elif defined(MAC_OSX_TK)
-  inline cairo::SurfacePtr nativeWindowSurface(unsigned long window)
-  {/* TODO */}
+  inline cairo::SurfacePtr createNativeWindowSurface(WindowInformation *winInfo)
+  { /* TODO */
+  }
 #else
 
-  int throwOnXError(Display*, XErrorEvent* ev)
+  int throwOnXError(Display *, XErrorEvent *ev)
   {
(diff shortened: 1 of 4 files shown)


model/renderNativeWindow.cc
```diff
@@ -54,41 +63,127 @@ using namespace ecolab;
...
+  int throwOnXError(Display *, XErrorEvent *ev)
   {
     char errorMessage[256];
     XGetErrorText(ev->display, ev->error_code, errorMessage, sizeof(errorMessage));
     throw runtime_error(errorMessage);
   }
-  
-  inline cairo::SurfacePtr nativeWindowSurface(unsigned long window)
+
+  inline cairo::SurfacePtr createNativeWindowSurface(WindowInformation *wi)
```
highperformancecoder, line 82: There's no good reason to pass this as a pointer. Better to use a reference.

model/renderNativeWindow.h
```diff
@@ -24,12 +24,44 @@
...
+    ~WindowInformation();
+    void initialize(unsigned long parentWin, int offsetLeft, int offsetTop, int childWidth, int childHeight);
+    void copy(WindowInformation *winInfo);
+  };
+
+  class RenderNativeWindow : public ecolab::CairoSurface
   {
   public:
-    void renderToNativeWindow(unsigned long window);
+    /* Perhaps needs to be public for some auto-binding code */
```
highperformancecoder, line 48: Definitely not. If you're managing raw pointers in this way (new in constructor/delete in destructor), then the actual pointer *must* be private, and the copy operations *must* be deleted. But most classes in Minsky need to be DCAS (default constructible, copy constructible, assignable and serialisable), so you won't get very far with this strategy.

If you do need pointers (I'm not sure you really need them here, mind you), then using smart pointers is the smart thing to do, as all these issues have been thought out for you. For example, std::shared_ptr is DCAS, and std::unique_ptr is nearly so.

model/renderNativeWindow.cc
```diff
@@ -54,41 +63,127 @@ using namespace ecolab;
...
-  void RenderNativeWindow::renderToNativeWindow(unsigned long window)
+  WindowInformation::WindowInformation()
+  {
+  }
+  WindowInformation::~WindowInformation()
+  {
+    cout << "Delete called for window information" << endl;
+  }
+
+  void WindowInformation::initialize(unsigned long parentWin, int left, int top, int cWidth, int cHeight)
```
highperformancecoder, line 98: This should really be implemented in the constructor (RAII and all that).

model/renderNativeWindow.cc
```diff
@@ -54,41 +63,127 @@ using namespace ecolab;
...
+
+    cout << childWidth << "::" << childHeight << "::" << offsetLeft << "::" << offsetTop << endl;
+    childWindowId = XCreateSimpleWindow(display, parentWin, offsetLeft, offsetTop, childWidth, childHeight, 0, 0, 0); //TODO:: Should we pass visual and attributes at the end?
+    XMapWindow(display, childWindowId);
+  }
+
+  void WindowInformation::copy(WindowInformation *winInfo)
+  {
+    this->parentWindowId = winInfo->parentWindowId;
+    this->childWindowId = winInfo->childWindowId;
```
highperformancecoder, line 131: childWindowId is managing resources in just the same way as a pointer does, so this line leaks resources. The proper way to solve this is to apply RAII.

model/renderNativeWindow.h
```diff
@@ -24,12 +24,44 @@
...
+  {
+    unsigned long parentWindowId;
+    unsigned long childWindowId;
+    int childWidth;
+    int childHeight;
+    int offsetLeft;
+    int offsetTop;
+    Display*	display;
+    XWindowAttributes wAttr;
+    ecolab::cairo::SurfacePtr *childSurface; // TODO:: Review right way to store surface ptr
```
highperformancecoder, line 37: A SurfacePtr is a smart pointer - basically a std::shared_ptr IIRC. Taking a pointer to it will almost invariably lead to resource leaks and object lifetime problems.

model/renderNativeWindow.h
```diff
@@ -24,12 +24,44 @@
...
-  class RenderNativeWindow: public ecolab::CairoSurface
+  struct WindowInformation
+  {
+    unsigned long parentWindowId;
+    unsigned long childWindowId;
+    int childWidth;
+    int childHeight;
+    int offsetLeft;
+    int offsetTop;
+    Display*	display;
```
highperformancecoder, line 35: Because of the problems pointers create, it is worthwhile noting that this is a weak pointer, ie ownership of the display object remains with the X11 system.

model/renderNativeWindow.cc
```diff
@@ -54,41 +63,127 @@ using namespace ecolab;
...
-    auto tmp=nativeWindowSurface(window);
+    cout << "Copy constructor for RenderNativeWindow called... TODO:: implement this" << endl;
+    //a.winInfo->copy(this->winInfo); // reverse?
+  };
+
+  void RenderNativeWindow::renderFrame()
+  {
+    auto tmp = createNativeWindowSurface(winInfo);
+
+    //TODO:: Review if this paint (below 3 lines) is really needed with each frame
```
highperformancecoder, line 166: My intuition is that painting the background canvas is not necessary for X11, but might be for Windows. I guess we should suck it and see.

highperformancecoder:
> We have created a struct `WindowInformation` that stores the `childWindowId` along with other details like display and window attributes. This information is reused across multiple calls to `renderFrame`.
> 
> The flow for code will be -- when minsky starts:
> A call to`/minsky/canvas/initializeNativeWindow` will be made, with parentWindowId (and offsets) as the parameters (creating child window in electron did not work as expected, so we need to work with offsets).

Why not check whether windowInfo is initialised in renderNativeWindow, and initialise if not at that time? Lazy initialisation.

> Subsequent repaints can be requested with ` /minsky/canvas/renderFrame`
> 
> As of now, we create a cairo surface with each call to `renderFrame`, though I think the surface can also be reused. I tried having pointer to cairo::SurfacePtr (not sure we should have pointer to pointer) but it didn't work as expected (maybe syntax related usage issue), so for now I reverted to recreating the surface in the call to `renderFrame`
> 

Yes - the SurfacePtr needs to have a shorter lifetime (only just shorter) than the child Window. So it should probably be declared after the WindowInfo to ensure it is destroyed before the child window.

> Please especially review the lifecycle (constructors, desctructors and copy constructors) that I have defined in `renderNativeWindow.cc `. I think the `WindowInformation` object that is destroyed in the destructor for `RenderNativeWindow` can also be reused (perhaps it can be made a static object?).

Yes - lots of problems, as I identified in the comments. Please review RAII principles.

> 
> Also - am not sure how to distinguish between destructor for `RenderNativeWindow` that will be called with each call to load model (or undo/redo as you mentioned), and the final call when minsky is closed.

There is no need to distinguish these (it would be a mistake to do so).

model/renderNativeWindow.cc
```diff
@@ -54,41 +63,69 @@ using namespace ecolab;
...
-  int throwOnXError(Display*, XErrorEvent* ev)
+  RenderNativeWindow::~RenderNativeWindow()
   {
-    char errorMessage[256];
-    XGetErrorText(ev->display, ev->error_code, errorMessage, sizeof(errorMessage));
-    throw runtime_error(errorMessage);
+    if (this->winInfo)
+    {
+      delete this->winInfo;
+    }
```
highperformancecoder, line 87: This still won't do. RenderNativeWindow needs to be DCAS, for Classdesc reflection reasons.

model/renderNativeWindow.cc
```diff
@@ -54,41 +63,69 @@ using namespace ecolab;
...
-#endif
-  
 
-  void RenderNativeWindow::renderToNativeWindow(unsigned long window)
+  RenderNativeWindow &RenderNativeWindow::operator=(const RenderNativeWindow &a)
   {
-    auto tmp=nativeWindowSurface(window);
+    //TODO:: I expected this to be the "new" instance with uninitialized winInfo and a to have the previous winInfo.
+    // However, it seems to be the other way round in practice --- Janak
+    a.winInfo->initialize(*(this->winInfo));
```
highperformancecoder, line 99: In C++, just because a pointer is const, doesn't mean the target of the pointer is const. It's a little counterintuitive, but you have to add extra qualifiers to the type to ensure the target is const,

Nevertheless, it is is bad form to go changing the RHS of an assignment statement - it would be totally surprising to most C++ programmers.

model/renderNativeWindow.h
```diff
@@ -23,13 +23,28 @@
...
-{
-  class RenderNativeWindow: public ecolab::CairoSurface
+{  
+  class WindowInformation;
+  class RenderNativeWindow : public ecolab::CairoSurface
   {
+  private:
+    CLASSDESC_ACCESS(RenderNativeWindow); 
+    // std::shared_ptr<WindowInformation> winInfoPtr; // TODO:: cannot get things to work with this
+    WindowInformation* winInfo;
```
highperformancecoder, line 33: As mentioned, RenderNativeWindow should be DCAS, in particular have valid copy and assignment operators. Best to arrange all members of RenderNativeWindow be DCAS, and then take the compiler generated copy and assignment operators - this is more scalable.

I think we have two options for making the winInfo member DCAS - the shared_ptr approach I outlined, and the no-op copy operations: both have their own set of issues and implications. Maybe other methods will work, but I can't think of any.

As for the 'S' part, a simple Exclude wrapper is sufficient.

model/windowInformation.cc
```diff
@@ -0,0 +1,110 @@
...
+    int err = XGetWindowAttributes(display, parentWin, &wAttr);
+    if (err > 1)
+      throw runtime_error("Invalid window: " + to_string(parentWin));
+
+    childWidth = wAttr.width - offsetLeft;
+    childHeight = wAttr.height - offsetTop;
+
+    // Todo:: currently width, height parameters are not getting passed properly
+    // Eventually, we need those in order to ensure we don't need to worry about
+    // paddings, scrollbars etc
```
highperformancecoder, line 84: Why wouldn't you arrange for parentWindow to be a dedicated childWindow within electron, that way there will be no scrollbars or padding involved?

model/windowInformation.cc
```diff
@@ -0,0 +1,110 @@
...
+    {
+      childWidth = min(childWidth, cWidth);
+    }
+    if (cHeight > 0)
+    {
+      childHeight = min(childHeight, cHeight);
+    }
+
+    // std::cout << childWidth << "::" << childHeight << "::" << offsetLeft << "::" << offsetTop << std::endl;
+    childWindowId = XCreateSimpleWindow(display, parentWin, offsetLeft, offsetTop, childWidth, childHeight, 0, 0, 0); //TODO:: Should we pass visual and attributes at the end?
```
highperformancecoder, line 95: This statement leaks a child window resource. I think the simplest approach is to move the contents of this method into the constructor that calls it, but if you really insist on this two-phase initialisation protocol, then you must destroy any existing childWindow before this call.

model/windowInformation.h
```diff
@@ -0,0 +1,63 @@
...
+    
+    WindowInformation();
+    ~WindowInformation();
+    WindowInformation(unsigned long parentWin, int left, int top, int cWidth, int cHeight);
+    void initialize(unsigned long parentWin, int left, int top, int cWidth, int cHeight);
+    void initialize(const WindowInformation &other);
+
+    ecolab::cairo::SurfacePtr getSurface();
+
+    void copyFrom(const WindowInformation &other);
```
highperformancecoder, line 57: I don't see the implementation of this method here, but I'm concerned you're trying to introduce a copy operator in by the back door, and no correctly handling resource lifetime (like the initialize() example above).

model/renderNativeWindow.h
```diff
@@ -23,13 +23,19 @@
...
-  class RenderNativeWindow: public ecolab::CairoSurface
+{  
+  class WindowInformation;
+  class RenderNativeWindow : public ecolab::CairoSurface
   {
-  public:
-    void renderToNativeWindow(unsigned long window);
+  private:
+    CLASSDESC_ACCESS(RenderNativeWindow); 
+    std::shared_ptr<WindowInformation> winInfoPtr;
```
highperformancecoder, line 32: I think you will probably have to classdesc::Exclude this member.

highperformancecoder:
Visually, the code now passes muster. You also need to for the code to pass continuous integration. Looking at the errors Travis flags, I'd say this can be dealt with by wrapping winInfoPtr in a classdesc::Exclude construct.

I'll try checking out this PR branch and seeing if that does the trick.

PR #287, opened 2021-02-26
Use x11 native window clear

digiperfect-janak:
This change is [Reviewable](https://reviewable.io/reviews/highperformancecoder/minsky/287)

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -2,11 +2,14 @@
 *.d
 *.cd
 *.xcd
+*.rcd
 *~
 *.aux
 *.dvi
 *.log
 *.out
+.vscode
+doxydoc
 test/unittests
 test/tcl-cov
 test/testDatabase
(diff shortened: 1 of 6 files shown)


.gitignore
```diff
@@ -2,11 +2,14 @@
...
 *.cd
 *.xcd
+*.rcd
 *~
 *.aux
 *.dvi
 *.log
 *.out
+.vscode
+doxydoc
```
highperformancecoder, line 12: Thanks!

model/renderNativeWindow.cc
```diff
@@ -41,17 +40,22 @@ namespace minsky
 {
   void RenderNativeWindow::renderFrame(unsigned long parentWindowId, int offsetLeft, int offsetTop, int childWidth, int childHeight)
   {
-    if (!winInfoPtr.get())
-     winInfoPtr=std::make_shared<WindowInformation>(parentWindowId, offsetLeft, offsetTop, childWidth, childHeight);
+    bool isInitial = false;
```
highperformancecoder, line 43: What is this flag used for?

highperformancecoder: You should have received a note from Travis about the build failing. It turns out the compiler is asking the same question :).

digiperfect-janak: @highperformancecoder   Oh yes, sorry about that. Commented it out. That block of code may go away finally but leaving for now in case some optimization is required/possible on other platforms.

PR #289, opened 2021-03-25
Double buffering

digiperfect-janak:
Double buffering is working fine, using an off-screen surface created using `cairo_surface_create_similar`

---
This change is [Reviewable](https://reviewable.io/reviews/highperformancecoder/minsky/289)

### DIFF
--- a/model/renderNativeWindow.cc
+++ b/model/renderNativeWindow.cc
@@ -40,26 +40,33 @@ namespace minsky
 {
   void RenderNativeWindow::renderFrame(unsigned long parentWindowId, int offsetLeft, int offsetTop, int childWidth, int childHeight)
   {
-    // bool isInitial = false;
     if (!(winInfoPtr.get())) {
-      // isInitial = true;
       winInfoPtr = std::make_shared<WindowInformation>(parentWindowId, offsetLeft, offsetTop, childWidth, childHeight);
     }
+    if(winInfoPtr->getRenderingFlag()) {
+      return;
+    }
 
-    winInfoPtr->clear();
-    auto tmp = winInfoPtr->getSurface();
+    winInfoPtr->setRenderingFlag(true);
 
-    //TODO:: Review if this paint with each frame (below 3 lines) are needed on other systems (Windows/MacOS). If not, delete the code
-    // if (isInitial || true)
-    // {
-    //   cairo_move_to(tmp->cairo(), 0, 0);
-    //   cairo_set_source_rgb(tmp->cairo(), 1, 1, 1);
-    //   cairo_paint(tmp->cairo());
-    // }
+    auto surfaceToDraw = winInfoPtr->getBufferSurface();
+    surfaceToDraw.swap(surface);
+    
+    // Draw a white rectangle and set source rgb back to black
+    // TODO:: Must be implemented in canvas - as depending on context colors might change. Also instead of white, should be transparent
 
-    tmp.swap(surface);
+    cairo_set_source_rgb(surface->cairo(), 1, 1, 1);
+    cairo_rectangle(surface->cairo(), 0, 0, surface->width(), surface->height());
+    cairo_fill(surface->cairo());
+    cairo_set_source_rgb(surface->cairo(), 0, 0, 0);
+
     redraw(0, 0, surface->width(), surface->height());
-    tmp.swap(surface);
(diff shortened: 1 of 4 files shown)


model/renderNativeWindow.cc
```diff
@@ -40,26 +40,33 @@ namespace minsky
...
-    // bool isInitial = false;
     if (!(winInfoPtr.get())) {
-      // isInitial = true;
       winInfoPtr = std::make_shared<WindowInformation>(parentWindowId, offsetLeft, offsetTop, childWidth, childHeight);
     }
+    if(winInfoPtr->getRenderingFlag()) {
+      return;
+    }
+
+    winInfoPtr->setRenderingFlag(true);
```
highperformancecoder, line 50: Given this server is single threaded, is this flag logic necessary?

model/windowInformation.cc
```diff
@@ -67,44 +71,68 @@ namespace minsky
...
+  ecolab::cairo::SurfacePtr WindowInformation::getBufferSurface()
+  {
+    return bufferSurface;
   }
 
-  void WindowInformation::createSurface(){
+  void WindowInformation::copyBufferToMain()
+  {
+    cairo_set_source_surface(windowSurface->cairo(), bufferSurface->surface(), 0, 0);
+    cairo_paint(windowSurface->cairo());
```
highperformancecoder, line 94: I am sceptical this will be performant enough. We'll see.

model/windowInformation.cc
```diff
@@ -67,44 +71,68 @@ namespace minsky
...
 #else
-      {childSurface.reset(new cairo::Surface(cairo_xlib_surface_create(getDisplay(), getChildWindowId(), wAttr.visual, childWidth, childHeight), childWidth, childHeight));
-  cairo_surface_set_device_offset(childSurface->surface(), -wAttr.x, -wAttr.y);
-}
+    {
+      windowSurface.reset(new cairo::Surface(cairo_xlib_surface_create(getDisplay(), childWindowId, wAttr.visual, childWidth, childHeight), childWidth, childHeight));
+      cairo_surface_set_device_offset(windowSurface->surface(), -wAttr.x, -wAttr.y);
+
+
+      bufferSurface.reset(new cairo::Surface(cairo_surface_create_similar(windowSurface->surface(), CAIRO_CONTENT_COLOR_ALPHA, childWidth, childHeight), childWidth, childHeight));
```
highperformancecoder, line 114: Does this paint to the same XWindow as the original, or a new child window? The Cairo documentation doesn't give much of a clue.

highperformancecoder (approved):
I'm going to merge this, and have a play with it, but I'm still not convinced it is quite right. It looks like you might be copying the contents of the active XWindow onto itself - I really don't understand why that should fix the flickering.

model/renderNativeWindow.cc
digiperfect-janak: Indeed, the flag is not necessary for single threaded application.

model/windowInformation.cc
digiperfect-janak: Please do let me know once you have tested. It seemed to work good enough when I and Ravi tested. Maybe there is an even faster way but this seemed simple and worked.

digiperfect-janak:
I borrowed ideas from https://lists.gnu.org/archive/html/gnustep-dev/2011-10/msg00003.html
We are not copying contents of window onto itself. The buffer surface is a new off-screen surface. Application draws to it. Then we just copy contents of this surface to the x11 window surface.   `cairo_surface_create_similar` attempts to create a surface "as similar as possible" to the original surface - specially designed for the purpose of copying I read somewhere.

highperformancecoder:
On Fri, Mar 26, 2021 at 03:15:52AM -0700, Janak wrote:
> @digiperfect-janak commented on this pull request.
> 
> 
> In model/windowInformation.cc:
> 
> > +    cairo_set_source_surface(windowSurface->cairo(), bufferSurface->surface(), 0, 0);
> +    cairo_paint(windowSurface->cairo());
> 
> Please do let me know once you have tested. It seemed to work good enough when
> I and Ravi tested. Maybe there is an even faster way but this seemed simple and
> worked.

I agree it works (which is why I accepted the PR). However,
performance does seem to be lacking - for example when exposing the
window by moving another window on top. I would expect performance to be on par with the TCL/Tk version.

I'll investigate some more next week.


-- 

                      http://www.hpcoders.com.au

PR #294, opened 2021-06-12
Fixed incorrect usage of device offset and made redraw return a bool

digiperfect-janak:
Hi Russel,

This (along with a small change in the electron app) fixes the mouse offset issue we have been encountering

---
This change is [Reviewable](https://reviewable.io/reviews/highperformancecoder/minsky/294)

### DIFF
--- a/model/CSVDialog.cc
+++ b/model/CSVDialog.cc
@@ -271,7 +271,7 @@ namespace
   };
 }
 
-void CSVDialog::redraw(int, int, int width, int height)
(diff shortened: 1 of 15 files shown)


digiperfect-janak:
@highperformancecoder   I have made redraw return a _bool_ to indicate if something was drawn on the canvas. This helps us avoid the "dummy copy to png buffer" when something is drawn (it is needed only when model goes out of view).

PR #317, opened 2022-09-26
context menus for Ravel

Rdmkr:
includes various dialog windows for editing ravel handle properties & lock groups

- - -
This change is [Reviewable](https://reviewable.io/reviews/highperformancecoder/minsky/317)


model/ravelWrap.cc
```diff
@@ -439,7 +449,13 @@ namespace
   /// @throw if type does not match global dimension type
   void Ravel::setDimension(Dimension::Type type,const std::string& units)
   {
-    auto descr=description();
+    return setDimension(selectedHandle(), type, units);
+  }
+
+  /// @throw if type does not match global dimension type
```
highperformancecoder, line 455: Minor thing - it is not necessary to duplicate the doxygen comment from the declaration to definition.

model/ravelWrap.cc
```diff
@@ -439,7 +449,13 @@ namespace
   /// @throw if type does not match global dimension type
   void Ravel::setDimension(Dimension::Type type,const std::string& units)
   {
-    auto descr=description();
+    return setDimension(selectedHandle(), type, units);
```
highperformancecoder, line 452: This compiles? I didn't think it possible to return a void. It would save a lot of contorted generic code if possible.
Apparently so: https://stackoverflow.com/questions/2249108/can-i-return-in-void-function

Just learned something new.

gui-js/apps/minsky-electron/src/app/managers/CommandsManager.ts
```diff
@@ -1428,4 +1421,94 @@ export class CommandsManager {
...
+      Object.defineProperty(window,'dontCloseOnReturn',{value: true,writable:false});
+    }
+
+    static async saveHandleDimension(payload: HandleDimensionPayload) {
+      const args = JSON5.stringify([payload.handleIndex, payload.type, payload.units]);
+      await RestServiceManager.handleMinskyProcess({ command: `${commandsMapping.CANVAS_RAVEL_HANDLE_SET_DIMENSION} ${args}` });   
+    }
+
+    static async pickSlices(handleIndex: number) {
+      const allSliceLabels = <string[]>await RestServiceManager.handleMinskyProcess({ command: commandsMapping.CANVAS_RAVEL_HANDLE_ALL_SLICE_LABELS });
```
highperformancecoder, line 1461: Is the <string[]> a type cast operator, like the "as string[]" suffix form?

gui-js/libs/shared/src/lib/constants/constants.ts
```diff
@@ -180,6 +181,30 @@ export const commandsMapping = {
...
+  CANVAS_RAVEL_HANDLE_PICKED_SLICE_LABELS: '/minsky/canvas/item/pickedSliceLabels',
+  CANVAS_RAVEL_HANDLE_PICK_SLICE_LABELS: '/minsky/canvas/item/pickSliceLabels',
+  CANVAS_RAVEL_LOCKGROUP: '/minsky/canvas/item/lockGroup',
+  CANVAS_RAVEL_LOCKGROUP_HANDLELOCKINFO: '/minsky/canvas/item/lockGroup/handleLockInfo',
+  CANVAS_RAVEL_LOCKGROUP_ALL_LOCK_HANDLES: '/minsky/canvas/item/lockGroup/allLockHandles',
+  CANVAS_RAVEL_LOCK_RAVELS_IN_SELECTION: '/minsky/canvas/lockRavelsInSelection',
+  CANVAS_RAVEL_LOCKGROUP_SET_LOCK_HANDLES: '/minsky/canvas/item/lockGroup/setLockHandles',
+  CANVAS_RAVEL_LOCKGROUP_RAVELNAMES: '/minsky/canvas/item/lockGroup/ravelNames',
+  CANVAS_RAVEL_LOCKGROUP_VALIDATE_LOCK_HANDLE_INFO: '/minsky/canvas/item/lockGroup/validateLockHandleInfo/',
+
```
highperformancecoder, line 207: Yeah - personally I think this commandsMapping structure is a mistake. I'd prefer to have the string explicitly coded at the call site, not buried away in this enum in a totally different part of the codebase, as it adds one level of abstraction too many. But this can be fixed up with a future refactor.

I'm working on a wrapper that turned these REST calls into standard Typescript calls, with proper compile time type checking, so that'll get rid of most of this nonsense :).

highperformancecoder (approved):

gui-js/libs/ui-components/src/lib/ui-components.module.ts
```diff
@@ -12,7 +14,6 @@ import { EditUserFunctionComponent } from './edit-user-function/edit-user-functi
 import { EquationsComponent } from './equations/equations.component';
 import { FindAllInstancesComponent } from './find-all-instances/find-all-instances.component';
 import { GodleyWidgetViewComponent } from './godley-widget-view/godley-widget-view.component';
-import { VariablePaneComponent } from './variable-pane/variable-pane.component';
```
highperformancecoder, line 15: I suppose variable-pane.component being included into ui-components/src/index.ts suffices to include this module into the software. What is the purpose of this file then?

model/ravelWrap.cc
Rdmkr: I don't remember doing that very intentionally, but yeah, no complaints from the compiler. There are some other languages I know that allow this, even though it's always a word more verbose than necessary :)

gui-js/apps/minsky-electron/src/app/managers/CommandsManager.ts
Rdmkr: yes... although it's best to think of it as a type hint for the compiler... in runtime this does absolutely nothing. If I recall correctly, the handleMinskyProcess method returns an Unknown type, which you're pretty much forced to cast to something else to stop the compiler from complaining. One option we always have in typescript is to cast to <any>, which removes all type-based compilation checks. Can be convenient when you're in a hurry, but obviously a double-edged sword.

gui-js/libs/shared/src/lib/constants/constants.ts
Rdmkr: Sounds good. Definitely in line with my thinking, as we've discussed before.

gui-js/libs/ui-components/src/lib/ui-components.module.ts
Rdmkr: I think the only thing the presence of the file in intex.ts enables is the possibility to import the component with an alias... including the components in a module happens in the .module.ts files like this one. Specifically the component needs to be in declarations to be used within the module itself, and in exports to be used in other modules that reference the current one.
Generally the import alias is not used (by convention) for files within the same module... within a module we tend to use relative paths, although aliasing is allowed...

Rdmkr: example of using the alias:
import { events, MinskyProcessPayload, DescriptionPayload, HandleDescriptionPayload, HandleDimensionPayload, PickSlicesPayload, LockHandlesPayload } from '@minsky/shared';

PR #318, opened 2022-09-28
Develop

highperformancecoder:
This change is [Reviewable](https://reviewable.io/reviews/highperformancecoder/minsky/318)


gui-js/apps/minsky-electron/src/app/backend/backend.ts
```diff
@@ -0,0 +1,120 @@
...
+      (key,this.valueType? new this.valueType(cmd): this.callMethod(cmd));
+  }
+  insert(keyValue: Pair<Key, Value>) {this.callMethod("@insert",keyValue);}
+  erase(key: Key) {this.callMethod("@erase",key);}
+};
+
+export class Container<Key,Value=Key> extends CppClass
+{
+  type: any;
+  constructor(prefix: string, type: any=null) {super(prefix); this.type=type}
```
Rdmkr, line 106: Single-line formatting like this is kinda frowned upon in most circles I go... You're packing a lot of information in a small number of lines... which I like on some level, but I admit I had to read it a few times to fully grasp what's going on.

You can shorten the this.type = type stuff by declaring type as "protected type?" in the constructor. The question mark will also handle the optionality with null default (although I think it will use "undefined" instead of "null", which may have subtly different behavior!).

Depending on tsconfig settings, specifying the type of a variable as "any" may also not be strictly necessary.

gui-js/apps/minsky-electron/src/app/backend/backend.ts
```diff
@@ -0,0 +1,120 @@
...
+    log.error('backend called without any command');
+    return {};
+  }
+  if (!restService) {
+    log.error('Rest Service not ready');
+    return {};
+  }
+  try {
+    var arg='';
+    if (args.length>0)
```
Rdmkr, line 50: I noticed there are some formatting peculiarities that you're partial to, such as the lack of spaces around the > symbol here... I'm normally fine with that, but it might be worth it for us to settle on a set of standards and apply them consistently across the TS code base... maybe even automatically. Currently there isn't much consistency...

Rdmkr (approved):

gui-js/apps/minsky-electron/src/app/backend/backend.ts
```diff
@@ -0,0 +1,120 @@
+import {  importCSVerrorMessage, } from '@minsky/shared';
+import { Utility } from '../utility';
+import { dialog } from 'electron';
+import * as elog from 'electron-log';
+import { join } from 'path';
+
+const JSON5 = require('json5');
```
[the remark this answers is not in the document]
highperformancecoder: I think I went with this at the time because I couldn't figure which of the many import statements work with this package. But I'll try again.

gui-js/apps/minsky-electron/src/app/backend/backend.ts
```diff
@@ -0,0 +1,120 @@
...
+    this.second=value;
+  }
+};
+
+export class Map<Key, Value extends CppClass> extends CppClass
+{
+  valueType: any;  // stash a reference to the actual type here, for use in a new expression
+  constructor(prefix: string, valueType: any=null) {super(prefix); this.valueType=valueType;}
+  elem(key: Key) {
+    const cmd=this.prefix+"/@elem/"+JSON5.stringify(key)+"/second";
```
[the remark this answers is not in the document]
highperformancecoder: Makes sense.

gui-js/apps/minsky-electron/src/app/backend/backend.ts
highperformancecoder: If we do that, the only thing I ask is to settle on a formatting program to do it automatically. I don't want to think about whether to put spaces in or not when writing code. And I certainly don't care about it when reading code.

gui-js/apps/minsky-electron/src/app/backend/backend.ts
highperformancecoder: I'm used to the more condensed style. Vertical space is at a premium, you want maximum context in your editor window. So I find the more spaced out code styles slightly irritating, actually. However, I can happily work in code bases that insist on that sort of formatting if there's an automatic formatting tool that can be used - eg clang-format for C++ as used in another project I work with.

gui-js/apps/minsky-electron/src/app/backend/backend.ts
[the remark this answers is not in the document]
highperformancecoder: Sure - the back tick composition of strings is a really nice feature of typescript. Happy to use it here, and anywhere else.

gui-js/apps/minsky-electron/src/app/backend.ts
```diff
@@ -0,0 +1,63 @@
...
+
+setErrorDialog((message: string)=>{
+  dialog.showMessageBoxSync(WindowManager.getMainWindow(),{
+    message: message,
+    type: 'error',
+  });
+});
+
+// Sanity checks before we get started
+if (backend("/minsky/minskyVersion")!=version)
```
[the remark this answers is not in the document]
highperformancecoder: Sure.

gui-js/apps/minsky-electron/src/app/backend/backend.ts
highperformancecoder: Sounds good. The main thing I want is to avoid spending time in code reviews over "mere" style things, when it's more important to focus on maintainability, possible bugs and sometimes performance things. I've been on teams that had rigid style guides, and others that were a free-for-all. To be honest, consistency in style is overrated, but it is good to use idiomatic expressions where established.

gui-js/apps/minsky-electron/src/app/backend/backend.ts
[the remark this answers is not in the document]
highperformancecoder: `import * as JSON5 from 'json5';`works fine.

gui-js/apps/minsky-electron/src/app/backend/backend.ts
Rdmkr: Style consistency can be useful for avoiding merging complications, though, especially when enforced automatically.
Also if it needed to be said: the reason I'm focusing on style related factors is that everything looks excellent on the other fronts. I don't expect my future reviews will have to deal with it much.

PR #319, opened 2022-10-24
Continue backend refactor

highperformancecoder:
This change is [Reviewable](https://reviewable.io/reviews/highperformancecoder/minsky/319)


highperformancecoder:
This is still work in progress, but I'd be keen to get a review of what's done. I have implemented automatic exposure of the backend C++ Minsky API to both the main and renderer processes, using a common set of proxy objects, with the renderer proxy using IPC to communicate the marshalled C++ call to the main process for passing the call to the actual C++ addon.

I have committed minsky.ts to the repo, even though it is a built file, the reason being is that it will enable someone to work completely within the typescript world without building the C++ code. But that file (some 2500 line at present) shouldn't be counted in the LOC as it is automatically generated, in which case this PR reduces the LOC by about 250 LOC.

The advantage now is that the typescript compiler will pick up at compile time any backend API changes, which is a much more robust way of ensuring the frontend code matches the backend API.

I'm still working to exterminate commandsMapping from the codebase.

Rdmkr (commented):
> Reviewed 65 of 80 files at r1, 4 of 5 files at r2, 10 of 10 files at r3, all commit messages.
> *[Reviewable](https://reviewable.io/reviews/highperformancecoder/minsky/319)* status: 79 of 80 files reviewed, 1 unresolved discussion (waiting on @highperformancecoder)

___
*[`gui-js/libs/shared/src/lib/backend/minsky.ts` line 16 at r3](https://reviewable.io/reviews/highperformancecoder/minsky/319#-NFPbYsj24f5dNUEGOWW:-NFPbYsj24f5dNUEGOWX:b-5hrpof) ([raw file](https://github.com/highperformancecoder/minsky/blob/d60effe1b841bd7a4633dbe15993a69428090673/gui-js/libs/shared/src/lib/backend/minsky.ts#L16)):*
> ```XML
> class ecolab__TCL_args {}
> 
> export class Item extends CppClass {
> ```

The fact that this file is auto-generated excuses it to some extent, but this file seems excessively large and containing too many classes... might it be worth splitting up to make it easier to see at a glance which classes this API is composed of?

highperformancecoder:
*[`gui-js/libs/shared/src/lib/backend/minsky.ts` line 16 at r3](https://reviewable.io/reviews/highperformancecoder/minsky/319#-NFPbYsj24f5dNUEGOWW:-NFReoSv-Bus4UhMvZ8i:b-xjpjol) ([raw file](https://github.com/highperformancecoder/minsky/blob/d60effe1b841bd7a4633dbe15993a69428090673/gui-js/libs/shared/src/lib/backend/minsky.ts#L16)):*
<details><summary><i>Previously, Rdmkr wrote…</i></summary><blockquote>

The fact that this file is auto-generated excuses it to some extent, but this file seems excessively large and containing too many classes... might it be worth splitting up to make it easier to see at a glance which classes this API is composed of?
</blockquote></details>

Yes this is a good idea, and I did think about it, particularly as for one hot minute, it looked like the Typescript compiler was going to be overwhelmed.
However, it is  not so trivial to implement, as generating the necessary import statement automatically will require additional analysis of what C++ type depends on what. ATM, apart from the baseclass dependence on a few classes (Item, OperationBase and VariableBase), which we can declare first, remaining class references can rely on Typescript's implicit forward declaration to work. 
So for now, lets add a ticket to come back to this at some point.

PR #327, opened 2022-11-17
Ravel 224 bump electron

highperformancecoder:
This change is [Reviewable](https://reviewable.io/reviews/highperformancecoder/minsky/327)


highperformancecoder:
This PR is about updating the version of Electron. From Electron 12, a tighter security model, providing context isolation between renderer processes and the main process. Electron and node functionality is not available by default in the renderer process. To wit, no more remote api, ipcRenderer is not available by default, but must be passed in using a context bridge in preload.ts - even that is not recommended - it should really just be an explicit API defined on the context bridge. We can do that in a later refactor. For now, however, we pass in the ipcRenderer object, which is exposed via the electronService API. I don't think it is a security problem, since we are loading our own pages, but we shouldn't define preload for hosted 3rd party pages. ATM, all 3rd party pages are passed off to the OS installed browser via the shell plugin, but if we want to do a more integrated hosting, it's something to keep in mind.

Rdmkr (commented):
> Reviewed 24 of 38 files at r1, 5 of 5 files at r2, 8 of 8 files at r3, 2 of 2 files at r4, 2 of 2 files at r5, all commit messages.
> *[Reviewable](https://reviewable.io/reviews/highperformancecoder/minsky/327)* status: all files reviewed, 2 unresolved discussions (waiting on @highperformancecoder)

___
*[`gui-js/libs/ui-components/src/lib/plot-widget-view/plot-widget-view.component.ts` line 34 at r5](https://reviewable.io/reviews/highperformancecoder/minsky/327#-NH4Jd1F7A1ZcMiG908l:-NH4Jd1F7A1ZcMiG908m:bvcedom) ([raw file](https://github.com/highperformancecoder/minsky/blob/c7df1dead538385aefcd7f7fbc1a444244572c64/gui-js/libs/ui-components/src/lib/plot-widget-view/plot-widget-view.component.ts#L34)):*
> ```XML
> 
>   ngOnInit() {
>     setTimeout(()=>{this.render();},10);
> ```

using setTimeout to await view initialization seems questionable; can you see if putting this.render() in ngAfterViewInit() instead works? Also add the interface AfterViewInit if you do this.
___
*[`gui-js/libs/ui-components/src/lib/wiring/wiring.component.ts` line 41 at r5](https://reviewable.io/reviews/highperformancecoder/minsky/327#-NH4L79Z3pWVo-UQu1_q:-NH4L79Z3pWVo-UQu1_r:b-bi42y9) ([raw file](https://github.com/highperformancecoder/minsky/blob/c7df1dead538385aefcd7f7fbc1a444244572c64/gui-js/libs/ui-components/src/lib/wiring/wiring.component.ts#L41)):*
> ```XML
>     const isMainWindow = await this.windowUtilityService.isMainWindow();
>     if (isMainWindow)
>       setTimeout(async ()=> 
> ```

try ngAfterViewInit()

highperformancecoder:
This is exactly why I like code reviews. The setTimeout stuff is just a more-or-less direct translation of Tk's update command, which interrupts the flow of control to ask the window system to draw the updates on screen. But Angular's lifecycle hooks sounds like the better way to do this, so I'll take a deep dive into this.

highperformancecoder:
ngAfterViewInit is still not late enough. There is similar code (using ngAfterViewInit) in godley-widget-view, which does work, so its very strange.

I do think the problem is that this stuff should really be happening on the main process. There is an APP_LAYOUT_CHANGED event handler in electron.events.ts, which is how the main window handles resize events (and presumably on opening the main window initially). This event handler should be expanded to handle all toplevel windows. But that's way too much code change to fit into this PR, so I'll raise a separate ticket for that.

PR #344, opened 2022-12-23
Feature/christmas fixes

Rdmkr:
This change is [Reviewable](https://reviewable.io/reviews/highperformancecoder/minsky/344)

### DIFF
--- a/model/canvas.cc
+++ b/model/canvas.cc
@@ -148,15 +148,15 @@ namespace minsky
           if (auto to=closestInPort(x,y)) {
             model->addWire(static_cast<shared_ptr<Port>&>(fromPort),to);
             fromPort.reset();
-            auto toItem = to.get()->item();
-            if (auto ravel=static_cast<Ravel&>(toItem))
+            auto& toItem=to->item();
+            if (auto ravel=dynamic_cast<Ravel*>(&toItem))
             {
-              auto state = ravel.getState();
+              auto state = ravel->getState();
               if(state.empty()) {
                 minsky().reset();
               }
-            } else if(auto sheet=static_cast<Sheet&>(toItem)) {
-              if(sheet.empty()) {
+            } else if(auto sheet=dynamic_cast<Sheet*>(&toItem)) {
+              if(sheet->empty()) {
                 minsky().reset();
               }
             }
--- a/model/canvas.cc
+++ b/model/canvas.cc
@@ -145,9 +145,25 @@ namespace minsky
       }
     if (fromPort.get())
       {
-          if (auto to=closestInPort(x,y))
+          if (auto to=closestInPort(x,y)) {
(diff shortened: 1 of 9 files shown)


model/canvas.cc
```diff
@@ -145,9 +145,25 @@ namespace minsky
       }
     if (fromPort.get())
       {
-          if (auto to=closestInPort(x,y))
+          if (auto to=closestInPort(x,y)) {
             model->addWire(static_cast<shared_ptr<Port>&>(fromPort),to);
-        fromPort.reset();
+            fromPort.reset();
+            auto toItem = to.get()->item();
+            if (auto ravel=static_cast<Ravel&>(toItem))
```
highperformancecoder, line 152: This won't work. static_cast<Ravel&> just tells the compiler that this is a Ravel, regardless of whether it actually is.
In fact it is not, currently, because it has been sliced on line 151.

What you want is `auto ravel=dynamic_cast<Ravel*>(&toItem)`, as dynamic_cast returns nullptr if it is not, in fact a Ravel.

dynamic_cast<Ravel&> is different - it throw an exception it the item is not a Ravel - sometimes that is what you want to do - no checks are required before using the result.

model/canvas.cc
```diff
@@ -145,9 +145,25 @@ namespace minsky
...
+            {
+              auto state = ravel.getState();
+              if(state.empty()) {
+                minsky().reset();
+              }
+            } else if(auto sheet=static_cast<Sheet&>(toItem)) {
+              if(sheet.empty()) {
```
highperformancecoder, line 159: If we're doing this, maybe we could define empty() to be a virtual method of Item.  empty() is perhaps the wrong name - maybe something like inputDefined() might be better. There is something like that already for variables, so that could be generalised.  This will obviate any requirement for dynamic_cast.

highperformancecoder (changes requested):

model/canvas.cc
```diff
@@ -145,9 +145,25 @@ namespace minsky
...
+          if (auto to=closestInPort(x,y)) {
             model->addWire(static_cast<shared_ptr<Port>&>(fromPort),to);
-        fromPort.reset();
+            fromPort.reset();
+            auto& toItem=to->item();
+            if (auto ravel=dynamic_cast<Ravel*>(&toItem))
+            {
+              auto state = ravel->getState();
+              if(state.empty()) {
+                minsky().reset();
```
highperformancecoder, line 156: This is OK for now, but I had another thought overnight. In addon.cc, there is a separate thread that runs when the backend has been idle for a little  - around 100ms I think. Why not check the state of the reset_needed flag, and call reset at that point. That way, if reset takes a little while (which will be the case for some extensive tensor operations), it will not interfere with model responsiveness when doing multiple mousing operations on the screen.

I'm a little worried about the call to reset on line 144 for similar reasons too.

highperformancecoder (approved):

PR #346, opened 2022-12-29
PlotWidget::addConstantCurves(): Get underlying pointer from vectors directly

aledomu:
This change is [Reviewable](https://reviewable.io/reviews/highperformancecoder/minsky/346)

### DIFF
--- a/model/plotWidget.cc
+++ b/model/plotWidget.cc
@@ -427,7 +427,7 @@ namespace minsky
     // determine if any of the incoming vectors has a ptime-based xVector
     xIsSecsSinceEpoch=false;
     for (auto& i: yvars)
-      if (i && xvars[&i-&yvars[0]] && !i->hypercube().xvectors.empty())
+      if (i && xvars[&i-yvars.data()] && !i->hypercube().xvectors.empty())
         {
           const auto& xv=i->hypercube().xvectors[0];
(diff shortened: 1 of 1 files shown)


model/plotWidget.cc
```diff
@@ -427,7 +427,7 @@ namespace minsky
     // determine if any of the incoming vectors has a ptime-based xVector
     xIsSecsSinceEpoch=false;
     for (auto& i: yvars)
-      if (i && xvars[&i-&yvars[0]] && !i->hypercube().xvectors.empty())
+      if (i && xvars[&i-yvars.data()] && !i->hypercube().xvectors.empty())
```
highperformancecoder, line 430: This is probably mostly stylistic, but I suppose &x[0] does involve adding 0 to the pointer, so I suppose data() is better.

highperformancecoder (approved):

model/plotWidget.cc
aledomu: It is not merely stylistic and it is not the reason why I did it. The problem lies in that it might trigger an assertion in the stdlib in case the vector is empty, because it may try accessing the element and then return the pointer.

highperformancecoder: Assertion will not be triggered, as the code is not run when yvars is empty.

PR #361, opened 2023-02-12
Feature/parameter variable tabs

Rdmkr:
This change is [Reviewable](https://reviewable.io/reviews/highperformancecoder/minsky/361)


highperformancecoder:
We do need to think about where we want to get to. IMHO, it was probably a mistake to toggle variable display on the variable tab - we should have all variables displayed on the tab, like the parameter tab. Ultimately, we want these tabs to be editable.

What is toggleable is whether variable definition networks appear in the canvas or not. For an editable tab, we will need to create code to parse the definition expression, and convert it to a network and lay it out on the canvas. Some of the pieces are in place, but not all.

highperformancecoder:
Sorry for getting side-tracked on other things. I've decided to create a [ticket](https://sourceforge.net/p/minsky/features/212/) to deal with the issue of refactoring the code to use VariableValues, and the attributes that should be in the table. 

That gets rid of a big impediment to accepting this PR. I also took this branch for a spin to see how it works in practice:

1. The font chosen for the Godley table looks pretty ugly when italicised. Maybe use the global font preference for that? But LaTeX rendering otherwise looks good.
2. Glyphs for ←→↑↓ in the row/column buttons just show up as boxes.
3. Variable/parameter tabs - names should be LaTeX rendered, like the Godley table.
4. The last button on variables tab saying "remove" - please change this to "Show on Canvas". We shouldn't be removing the variable from the tab, but that is current behaviour, so we can fix this in ticket feature-#212

Rdmkr:
"Glyphs for ←→↑↓ in the row/column buttons just show up as boxes."

Ideally I'd like to install an icon library like Font Awesome or Material Icons for this... Just spent some time trying to get the latter to recognize the relevant font but this doesn't seem to be a straightforward in an Electron app...

Rdmkr:
"The font chosen for the Godley table looks pretty ugly when italicised. Maybe use the global font preference for that? But LaTeX rendering otherwise looks good.¨

Are you talking about the Flows column? These are currently Latex-rendered, so they can contain mathematical symbols if desired. If this is not a requirement, I agree the standard font makes more sense.

highperformancecoder:
> "The font chosen for the Godley table looks pretty ugly when italicised. Maybe use the global font preference for that? But LaTeX rendering otherwise looks good.¨
> 
> Are you talking about the Flows column? These are currently Latex-rendered, so they can contain mathematical symbols if desired. If this is not a requirement, I agree the standard font makes more sense.

Yes - that is the latex rendered flow columns. Yes - we do want latex rendering, but the italic font is pretty ugly. If that is not so easy to do with MathJAX, then raise a ticket to come back to that later.

highperformancecoder:
> 

Can't we just use the UTF-8 characters above, and just render in the normal font?

highperformancecoder:
> > "The font chosen for the Godley table looks pretty ugly when italicised. Maybe use the global font preference for that? But LaTeX rendering otherwise looks good.¨
> > Are you talking about the Flows column? These are currently Latex-rendered, so they can contain mathematical symbols if desired. If this is not a requirement, I agree the standard font makes more sense.
> 
> Yes - that is the latex rendered flow columns. Yes - we do want latex rendering, but the italic font is pretty ugly. If that is not so easy to do with MathJAX, then raise a ticket to come back to that later.

Funnily enough, viewing this through vncviewer on my laptop, using your latest commits, it actually looks OK. Just not on my desktop monitor previously. Lets see if it causes an issue with the end-users.

highperformancecoder:
Just checked the latest on your branch - still not rendering variable names as LaTeX on the var/param tabs, and still not rendering the arrow glyphs on the Godley Table. Both are showstoppers for merging into master, but fix those, and I think this is good to go.

Rdmkr:
"Can't we just use the UTF-8 characters above, and just render in the normal font?"
I've been having trouble getting those to show up large enough... I don't think their shape is ideal for buttons like these. We need something with a more square-shaped outline. I managed to get things working with material icons in the latest commit, please have a look. Material icons will be useful for many other purposes. I can't live without an icon library of this kind in any of my other development projects.
We do need to check if the node_modules path resolution in styles.scss will keep working in every execution context, for example after all the files are built and packaged up for final release...

Rdmkr:
"still not rendering variable names as LaTeX on the var/param tabs¨

Can you check again, I'm positive it's in there and it's working on my PC. This taken care of by the latex directive on the div in parameters.component.html:

<div [title]="variable.name" class="data-cell" latex [equation]="variable.name"></div>

highperformancecoder:
All good on both the arrow buttons and the LaTeX rendering on the var/par tabs.

On Sun, Feb 26, 2023 at 01:15:37PM -0800, Rdmkr wrote:
> "Can't we just use the UTF-8 characters above, and just render in the normal
> font?"
> I've been having trouble getting those to show up large enough... I don't think
> their shape is ideal for buttons like these. We need something with a more
> square-shaped outline. I managed to get things working with material icons in
> the latest commit, please have a look. Material icons will be useful for many
> other purposes. I can't live without an icon library of this kind in any of my
> other development projects.
> We do need to check if the node_modules path resolution in styles.scss will
> keep working in every execution context, for example after all the files are
> built and packaged up for final release...
>

Sure - but I guess that's my issue as packager maintainer :).


-- 

                      http://www.hpcoders.com.au

highperformancecoder:
On Sun, Feb 26, 2023 at 01:23:22PM -0800, Rdmkr wrote:
> "still not rendering variable names as LaTeX on the var/param tabs¨
> 
> Can you check again, I'm positive it's in there and it's working on my PC. This
> taken care of by the latex directive on the div in parameters.component.html:

Yup - working now with you latest pushes.


-- 

                      http://www.hpcoders.com.au

PR #363, opened 2023-02-16
Update ExprTk package

ArashPartow:
This change is [Reviewable](https://reviewable.io/reviews/highperformancecoder/minsky/363)


highperformancecoder:
This is failing continuous integration. What should I expect from the new version of exprtk?

ArashPartow:
I think the following two lines in ExprTk and the way the exprtk include is being wrapped in an anon namespace are causing issues:

```
using ::exprtk::details::char_ptr;
using ::exprtk::details::char_cptr;
```

and

```
#pragma GCC visibility push(hidden)
  namespace {
    // include anonymously to reduce the number of exported linker symbols
    #include <exprtk/exprtk.hpp>
```

It's generally not a good idea to have includes within namespaces.

highperformancecoder:
I know - there was a reason I did that at the time, knowing full well it will create maintenance issues going forward. As I think I mentioned to you, I was pushing exprtk well beyond its design limits, whilst praising your software for being able to do so :).

I merely asked what are the benefits to upgrading the exprtk reference now, so I can prioritise when to do the upgrade, as it will take a bit of work.

highperformancecoder:
Actually that code comment indicates I was trying to address the slow linker times, which wasn't too successful. On Linux I get good results using the clang toolchain, but on Windows, we're stuck with that god awful PE/COFF format.

ArashPartow:
@highperformancecoder The main changes are:

1. Runtime checks for expressions (eg: loops iterations, vector and string element accesses)
2. Stackoverflow protections during compilation and evaluation
3. Various evaluation performance improvements

Most of the benefit will be useful when processing expressions from unverified/untrusted sources, so it may not be a big deal for now. 

-----------------------------

In any case I'm going to see if i can replicate the error in a smaller context and see what can be done from my end.

highperformancecoder:
Sure - if you get it through CI, I'll just tick and flick.

Otherwise - I'll create a ticket for this, and get onto it in due course.

highperformancecoder:
I have now upgraded exprtk to the latest, and all looks good. You're right, the anonymous namespace wrapper in a vain attempt to optimise linker times was the culprit.

PR #366, opened 2023-03-13
Avoid accessing the first element of vectors when only the pointer to the array is needed

aledomu:
This avoids the assertion checking even with the flag enabled

- - -
This change is [Reviewable](https://reviewable.io/reviews/highperformancecoder/minsky/366)


highperformancecoder:
I agree with this - data() is the much more modern way of doing this. I've been  a bit old-school about this, particularly as some of this code predates C++11. Ready to merge when you're ready.

aledomu:
Done, I was waiting for the checks to pass.

PR #367, opened 2023-03-13
Linux: don't build Tk GUI if not necessary

aledomu:
This change is [Reviewable](https://reviewable.io/reviews/highperformancecoder/minsky/367)

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -140,17 +140,30 @@ endif
 
 ALL_OBJS=$(MODEL_OBJS) $(ENGINE_OBJS) $(SCHEMA_OBJS) $(GUI_TK_OBJS) $(TENSOR_OBJS) $(RESTSERVICE_OBJS) RESTService.o httpd.o addon.o typescriptAPI.o
 
-EXES=gui-tk/minsky$(EXE) RESTService/minsky-RESTService$(EXE) RESTService/minsky-httpd$(EXE)
+
+ifneq ($(GUI_TK),1)
+
+EXES=RESTService/minsky-RESTService$(EXE) RESTService/minsky-httpd$(EXE)
+ifeq ($(HAVE_NODE),1)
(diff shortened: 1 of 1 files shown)


highperformancecoder:
We still need the Tk build for some regression tests, and utility scripts. It is also handy from some debugging cases. I plan to replace Tk completely with a python binding in due course, but it is relatively low priority, so not yet scheduled.

I might also add the export:package step to the main Makefile, so that you don't need to compile the Typescript *every* *single* *time* you run Minsky.

aledomu:
This does not remove the ability to compile the Tk version. It only disables it for the standard compilation procedure.

highperformancecoder:
> This does not remove the ability to compile the Tk version. It only disables it for the standard compilation procedure.

I was aware of that. But the Tk version really doesn't add any build time when building in parallel, as the final linkage step for the node addon takes so long, and it would be annoying to have to turn around and build the Tk version when needed (which is often).

If it helps, we don't need to build the Tk version on either Windows or Mac, so I'm happy if you want to disable the builds on those systems. Windows builds have MXE set to true, and Mac builds have OS set to Darwin.

aledomu:
I chose the last option, with the caveat that explicitly setting `GUI_TK=0` does avoid the compilation of the Tk GUI binary in Linux.

highperformancecoder:
OK - that will work. Although I think the logic might be simpler if you define GUI_TK=1 in the Makefile, then call make with GUI_TK= (ie undefining it) to disable the gui-tk build.

You could do something similar with a GUI_JS flag as well.

aledomu:
`GUI_TK=1` disables the REST service compilation, so it is the same as having `GUI_JS=1`, but without needing a separate parameter. I prefer it this way since it has the nice effect of keeping the same behaviour as until now if no flag is set.

PR #370, opened 2023-03-30
Enable range checks for arrays, vectors and strings in debug builds

aledomu:
`-Wp,-D_GLIBCXX_ASSERTIONS` is the flag that has been causing the problems in Fedora. With the recent changes that have been merged and in my cursory testing I have not come across any crash. I suggest this flag is enabled in upstream itself (here) to protect against this class of bugs.

- - -
This change is [Reviewable](https://reviewable.io/reviews/highperformancecoder/minsky/370)

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -175,6 +175,10 @@ FLAGS+=-std=c++14 -Ischema -Iengine -Imodel -Icertify/include -IRESTService -IRa
 # NB see #1486 - we need to update the use of rsvg, then we can remove -Wno-deprecated-declarations
 #-fvisibility-inlines-hidden
 
+ifeq ($(DEBUG), 1)
+FLAGS+=-Wp,-D_GLIBCXX_ASSERTIONS
+endif
+
 VPATH= schema model engine gui-tk RESTService RavelCAPI/civita RavelCAPI $(ECOLAB_HOME)/include 
 
 .h.xcd:


aledomu:
> Actually something else that's puzzling. -Wp means pass the following to the C preprocessor. But in this case, -D_GLIBCXX_ASSERTIONS defines that macro, so that is passed to the preprocessor anyway. The -Wp leadin should have no effect.

I did it that way because it is how the default flags are set up in Fedora packages.

> I'm not convinced. Happy to have these enabled for a DEBUG build, but bounds checking usually has significant performance impact, and I can't see anything in the documentation to say it doesn't.

My real concern is that catching this class of bugs right now depends on the user using Fedora. I would prefer if it got to the whole userbase so testing is more extensive.

> Also, will it work, now that we've switched to using clang by default?

I am using gcc, with no testing on clang. I will report on this one.

highperformancecoder:
As I mentioned, happy to have this enabled as a DEBUG build flag, but not for production.

Also, for Fedora build processes, you can provide a mechanism for adding this flag to the build line. What I usually do is pass special flags via the OPT argument. Normally, OPT="-g" for DEBUG=1 and "-O3 -DNDEBUG" for DEBUG undefined, but can be overridden on the command line.

aledomu:
What I mean is that Fedora by default does _production_ builds for all packages with this flag, which is why minsky was crashing in my setup. There is no need to add it (in fact, if it is desirable and possible to remove it, then so be it).

I can change this PR so it is just enabled for debug builds, but this issue might arise again in the future. Just for comparison, KiCad developers had the same discussion and, according to Fedora packagers, the performance impact is "negligible" (`-D_GLIBCXX_DEBUG` is the one that is actually expensive). They really discourage avoiding these checks. Link to the discussion: https://lists.fedoraproject.org/archives/list/devel@lists.fedoraproject.org/thread/TMDJMX2OYE7Z2WHKXMOMTOBDHJKZJ3NV/

highperformancecoder:
On Fri, Mar 31, 2023 at 04:16:56AM -0700, Alejandro Domínguez wrote:
> What I mean is that Fedora by default does production builds for all packages
> with this flag, which is why minsky was crashing in my setup. There is no need
> to add it (in fact, if it is desirable and possible to remove it, then so be
> it).
> 
> I can change this PR so it is just enabled for debug builds, but this issue
> might arise again in the future.

Yes - that would be best. As developers, we run minsky in debug build
a lot of the time, which have assertions enabled for all sorts of
things, but not vector subscript bounds checking, so this would, in
fact, be a useful addition in debug builds. But I would not want a
hard crash out of a production build - instead any stoppage must be
via an exception, which gets displayed back to the user as an error
message.

I'm not convinced that "performance impact is neglible". As I
understand it, consecutive access is somehow optimised away, but
random access is not. So there may be some circumstances where the
extra overhead is invisible, but others where it is not.


> Just for comparison, KiCad developers had the
> same discussion and, according to Fedora packagers, the performance impact is
> "negligible" (-D_GLIBCXX_DEBUG is the one that is actually expensive). They
> really discourage avoiding these checks. Link to the discussion: https://
> ***@***.***/thread/
> TMDJMX2OYE7Z2WHKXMOMTOBDHJKZJ3NV/
> 
> —
> Reply to this email directly, view it on GitHub, or unsubscribe.
> You are receiving this because you commented.*Message ID: <highperformancecoder
> ***@***.***>
> 

-- 

                      http://www.hpcoders.com.au

aledomu:
I will disable this flag for rpm builds in another PR, just to keep the scope here.

aledomu:
I do not comprehend why the tests are failing, again.

> Also, will it work, now that we've switched to using clang by default?

It seems to accept the flag, so yes.

Apart from the test results, I think this is finished.

highperformancecoder:
On Tue, Apr 04, 2023 at 07:34:50AM -0700, Alejandro Domínguez wrote:
> I do not comprehend why the tests are failing, again.

My fault. Minsky needs to load a node module, which AFAICT is done via
the "require" command. For some reason, the require command in
typescript differs from the require command in javascript, in that if
you pass it an absolute path that you have carefully calculated prior,
it makes its own damned mind up about where to look for the file!!!
:P.

Up to now, I had been loading the module from
../../node-addons/minskyRESTService.node, which worked because CWD was
set based on where the application was launched.

But when opening minsky from the "shell" in Windows, the CWD could be
anywhere, or even undefined. So in order to get double clicking to
work from Windows Explorer, I have to set the CWD to the directory
containing the executable. All fine and good, only it buggers up the
jest tests. 

I managed to get the jest tests to work locally in the develop branch,
but they fail to run in the CI environment. It is a complete mess of
layers of abstractions, supposedly to make the developer's life
easier, but in fact making it more difficult by hiding the source
code, and not providing any documentation about what is being done
"under the hood".

I have now disabled the failing jest test for CI builds. Sorry for the rant!

> 
>     Also, will it work, now that we've switched to using clang by default?
> 
> It seems to accept the flag, so yes.
> 
> Apart from the test results, I think this is finished.

I think so too. I'll try to merge today.


-- 

                      http://www.hpcoders.com.au

PR #372, opened 2023-04-04
Fix errors when building with clang

aledomu:
This change is [Reviewable](https://reviewable.io/reviews/highperformancecoder/minsky/372)


aledomu:
This one is needed so clang stops complaining about `move()` not being superseded by the `std` namespace.

highperformancecoder:
I'm using clang all the time, and I've never seen this error. But this is a harmless change, so I'll merge it now.

PR #376, opened 2023-04-12
Ravel 209 progess bar csv import

highperformancecoder:
This change is [Reviewable](https://reviewable.io/reviews/highperformancecoder/minsky/376)


highperformancecoder:
This PR extends the progress bar work previously done to handle cases where the number of steps is unknown, but the fraction of work done is known, and also to enable a big red cancel button on the progress dialog.

PR #378, opened 2023-04-24
Feature 208 summary tab

highperformancecoder:
This change is [Reviewable](https://reviewable.io/reviews/highperformancecoder/minsky/378)


Rdmkr (commented):
:lgtm:

> Reviewed 33 of 33 files at r1, all commit messages.
> *[Reviewable](https://reviewable.io/reviews/highperformancecoder/minsky/378#-:-NToPBs64dJA-GH4WabI:bnfp4nl)* status: all files reviewed, 3 unresolved discussions (waiting on @highperformancecoder)

___
*[`gui-js/libs/shared/src/lib/constants/constants.ts` line 14 at r1](https://reviewable.io/reviews/highperformancecoder/minsky/378#-NToMGLt-6aufyEqzd2M:-NToMGLt-6aufyEqzd2N:b-2y3kh7) ([raw file](https://github.com/highperformancecoder/minsky/blob/01a72a6475a133a3febb55a9a1899b18025d8431/gui-js/libs/shared/src/lib/constants/constants.ts#L14)):*
> ```XML
> export const RESET_ZOOM_FACTOR = 1;
> 
> export const OPEN_DEV_TOOLS_IN_DEV_BUILD = true;
> ```

was committing this intentional? In all honestly I can't think of a good reason why this shouldn't always be true since it only happens in dev build anyway.
___
*[`gui-js/libs/ui-components/src/lib/summary/summary.component.html` line 7 at r1](https://reviewable.io/reviews/highperformancecoder/minsky/378#-NToMf6s1cCjH0Jb4fvf:-NToMf6tDg4Y9Nc-L5NJ:bfsan3n) ([raw file](https://github.com/highperformancecoder/minsky/blob/01a72a6475a133a3febb55a9a1899b18025d8431/gui-js/libs/ui-components/src/lib/summary/summary.component.html#L7)):*
> ```HTML
>         <span class="caret" (click)="toggleCaret($event)">{{labels[category]}} - {{numVars[category]}}
>           <div *ngFor="let type of types(category)" class="nested">
>             <span class="caret" style="margin:10px" (click)="toggleCaret($event)">{{type}} - {{this[category][type]?.length}}
> ```

it's preferable to put style specifications in .scss files and refer to them via classes
a single style attribute is not so bad, but multiple of them quickly start to litter the HTML code and make it less readable. also class definitions stimulate you to generalize your style rules.
for consistency we also tend to prefer having style info in one place, which is preferably the .scss files.
___
*[`gui-js/libs/ui-components/src/lib/summary/summary.component.html` line 43 at r1](https://reviewable.io/reviews/highperformancecoder/minsky/378#-NToOMe708rIHWbAaSb7:-NToOMe708rIHWbAaSb8:b-eq5orb) ([raw file](https://github.com/highperformancecoder/minsky/blob/01a72a6475a133a3febb55a9a1899b18025d8431/gui-js/libs/ui-components/src/lib/summary/summary.component.html#L43)):*
> ```HTML
>         <span class="caret" (click)="toggleCaret($event)">{{labels[category]}} - {{numVars[category]}}
>           <div *ngFor="let type of types(category)" class="nested">
>             <span class="caret" style="margin:10px" (click)="toggleCaret($event)">{{type}} - {{this[category][type]?.length}}
> ```

another style tag

highperformancecoder:
On Mon, Apr 24, 2023 at 10:56:36AM -0700, Rdmkr wrote:
> 
>     export const OPEN_DEV_TOOLS_IN_DEV_BUILD = true;
> 
> was committing this intentional? In all honestly I can't think of a good reason
> why this shouldn't always be true since it only happens in dev build anyway.

No it wasn't intentional - in fact I remember correcting it back to
false at the last minute, so maybe I forgot to push.

Not a biggie anyway - dev tools active can be annoying when debugging
non-html parts of the application, but its easy enough to switch this
flag. Even better would be if there was some kind of command line
switch "npm start -- --enable-dev-tool" or similar, but npm is such a
black box, its hard to figure what its doing half the time.

> 
> it's preferable to put style specifications in .scss files and refer to them
> via classes
> a single style attribute is not so bad, but multiple of them quickly start to
> litter the HTML code and make it less readable. also class definitions
> stimulate you to generalize your style rules.
> for consistency we also tend to prefer having style info in one place, which is
> preferably the .scss files.
> 

Fair enough - it started being in a single location, and when you're
not sure whether it will work, but then it got copied :). I'll call it
"one-indent". We may need to do a two-indent in the second version,
subject to user reactions.


-- 

                      http://www.hpcoders.com.au

PR #416, opened 2024-01-09
Feature/csv selectable dimensions

Rdmkr:
This change is [Reviewable](https://reviewable.io/reviews/highperformancecoder/minsky/416)

### DIFF
--- a/gui-js/libs/ui-components/src/lib/import-csv/import-csv.component.ts
+++ b/gui-js/libs/ui-components/src/lib/import-csv/import-csv.component.ts
@@ -77,6 +77,7 @@ export class ImportCsvComponent extends Zoomable implements OnInit, AfterViewIni
   selected: boolean[]; ///< per column whether column is selected
   mouseDown = -1;       ///< record of column of previous mouseDown
   dialogState: any;
+  existingDimensionNames: string[];
   @ViewChild('checkboxRow') checkboxRow: ElementRef<HTMLCollection>;
   @ViewChild('importCsvCanvasContainer') inputCsvCanvasContainer: ElementRef<HTMLElement>;
   @ViewChild('fullDialog') fullDialog: ElementRef<HTMLElement>;
@@ -181,6 +182,10 @@ export class ImportCsvComponent extends Zoomable implements OnInit, AfterViewIni
       }
     });
     document.onkeydown = this.onKeyDown;
+
+    this.electronService.minsky.dimensions.$properties().then(dims => {
+      this.existingDimensionNames = Object.entries(dims).map(d => d[0]);
+    });
   }
 
   ngAfterViewInit() {
@@ -195,7 +200,8 @@ export class ImportCsvComponent extends Zoomable implements OnInit, AfterViewIni
       this.selectRowAndCol(this.dialogState.spec.dataRowOffset, this.dialogState.spec.dataColOffset);
     })();
   }
-  ngAfterViewChecked() {
+
+  afterViewChecked() {
     // set state of dimension controls to reflect dialogState
     if (this.inputCsvCanvasContainer) {
       var table = this.inputCsvCanvasContainer.nativeElement.children[0] as HTMLTableElement;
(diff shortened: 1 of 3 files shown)


gui-js/libs/ui-components/src/lib/import-csv/import-csv.component.ts
```diff
@@ -181,6 +182,10 @@ export class ImportCsvComponent extends Zoomable implements OnInit, AfterViewIni
       }
     });
     document.onkeydown = this.onKeyDown;
+
+    this.electronService.minsky.dimensions.$properties().then(dims => {
+      this.existingDimensionNames = Object.entries(dims).map(d => d[0]);
+    });
```
highperformancecoder, line 188: Wouldn't it be simpler to use the Object.keys() method here? That way, you could write something like:
`this.existingDimensionNames =(await this.electronService.minsky.dimensions.$properties()).keys();`

gui-js/libs/ui-components/src/lib/import-csv/import-csv.component.ts
```diff
@@ -195,7 +200,8 @@ export class ImportCsvComponent extends Zoomable implements OnInit, AfterViewIni
       this.selectRowAndCol(this.dialogState.spec.dataRowOffset, this.dialogState.spec.dataColOffset);
     })();
   }
-  ngAfterViewChecked() {
+
+  afterViewChecked() {
```
highperformancecoder, line 204: What went wrong here?

highperformancecoder:
One further comment - this PR address the horizontal dimension box only, and works as it should. However, for ravel #413, I think we also need to apply this to the Name fields of all the axis columns, and also include the content of the header row cell as one of the choices.

gui-js/libs/ui-components/src/lib/import-csv/import-csv.component.ts
Rdmkr: I'm reverting this change, but I am adding the AfterViewChecked interface to make clearer this is referring to a component lifecycle hook (did it even work without that interface..?)

gui-js/libs/ui-components/src/lib/import-csv/import-csv.component.ts
Rdmkr: Changing ngOnInit to an async method is not recommended practice. The code will not be awaited so the .then syntax is less misleading. I will use the Object.keys suggestion.

gui-js/libs/ui-components/src/lib/import-csv/import-csv.component.ts
Rdmkr: I guess using the interface is optional, but strongly recommended practice in Angular.

Rdmkr:
Enabled this for the column names as well.

Rdmkr:
If you give me a moment I will also add the cell header value as an option, that still isn't there.

Rdmkr:
Made one more change that makes it so the selectable column names do not contain duplicates.

gui-js/libs/ui-components/src/lib/import-csv/import-csv.component.ts
highperformancecoder: Why is declaring ngOnInit async considered bad practice? IIUC, Angular will never await a call on that method, but an explicit call could. So where's the problem?

In any case, for our purposes, the .then syntax is equivalent, since this method is not called explicitly, so its just a matter of style.

gui-js/libs/ui-components/src/lib/import-csv/import-csv.component.ts
Rdmkr: The comment that calling an ng-prefixed method from own code is bad practice also applies here... it's what confused me to make the initial change here. Those "checked" hooks are a bit of a can of worms by the way. If the change detection strategy of the component is not set to "On Push"*, the onChecked hook methods get called extremely often, on things like mouse move events and in response to anything changing in the component template... These methods should never contain performance intensive logic and their use should preferably be avoided.

* at my current company we have a policy to use this change detection strategy on most of our components... refactoring our code to this (which mostly prevents negative performance surprises) would require changing a lot of variables that are dynamically read from the HTML template to rxjs Subjects with .next used to set their value. Probably a bridge too far at this point, but worth knowing about.

PR #483, opened 2024-04-29
Make resizeOnContents exposed on Group context menu.

highperformancecoder:
Fix up scaling of group items in Group::resize()
For Ravel #524.

- - -
This change is [Reviewable](https://reviewable.io/reviews/highperformancecoder/minsky/483)


highperformancecoder:
Merging because we now have two many open PRs.

PR #485, opened 2024-04-30
Construct a dense x dataset when the passed in x variable is sparse w…

highperformancecoder:
…ithin PlotWidget. For Ravel #536.

- - -
This change is [Reviewable](https://reviewable.io/reviews/highperformancecoder/minsky/485)


highperformancecoder:
Merging because we have too many open PRs.

PR #489, opened 2024-05-03
Ravel 545 difference bug

highperformancecoder:
This change is [Reviewable](https://reviewable.io/reviews/highperformancecoder/minsky/489)


highperformancecoder:
merging because too many open PRs

PR #506, opened 2024-05-27
Feature/csv download by line

Rdmkr:
This is very close to a working solution, but crashes hard (most of the time..?) on my PC... any idea what could be wrong? The logs aren't much help. Also I wonder how much of the logic from loadWebFile we still need, as it seems to me most of it is covered by electron's file download client.

- - -
This change is [Reviewable](https://reviewable.io/reviews/highperformancecoder/minsky/506)


gui-js/apps/minsky-electron/src/app/managers/CommandsManager.ts
```diff
@@ -1186,14 +1186,60 @@ export class CommandsManager {
...
+
+  static startCSVDownload(payload: DownloadCSVPayload) {
+    const window=this.createDownloadWindow();
+
+    const splitUrl = payload.url.split('.');
+    const extension = splitUrl.length === 1 ? 'csv' : splitUrl.pop();
+    const savePath = join(tmpdir(),`downloaded.${extension}`);
+    window.webContents.session.on('will-download', (e,i,w) => this.downloadCSV(e,i,w,savePath));
+    window.webContents.downloadURL(payload.url);
+    return savePath;
```
highperformancecoder, line 1198: This is returning too early. electron.events is awaiting on this, will return the file name to electronservice, which then returns to importCSV, which will then attempt to open the file, which is still in the process of being downloaded.

So we need to make this method async, create a promise here to be returned, and arrange for the promise to be resolved/rejected in the done handler of webContents.

Granted, Promises are not exactly the easiest APIs to wrap you head around, but there's an example of doing exactly this in CommandsManager.newPubTab().

highperformancecoder (changes requested):
Aside from one method that needs to be async, everything else looks good to me.

gui-js/apps/minsky-electron/src/app/managers/CommandsManager.ts
```diff
@@ -1186,14 +1186,60 @@ export class CommandsManager {
...
     });
   }
-  
-  static upgrade() {
-    let window=WindowManager.createWindow({
+
+  static startCSVDownload(payload: DownloadCSVPayload) {
+    const window=this.createDownloadWindow();
+
+    const splitUrl = payload.url.split('.');
```
highperformancecoder, line 1193: I think it would be more reliable to use the url package (https://nodejs.org/api/url.html) to grab the path name, then use the path package (https://nodejs.org/api/path.html) to extract the extension.

But even so, not all URLs even name the file, so it might be better to use the 'will-download' handler to grab the file name that the web service names it, and then pass the results back to the caller via a promise.

Rdmkr:
I dealt with the timing and URL parsing issues, but there is still a crash.

highperformancecoder:
> This is very close to a working solution, but crashes hard (most of the time..?) on my PC... any idea what could be wrong? The logs aren't much help. Also I wonder how much of the logic from loadWebFile we still need, as it seems to me most of it is covered by electron's file download client.
> 
> This change is [Reviewable](https://reviewable.io/reviews/highperformancecoder/minsky/506)

I think it is right to discard the loadWebFile C++ method.

I'll take this for a spin now, and see if I can spot the cause of the crash.

highperformancecoder:
The first file causes an assertion failure due to an out of bounds access on an empty line. I'll fix that tomorrow.
The second file loads fine - see the image snapshot.
paleo

highperformancecoder:
So the crash you noticed was a bug that was already fixed on master.
The land-open temperature index cannot quite be imported, because the merge-delimiters option is not quite working. I've raised a ticket for that, but it needn't block merging this PR, as it is completely orthogonal to the concerns of this PR.
There's 3 lines where the build is breaking.
We can raise a ticket to deal with saving the URL to schema for downloading updates later. Its not something we're supporting right now. We're planning on a full release early next week, so I'm keen to get this change merged ASAP.

Rdmkr:
Just committed some C++ build fixes.

Rdmkr:
I also committed code for zip decompression. This takes the first file in a zip... not sure what to do if there are multiple, but it's already a stretch to accommodate zip file users...

highperformancecoder:
> I also committed code for zip decompression. This takes the first file in a zip... not sure what to do if there are multiple, but it's already a stretch to accommodate zip file users...

Just a minor thing - you missed committing the changes to package.json. I did a "npm install decompress", and all was good.

Oh - the failure to exit on close stuff - I'm getting a message about     at Map.forEach (<anonymous>)
    at BrowserWindow.<anonymous> (/home/rks/github/minsky-niels/gui-js/dist/executables/linux-unpacked/resources/app.asar/minsky-electron/main.js:158:108)

I'd say the download window is not being destroyed properly. I'll take a look.

highperformancecoder:
The culprit is the progress bar. Putting       progress.close(); as the last line of the 'done' handler in CommandsManager.downloadCSV() solves the problem.

For completeness, it might be worth adding that line to the done handlers in downloadMinsky() and downloadRavel(), although I've never seen the same problem, possibly because app.quit() is called in those methods.

highperformancecoder (dismissed):

gui-js/apps/minsky-electron/src/app/managers/CommandsManager.ts
```diff
@@ -1144,16 +1150,31 @@ export class CommandsManager {
...
+      }
+    });
   }
 
   // handler for downloading Minsky
   static downloadMinsky(event,item,webContents) {
     item.setSavePath(join(tmpdir(),item.getFilename()));
 
-    let progress=new ProgressBar({text:"Downloading Ravel application",value: 0, indeterminate:false, closeOnComplete: true,});
+    let progress=new ProgressBar({text:"Downloading Minsky application",value: 0, indeterminate:false, closeOnComplete: true,});
```
highperformancecoder, line 1172: We're actually rebadging this as "Ravel application", rather than Minsky. Purely for marketing reasons :).

Rdmkr:
One remaining issue I can think of: this puts one downloaded.{extension} file in /tmp for every extension that is ever downloaded. In the case of zip it will first put the .zip file there and then the inner file next to it. Maybe we should delete files from there that aren't used.

Rdmkr:
What's still preventing us from completing this PR?

highperformancecoder:
On Sat, Jun 01, 2024 at 08:02:35AM -0700, Rdmkr wrote:
> What's still preventing us from completing this PR?

Nothing! I intend to merge tomorrow morning :). I just had other
things to do today. I only switched on my laptop for the first time
today 10 minutes ago.


-- 

                      http://www.hpcoders.com.au

highperformancecoder:
One slight hiccup - the hang on quit bug was back again, and closing the progress bar turned out to be a red herring (sorry!). I was able to solve the problem by adding a 'destroyed' handler to each Windows' webContents to remove the window from the activeWindows map.

PR #526, opened 2024-11-06
Made suggested changes to deleteKey.sh and now it passes

wdednam:
I see you already made the changes you suggested when you first merged this test, but it was failing on my side, so I also made your suggested changes. It now works, but probably because of the change in the coordinates of the wire that gets deleted. I also tried to pythonise three other tests this week, but have not managed to get them to pass. TCL GUI interactions are difficult to port to pyhton.

- - -
This change is [Reviewable](https://reviewable.io/reviews/highperformancecoder/minsky/526)

### DIFF
--- a/test/00/deleteKey.sh
+++ b/test/00/deleteKey.sh
@@ -39,6 +39,7 @@ minsky.canvas.recentre()
 # Store initial counts of items and wires
 numItems = minsky.model.numItems()
 numWires = minsky.model.numWires()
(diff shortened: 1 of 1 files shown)


wdednam (commented):
> Reviewed 1 of 1 files at r1, all commit messages.
> *[Reviewable](https://reviewable.io/reviews/highperformancecoder/minsky/526)* status: <span>:shipit:</span> complete! all files reviewed, all discussions resolved (waiting on @wdednam)

